PR #73, opened 2014-07-14
Numpy arrays

marscher:
With this PR sliced array operations will return numpy ndarrays of corresponding type containing the requested range. If numpy is not found during the setup process, the old behaviour of returning python lists of python objects is preserved. The new code is 5 times faster than the old one.

```
 jpype.JArray(jpype.JDouble, 1)(np.random.random(10**7))
 %timeit a[:] # np array
 #10 loops, best of 3: 56.6 ms per loop
 %timeit a[:] # pythonic list
 1 loops, best of 3: 334 ms per loop
```

### DIFF
--- a/src/native/common/jp_primitivetypes_autogen.cpp
+++ b/src/native/common/jp_primitivetypes_autogen.cpp
@@ -25,6 +25,13 @@
 	#define PY_ARRAY_UNIQUE_SYMBOL jpype_ARRAY_API
 	#define NO_IMPORT_ARRAY
 	#include <numpy/arrayobject.h>
+#else
+    #define NPY_BYTE 0
+    #define NPY_SHORT 0
+    #define NPY_INT 0
+    #define NPY_LONG 0
+    #define NPY_FLOAT32 0
+    #define NPY_FLOAT64 0
 #endif
 
 #define CONVERSION_ERROR_HANDLE \


pbcornelius:
Hi, I was just playing around with this great feature and wondering whether there are any plans to add support for 2D arrays (it was referenced [here](https://github.com/jpype-project/jpype/issues/133#issuecomment-99867636), but it doesn't look like it was implemented). Especially getting many Strings out of Java can take some time, so converting them to arrays of bytes might be faster?

Thrameos:
I am currently reworking class structure of JPype so that it is never running interpreted code during the return process.  We have already cut the time for type resolution for calling in half using the new class conversion API.   Thus it may address your needs as if we can improved the return time by a factor of 2 to 5 then requiring a special call would be less warranted.  Please see the thread regarding performance issues for details.   I currently have 2 major errors (one undercounted reference on a thread, and one overreference) and single digits worth of minor errors before I can get a test version with some hard performance numbers to see what the payoff will be.  Thus far I have killed 2 sections of interpreted code (JClassFactory, JObject.__new__), a few table look ups, and about a half dozen object creations (Wrapper, PyJPValue, tuples, argument lists, strings) and replaced it with an order 1 lookup and one direct object creation, so I expect it will be a substantial improvement.  It is taking a long while as it is a substantial overhaul change the structure that much.

Long term I will be reworking the Numpy interfaces so that it is not using the deprecated API.  I will try at that time to make it so that object arrays can be pulled through using the slice or toarray() method.  The limited factor for this is that JNI only supports block copy methods for primitives thus if I have to pull a large number of items I still have to pull each one individually which the JNI layer.  Arrays are objects to Java so everything that holds for objects also holds for multiple dimensional arrays.  Thus once I speed up the return path this also may be much faster.  JNI doesn’t have any direct support for 2D arrays that I am aware of.

PR #120, opened 2015-02-15
Unit tests 

baztian:
Updated the unittest: Made them compatible with Python 3 (tested agains @tcalmant's port). Also made some minor improvements to the tests. Of course there are tests failing due to bugs in the Python 3 port that have already been fixed in origninell's version.


marscher:
thanks for the great job!

PR #122, opened 2015-02-24
py3 python code

baztian:
Some changes to the python code to make it more python 3 compatible. The cpp part is of course still missing.


baztian:
There are still some python parts that won't run on python 3. I'll make a diff of the jpype directory to tcalmant's code to see what we can use. I will add another pull request then.

PR #124, opened 2015-03-03
Fix OutOfMemoryError (Heap) object cleanup

### LINKED ISSUE #121: Possible (java heap) memory leak
I have JPype 0.5.7 on python 2.7.9.
Windows 7, oracle java 1.8.0_31

I see java objects being leaked, so that I can very quickly hit OutOfMemoryErrorPyRaisable

This is enough to reproduce the problem from me :
https://gist.github.com/simon-og/a422cd452a2a12ba9d1f

I can't reproduce the issue on 0.5.4.2 .

If I don't pass the object back into java I don't see the problem (comment out line 6)

Let me know if you need any more information.

baztian:
This fixes #121. @marscher, please review if there isn't anything that should stay as it is/was.
I've chagned -Xms to 16M. Otherwise the new test runs unnecessarily long.

### DIFF
--- a/native/common/jp_class.cpp
+++ b/native/common/jp_class.cpp
@@ -605,7 +605,7 @@ bool JPClass::isSubclass(JPClass* o)
 	JPCleaner cleaner;
 
 	jclass jo = o->getClass();
-	cleaner.addGlobal(jo);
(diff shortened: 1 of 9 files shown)


marscher:
thank you!

PR #128, opened 2015-03-10
[setup] handle jni_md for all known systems

marscher:
simplified code a bit and fix freebsd jni_md.h inclusion. We can add support for more systems more easily.


baztian:
Can you please add a line to the changelog? I'm guessing it fixes JPype build onf FreeBSD machines?

marscher:
Actually this does not add new functionality but simplifies the current one and makes it more easy to extend. I'm not sure, if that is worth a user's changelog entry.

PR #130, opened 2015-04-05
Update install.rst

kristi:
add pip install command


baztian:
Thanks for the pull request.

As you'll find in the Install section you also have the possibility to download from git and install via setup.py. Can you please restructure your PR to make it clear one could either install via pip or via setup.py?

Issue #119, opened 2015-02-07
Port to Python 3


baztian:
Hi,
anyone willing to port jpype to python 3? I know there is the @tcalmant port. But it is not up to date.
I could help out porting the Python part but probably won't be a big help porting the C code ;)
Nice infos can be found here:
http://python3porting.com/
https://docs.python.org/3/howto/pyporting.html

tcalmant:
Hi,
When I started porting the project, I made a very small documentation (in French, sorry :p).
Here are the links, maybe Google translate can help for now. I'll translate them ASAP.

https://github.com/isandlaTech/cohorte-runtime/blob/master/documents/doc-jpype-py3/source/jpype-py.rst
https://github.com/isandlaTech/cohorte-runtime/blob/master/documents/doc-jpype-py3/source/jpype-c.rst
https://github.com/isandlaTech/cohorte-runtime/blob/master/documents/doc-jpype-py3/source/liens.rst

FYI, the most "painful" parts for the conversion are:
* the conversion of string/unicode to bytes/string in Python 3
* the replacement of PyObject by PyCapsule to let Python keep track of our C objects

Finally, the setup.py file of jpype-py3 adds support for Cygwin, if you're interested.

marscher:
@baztian porting the c-part is also the hardest part. I've tried once, but especially the string conversion is really painful.

tcalmant:
Here is a translated version of the C++ part of my previous links:
https://github.com/tcalmant/jpype-py3/wiki/py3_cpp

Another solution could be to update the jpype-py3 fork project, and to merge them once they'll have equivalent features :)

baztian:
Hi @tcalmant, your documentation looks really promising. Maybe it convices @marscher to give it another try on the c/c++ part :)

marscher:
a nice starting point would be to have equal test suites for py2/3, so
progress can be tracked during development.

baztian:
That's definitely something where I can help / what I can do. 



"Martin K. Scherer" <notifications@github.com> schrieb:
>a nice starting point would be to have equal test suites for py2/3, so
>progress can be tracked during development.
>
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/originell/jpype/issues/119#issuecomment-73742639

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

marscher:
[...]

PR #132, opened 2015-04-07
Feature/python3 compatibility4

kristi:
initial stab at Python3 compatibility

proxy unit tests cause segfault, but other unit tests pass

Tested on OSX 10.9 / (pyenv) Python 2.6, 2.7, 3.4

issue #119

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ language: python
 python:
   - '2.6'
   - '2.7'
+  - '3.0'
+  - '3.1'
+  - '3.2'
+  - '3.3'
+  - '3.4'
 
 env:
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,10 @@ language: python
 python:
   - '2.6'
   - '2.7'
-  - '3.0'
-  - '3.1'
-  - '3.2'
   - '3.3'
   - '3.4'
+  - "nightly"
+
 
 env:
   matrix:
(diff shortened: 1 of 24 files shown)


marscher:
can you edit your .travis.yml and add python 3 to the build matrix?

baztian:
Fantastic news! Good work @kristi!
The proxy tests did pass using our .py files with .so from tcalmant's fork as far as I remember. Maybe you can spot the relevant delta there...

marscher:
http://docs.travis-ci.com/user/languages/python/

Maybe you could also add 3.3?

marscher:
super cool! The tests are all running, very impressive! Have you also checked already if there are any memory leaks under Py3?

kristi:
No, I haven't checked for memory leaks yet. Do you have any recommended
tools or methods?

baztian:
I think we can merge that pull request, can't we? I would suggest to release a 0.6 (`bumpversion minor`) as Python 3 support is really a big new feature. If there are memory leaks they can be fixed later on. Thanks @kristi for making the cpp part. Absolutely awesome!

baztian:
Oh, I forgot something: We have to update the README and the documentation (still stating if you're looking for py 3 use the tcalmant version...)

marscher:
thank you very much @kristi! I will run some valgrind checks on the testsuite soon.

marscher:
valgrind reported no (new) issues. Only leaks in JVM8 itself, which we can not fix :+1:

Issue #106, opened 2014-09-26
jproxy passing general java objects


Eothred:
I have tried to read the documentation but I think I might not be enough of an expert to understand. Basically what I want to do is using in python an interface which defines a function that takes in another java object as input parameter. Looking at your tests it seems that you are only testing that you can pass primitive types (if that is the correct term). May I ask if this is supposed to be possible?

Something along the lines of:
```java
public interface MyInterface {
  public void myFun(AnotherInterface anotherInterface);
}
```
and then in Python:
```python
...
class MyInterface:
  def myFun(self, anotherInterface):
    anotherInterface.anotherFunction()
my_interface=MyInterface()
jpype.JProxy("MyInterface", inst=my_interface)
...
```

I have tried this with a toy setup. I managed fine to set up a proxy defining functions that would get doubles, integers or similar. As soon as I call a function like myFun() above though, the function runs once but then at exit I get a fatal error from Java:
```zsh
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00000001006a45b7, pid=8915, tid=19215
#
# JRE version: Java(TM) SE Runtime Environment (7.0_67-b01) (build 1.7.0_67-b01)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (24.65-b04 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [Python+0x555b7]  subtype_dealloc+0x42
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/yngvelevinsen/Downloads/hs_err_pid8915.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.sun.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
zsh: abort      python -d test_jpype_proxy.py
```

yetifrisstlama:
Hi,

I'm having the same problem here on
 * Linux  ( 2.6.32-504.3.3.el6.x86_64 )
 * java 1.7.0_45
 * Python 2.7.9 :: Anaconda 2.1.0 (64-bit)

I made a small test setup to reproduce the bug in an [ipython notebook](http://nbviewer.ipython.org/url/mbetz.web.cern.ch/mbetz/pyjapc/ReconstructCrash.txt).

[...]

PR #136, opened 2015-04-13
Fix proxy with argument issue. Calls PyGILState_Ensure/Release() to prot...

### LINKED ISSUE #67: testProxyWithArguments fails for  Python > 2.7.3
We didn't notice that, because travis currently uses 2.7.3. But this will be updated soon. It seems it happens somewhere in finalization code. Thats the stacktrace in current Python:

    #0  0x00007ffff6f06f79 in raise () from /lib/x86_64-linux-gnu/libc.so.6
    #1  0x00007ffff6f0a388 in abort () from /lib/x86_64-linux-gnu/libc.so.6
    #2  0x0000000000557f5f in Py_FatalError ()
    #3  0x00000000005537f7 in PyThreadState_Get ()
    #4  0x00000000004b108b in subtype_dealloc ()
    #5  0x000000000048dc57 in _Py_Dealloc ()
    #6  0x00007ffff65684b2 in PythonHostEnvironment::releaseRef(void*) () at src/native/python/py_hostenv.cpp:33
    #7  0x00007ffff653072f in HostRef::~HostRef() () at src/native/common/jp_env.cpp:332
    #8  0x00007ffff653075e in HostRef::~HostRef() () at src/native/common/jp_env.cpp:333
    #9  0x00007ffff65307ea in HostRef::release() () at src/native/common/jp_env.cpp:342
    #10 0x00007ffff652ffbe in JPCleaner::~JPCleaner() () at src/native/common/jp_env.cpp:209
    #11 0x00007ffff6548532 in Java_jpype_JPypeInvocationHandler_hostInvoke(JNIEnv_*, _jclass*, _jstring*, long, _jobjectArray*, _jobjectArray*, _jclass*) () at src/native/common/jp_proxy.cpp:138

baztian:
...ect ~JPCleaner() operations. (Cherry pick 9f559926 from tcalmant's fork. Fixes #67 and #106.)

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -4,6 +4,9 @@ Changelog
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
 - **Next version - unreleased**
+
+  - Fix proxy with arguments issue.
+
 - **0.6.0 - 2015-04-13**
 
   - Python3 support.
--- a/native/common/jp_env.cpp
+++ b/native/common/jp_env.cpp
@@ -17,8 +17,9 @@
 #include <jpype.h>
 
 /*
- * FIXME: use a less coupled way to call PyGILState_Ensure/Release() in
- * JPCleaner::~JPCleaner().
+ * FIXME: use a less coupled way to call PyGILState_Ensure/Release()
+ * in JPCleaner::~JPCleaner() if we wan't to target a non Python
+ * implementation.
  */
 #include <Python.h>
 
--- a/native/common/jp_env.cpp
+++ b/native/common/jp_env.cpp
@@ -16,6 +16,12 @@
 *****************************************************************************/   
 #include <jpype.h>
 
+/*
+ * FIXME: use a less coupled way to call PyGILState_Ensure/Release() in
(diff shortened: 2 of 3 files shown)


baztian:
@tcalmant do you remember why you commented your change with

    FIXME: use a less coupled way to call PyGILState_Ensure/Release() in JPCleaner::~JPCleaner().

I'm very bad in cpp and have no knowledge about the Python api. What is the problem with the include?

tcalmant:
Well, it's just an architectural note: ``JPCleaner`` is in the ``native/common`` folder, which means it should only work with the internal and Java API, not with the Python one.
I think that the source of JPype was initially structured to support multiple languages (Python, Ruby, ...), with a subfolder in ``native/`` for each of them.
Therefore the native/common folder shouldn't contain any Python-relative includes, which is not the case of ``JPCleaner`` after this commit.

A better way to do the job would be to accept macros in ``JPCleaner`` to call those methods in the case of compiling for Python.

But as JPype (currently) only targets Python, it is not (currently) an issue, but it's worth commenting ;)

baztian:
Thanks for the detailed explanation!

PR #137, opened 2015-04-13
Allow for faster travis builds with container based infrastructure.

baztian:
See http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/ and http://blog.travis-ci.com/2015-03-31-docker-default-on-the-way/


marscher:
nice!

Issue #140, opened 2015-05-13
Better error messages


kutschkem:
Some error messages don't give enough information for my taste. For example, the jp_method.cpp, the message "No matching overload found" should include information about the method name and the parameters that were searched.

baztian:
"Some error messages"  is a bit vague. Would be great if you can provide a pull request.

kutschkem:
:-P Fair enough. I meant explicitly that error message, since it is rather common and doesn't disclose what method it is trying to find with which parameters.

marscher:
I think it should be easy to just display the prototype of the function called, but "stringifying" all candidates (just in case of no match) is overkill and is likely to hurt performance.

Thrameos:
This is currently in the master branch with better text in the devel branch.  The question of performance depends on how often that an exception is issues.  The default case in python is that exceptions should be fairly light as it is supposed to be okay to try first and ask permission later.  However, this is a runtime  exception in which there is likely not going to be a good recovery, thus giving good feed back should not be an issue.  If we are concerned about performance, then we for this we should cache the alternative methods string on the first occurrence.

We will have a similar problem when we consider tracing information.  For the ease of users it is a good idea to have a fully integrated tracing output between Java, C++  and Python so that all the excecution path is clear.  But this means more overhead as we have to fake a traceback on the exception handling to create a object that knows the stack dump with the Java exception, the C++ stack and the existing python.  Python has pretty low overhead as it just keeps track of the internal execution stack.  But JPype does not have that option.  Thus we are going to need to decide which exception paths need to follow the Python philosophy and run with low overhead (say array out of bounds) and which we should pay the full price of clarity.

PR #144, opened 2015-06-23
improve error message for failed overload matching

kutschkem:
addresses #140 
This is not perfect, but shows the direction in which the error messages should go in my opinion. Even better would be to include the signature that the method tried to match. Hints at how to best achieve this would be appreciated ;-)

### DIFF
--- a/native/common/jp_method.cpp
+++ b/native/common/jp_method.cpp
@@ -81,7 +81,7 @@ void JPMethod::addOverloads(JPMethod* o)
 JPMethodOverload* JPMethod::findOverload(vector<HostRef*>& arg, bool needStatic)
 {
 	TRACE_IN("JPMethod::findOverload");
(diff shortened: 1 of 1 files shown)


marscher:
have a look at the matchreport method. It shows how to obtain the argument types.

kutschkem:
Actually, I can find no place where the HostRef is transformed to a String. I see I can query HostEnv about it, but can I also get the type? I can't even find a way to convert HostRef to JType. All the instances I have seen are a JType checking that the HostRef can be converted to it, the JType converting it. In order to properly display what arguments were given, I need to be able to print the type without having a signature beforehand, if that makes sense.

baztian:
Your pull request doesn't compile. See the [Travis output](https://travis-ci.org/originell/jpype/jobs/67965695).

marscher:
Think this has been addressed by #172

PR #154, opened 2015-09-07
Added "and" Python keyword

rrader:
Without this methods called "and" cannot be used via jpype


rrader:
build is failing for some reason

PR #156, opened 2015-09-15
suppose long when converting to java

Cosmius:
Python 3 does not have long type, which results an overflow when convert it to java long if it is too long


justussc:
this is actually a really necessary change for python3. maybe it would be worth adding a test in attr.py

the reason why this change is necessary is because we do
    `#define PyInt_Check PyLong_Check`
in `jpype.h` if we are compiling for python 3. then PythonHostEnvironment::intAsInt will return an int and python 3 long values get cast to an int losing the higher bits. 

i don't know the codebase well enough to judge if this is the right way of fixing it. but for example `jpype.java.util.Date(1447887600000)` currently does not work as expected with python 3.

@Cosmius if you do not have time to add a test i can create another pull request.

marscher:
closing in favour of #164

PR #157, opened 2015-09-18
Made some changes to userguide.rst.

icedwater:


baztian:
Thanks @icedwater for your excellent work!

Issue #159, opened 2015-10-22
Scala for Java Importing Error


widian:
I want to use some Jar file which is using Scala for Java packages, for instance scala.collection

But in progress of making _JavaClass, JPype1 raises below error.

TypeError: Cannot create a consistent method resolution
order (MRO) for bases scala.collection.GenTraversableLike$$Static, scala.collection.generic.HasNewBuilder$$Static

Maybe, I think it is conflict in inheritance definition. I try to avoid that case to modify a JPype1, but I couldn't. Anyone who can avoid this situation? 

These are used Scala for Java packages.

import scala.collection.Iterator;
import scala.collection.JavaConversions;
import scala.collection.Seq;

marcoeilers:
I had the same problem; I think the issue seems to be the following: 

JPype translates all Java types, both classes and interfaces, to Python types. That means that in complex subtyping hierarchies for interfaces, which may include diamonds and (importantly) the same parent interfaces several times in different orders for a given type, Python may not be able to find a consistent MRO. I think this is a general issue and has nothing to do with Scala, except that Scala has such complex hierarchies in commonly used types.

One solution I could think of is this: When computing the bases and meta bases for a new type, the base interfaces, whose order does not matter because they don't contain any actual method implementations, could be sorted (e.g. by name). This should not affect execution at all, but solves the problem of potentially having the same interfaces in different orders in a type's inheritance hierarchy.

I changed _jclass.py to include this code:

        # Prepare the meta-metaclass
        meta_bases = []
        def getMetaClass(base):
            if base is object:
                return cls
            else:
                return base.__metaclass__
        for i in bases:
            meta_bases.append(getMetaClass(i))

        minterfaces = []
        mclasses = []
        for mb in meta_bases:
            if hasattr(mb, '__isinterface__') and mb.__isinterface__:
                minterfaces.append(mb)
            else:
                mclasses.append(mb)
        def getName(metaclass):
            return metaclass.__name__
        minterfaces = sorted(minterfaces, key=getName)
        
        meta_bases = mclasses + minterfaces

        interfaces = []
        classes = []
        for b in bases:
            mb = getMetaClass(b)
            if hasattr(mb, 'isinterface') and mb.__isinterface__:
                interfaces.append(b)
            else:
                classes.append(b)
        def getBaseName(base):
            return getName(getMetaClass(base))

        interfaces = sorted(interfaces, key=getBaseName)

        bases = classes + interfaces     
[...]

PR #161, opened 2015-11-13
Fix MRO problems with interface hierarchies

justussc:
The test in mro.py demonstrates the problem leading to the classic mro error. In larger interface hierarchies this problem can be hidden far down in the hierarchy such that one cannot only drop duplicates from the immedate bases as was done until now. Using a metaclass the mro is replaced by a simple topsort of the hierarchy.

Quite likely also Issue #159 is fixed by this change. 

(its my first step on github so please advice if i should have done something different)

### DIFF
--- a/jpype/_jclass.py
+++ b/jpype/_jclass.py
@@ -99,6 +99,27 @@ def _javaGetAttr(self, name):
         return _jpype._JavaBoundMethod(r, self)
     return r
 
+def _mro_override_topsort(cls):
+    # here we run a topological sort to get a linear ordering of the inheritance graph.
+    parents = set().union(*[x.__mro__ for x in cls.__bases__])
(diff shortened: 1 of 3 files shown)


marscher:
Thank you for you contribution!
Can you please write a single comment in the code on how you are calculating the order? (and if you're willing to touch this again, please remove the semicolons, Python is not C) :+1:

justussc:
i removed the semicolons (a habit from to much java/c++ coding). 
i also added a comment at the top of the _mro_override_topsort  function. as there is no true multiple inheritance in java, any linear ordering of the inheritance graph should be ok. 

following a recommendation on stackoverflow i ammended my commit and force pushed it. let me know if that is not the way to do this.

marscher:
Thank you again!

PR #164, opened 2015-11-30
fix calls from python to methods taking integer/long primitives for larger intergers

justussc:
the most debatable changes in this PR are to `PythonHostEnvironment::isInt` and `PythonHostEnvironment::isLong`. the reason here is that on some platforms -- even with python2 -- integers have 64bit. if they are then returned in `jint PythonHostEnvironment::intAsInt` they get truncated. therefore calls with larger integers (>=2^31 or < -2^31) were broken. initially i though #156 was a good fix, but there was much more to this.  

i added couple of tests which cover the fixes. i started from jpypetest.attr.testCreateDate which is a very basic test.

i am happy to rework this change and would be glad if somebody could review it.

details from the commit:
   * overload resolution with JLong wrapper did not work (was checking for _int instead of _long)
   * on 64bit linux systems LONG_MAX is 2^63-1 and therefore returning JPyInt::asLong as a jint cuts of higher bits for values >= 2^31 or < -2^31
   * similar problem as above for all py3 on all systems as we have no long type anymore but only int
   * if a python exception was set we need to throw a PythonException not a JPypeException otherwise the message is overwritten
   * ref to member of temporary fixed in jpype_javaarray
   * tp_init has to either return 0 or -1

### DIFF
--- a/native/python/py_method.cpp
+++ b/native/python/py_method.cpp
@@ -284,12 +284,11 @@ int PyJPBoundMethod::__init__(PyObject* o, PyObject* args, PyObject* kwargs)
 		Py_INCREF(javaMethod);
 		self->m_Instance = inst;
 		self->m_Method = (PyJPMethod*)javaMethod;
-
 		return 0;
 	}
 	PY_STANDARD_CATCH
 
-	return 1;
+	return -1;
 }
 
 PyObject* PyJPBoundMethod::__call__(PyObject* o, PyObject* args, PyObject* kwargs)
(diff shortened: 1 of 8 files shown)


native/python/py_method.cpp
```diff
@@ -336,8 +335,8 @@ void PyJPBoundMethod::__dealloc__(PyObject* o)
 	TRACE_IN("PyJPBoundMethod::__dealloc__");
 	PyJPBoundMethod* self = (PyJPBoundMethod*)o;
 
-	Py_DECREF(self->m_Instance);
-	Py_DECREF(self->m_Method);
+	Py_XDECREF(self->m_Instance);
```
marscher: So we are not sure anymore that instance pointer is set?

marscher:
Thank you for this fix! It is well tested and I'm fine with merging it. @baztian what do you think?

native/python/py_method.cpp
justussc: it is a little defensive, but after correcting the return value in PyJPBoundMethod::__init__ i thought that if we'd throw before setting self->m_Instance then this would crash.

marscher: alright, that makes sense.

PR #166, opened 2015-12-09
disable memoryview dependent code branch on pypy

marscher:


marscher:
I'm not familiar with PyPy, after disabling access to the unsupported memory views, I get this error message:
```
native/python/jpype_python.cpp:226:16: error: ‘Py_Initialize’ was not declared in this scope
```

hydratk:
I tried following changes and the compilation passed ok.
Module jpype can be imported from Python code as well.

**native/python/include/jpype_python.h**
_#ifdef Py_PYTHON_H
#include "PyPy.h"
#endif_

**native/python/jpype_python.cpp**
_#if PY_MAJOR_VERSION >= 3
PyMODINIT_FUNC PyInit__jpype()
#else
PyMODINIT_FUNC init_jpype()
#endif
{
#ifndef Py_PYTHON_H
        Py_Initialize();
#endif
        PyEval_InitThreads();_

But when I call method startJVM, fatal error is raised
Fatal error in cpyext, CPython compatibility layer: converting a PyTupleObject into a W_TupleObject, but found NULLs as items

The cause is located in **native/python/py_class.cpp, method getMethods**.
Maybe there is some gap between tuple creation and item insertion.
See https://bitbucket.org/pypy/pypy/commits/94bf747cee5d/ (code raising fatal error).

_PyObject* res = JPySequence::newTuple((int)mth.size());
JPySequence::setItem(res, i, (PyObject*)ref->data());_

Any idea how to fix it ?

marscher:
one could check for NULL prior calling setItem, if it is built against pypy and do not call then. getMethods returns NULL, if the desired method name is not found within the class. So it looks like there is a wrong call/assumption about presence of methods of this class.

Issue #133, opened 2015-04-09
Reading from Java byte array with np.frombuffer


caspervdw:
I am trying to interface bioformats from python, as part of the PIMS project (https://github.com/soft-matter/pims/pull/144). My experience with JPype up to now is very good, but there is one issue that I hope to optimize.

Bioformats inputs images into a JByte array and separately I can readout the datatype which can be any number. Because we potentially have to readout a lot of them, I want this step to be as fast as possible.

So I read https://github.com/originell/jpype/issues/71 and https://github.com/originell/jpype/pull/73 and came up with a few options:

```
from __future__ import division
import numpy as np
import jpype
jpype.startJVM(jpype.getDefaultJVMPath())

nbytes = 2*1024**2
a = jpype.java.nio.ByteBuffer.allocate(2*1024**2).array()
_source_dtype = '>u4'

def convert_JArray1(a):
    arr = np.array(np.frombuffer(unicode(a), dtype=np.uint16), dtype=np.uint8)
    return np.frombuffer(buffer(arr), dtype=_source_dtype)

def convert_JArray2(a):    
    java_str = jpype.java.lang.String(a, 'ISO-8859-1')
    arr = np.array(np.frombuffer(java_str.toString(), dtype=np.uint16), dtype=np.byte)
    return np.frombuffer(buffer(arr), dtype=_source_dtype)

def convert_JArray3(a):
    return np.frombuffer(np.array(a, dtype=np.byte), dtype=_source_dtype)

%timeit convert_JArray1(a) # 100 loops, best of 3: 9.14 ms per loop
%timeit convert_JArray2(a) # 100 loops, best of 3: 7.9 ms per loop
%timeit convert_JArray3(a) # 1 loops, best of 3: 14 s per loop
```

My questions are:
- am I doing something wrong in the 3rd option?
- the second one works on my Win64 platform; on Travis it fails. Maybe something with endianness but I can't figure it out. Does anyone have a suggestion? (edit: with fails, I mean, the returned array has the wrong shape)

I am using jpype 0.5.7 on py2.7. Thanks in advance for your response!

marscher:
If you use the Numpy interface to get buffers (eg. byte[] arrays from JVM) it should be fast.

In [4]: a = jpype.java.nio.ByteBuffer.allocate(2*1024**2).array()

In [5]: a
Out[5]: <jpype._jarray.byte[] at 0x7f50f00c4e10>

In [6]: %timeit a[:]
[...]

PR #167, opened 2015-12-21
Fix Windows numpy bug

adbull:
The following edit seems to fix issue #133, which causes JPype on Windows to convert Java array slices to lists, even when NumPy support is enabled. I'm not sure why this fix works, but it solves the issue on my machine.

### DIFF
--- a/native/python/jpype_javaarray.cpp
+++ b/native/python/jpype_javaarray.cpp
@@ -134,7 +134,9 @@ PyObject* JPypeJavaArray::getArraySlice(PyObject* self, PyObject* arg)
 		else if (hi > length) hi = length;
 		if (lo > hi) lo = hi;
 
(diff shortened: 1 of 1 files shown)


adbull:
Figured out why this works: in the buggy version, we take a string reference to the typename, which is a member field of a temporary object, and so gets deleted before the call to is_primitive. By allocating the object explicitly, we ensure the string will still exist on the following line.

adbull:
Ignore this, I'll make a new pull request which fixes both getting and setting slices.

justussc:
i missed this fix in PR #164, where i fixed the same problem in setArraySlice a couple of lines below. will add it to PR #164.
btw i am not sure if we can use c++11 features like auto. taking a const reference like in PR #164 is enough to keep the object alive.

adbull:
Sure, looks good to me.

PR #168, opened 2015-12-21
Fix Windows bug ignoring NumPy when getting and setting array slices

adbull:
Bugfix for Issue #133: on Windows, JPype converts array slices to lists even when NumPy is installed. The bug occurs because JPype stores the name of the array object type in a temporary string, which gets deleted before it can be read. Making a local copy of the string fixes the bug.


adbull:
Closed as issue is already addressed in PR #164.

PR #170, opened 2016-01-14
@marscher for president!

originell:
I should have done this a long time ago. @marscher without all your efforts (and those of the other contributors) this project would probably be dead again.

I hope I haven't forgotten anyone in the list. Anyone I missed? If so, speak out or may you forever stay silent ;-)!

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -3,7 +3,7 @@ Authors
 
 The original author: Steve Menard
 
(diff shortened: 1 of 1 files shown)


marscher:
Thank you for your kind words @originell ! I'd accept this if you hand over admin rights on the Appveyor job, so I do not need to bug you (eg. restarting PR jobs now, that we have a working windows build again).

PR #172, opened 2016-02-21
improved overload resolution and support for java8 default methods

justussc:
using jpype one can always force an overload resolution by using the JObject,JLong,... wrappers. but this is tedious. sometimes an overload is added to a class by the compiler (for example synthetic methods for overriding mixed with generics, see the testcase.) in this PR i tried to implement the rules in http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.12.2.5 as far as this is possible at runtime. there are quite some tests around this.
also support for default methods in java 1.8 was added by using getMethods instead of a recursive getDeclaredMethods. the test code (in java) is unfortunately commented out as the travis hosts have only java1.7 by default.

### DIFF
--- a/appveyor/install.ps1
+++ b/appveyor/install.ps1
@@ -198,7 +198,7 @@ function unzipAnt($file, $destination) {
 }
 
 function DownloadAnt() {
-    $url = "http://www.us.apache.org/dist/ant/binaries/apache-ant-1.9.4-bin.zip"
+    $url = "http://www.us.apache.org/dist/ant/binaries/apache-ant-1.9.6-bin.zip"
     $webclient = New-Object System.Net.WebClient
     $filepath = "C:\ant.zip"
 	
--- a/jpype/_jclass.py
+++ b/jpype/_jclass.py
@@ -140,7 +140,7 @@ class _JavaClass(type):
         if name == 'java.lang.Object' or jc.isPrimitive():
             bases.append(object)
         elif not jc.isInterface():
-            bjc = jc.getBaseClass(jc)
+            bjc = jc.getBaseClass()
             bases.append(_getClassFor(bjc))
 
         if _JAVATHROWABLE is not None and jc.isSubclass("java.lang.Throwable"):
--- a/native/common/include/jp_classbase.h
+++ b/native/common/include/jp_classbase.h
@@ -37,7 +37,7 @@ public : // JPType implementation
 		return m_Name;
 	}
 
-	virtual jclass getClass()
+	virtual jclass getClass() const
 	{
 		return (jclass)JPEnv::getJava()->NewLocalRef(m_Class);
 	}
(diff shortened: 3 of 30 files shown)


marscher:
Thank you very much for this major improvement @justussc! I've pushed a fix for ant on Appveyor, but unfortunately I can not retrigger a build of your PR, so can you please add a dummy commit (or fix the trailing the whitespaces in test/jpypetest/array.py)?

@baztian can you please also have a look at this?

justussc:
sure. i just ran array.py through autopep8. but it seems to still have a problem with the ant thing. unfortunately i have never used appveyor so i cannot really help here.

marscher:
Am 21.02.2016 um 17:54 schrieb justussc:
> sure. i just ran array.py through autopep8. but it seems to still have a
> problem with the ant thing. unfortunately i have never used appveyor so
> i cannot really help here.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/originell/jpype/pull/172#issuecomment-186856688>.
>
Appveyor should be fixed in terms of the missing ant version. If you 
rebase on latest master it should at least test something ;)

justussc:
i saw that you pushed this change only to your fork. i cherry picked it from there and it gets further now. but then it fails in with
`powershell ./appveyor/runTestsuite.ps1
...........result code of nosetests: False 
System.IO.FileNotFoundException: Could not find file 'C:\projects\jpype\nosetests.xml'.
`
i don't have a windows box available. but i can try to play around with this in the next days.

marscher:
Am 22.02.2016 um 09:02 schrieb justussc:
> i saw that you pushed this change only to your fork.
ah sorry for that. I thought I pushed this to master.
> i cherry picked it
> from there and it gets further now.
I see you know how to use git :)
  but then it fails in with
> |powershell ./appveyor/runTestsuite.ps1
> ...........result code of nosetests: False
> System.IO.FileNotFoundException: Could not find file
> 'C:\projects\jpype\nosetests.xml'.
This suggests from my experience, that the python interpreter just 
crashed...
Hopefully we could pin this down.

justussc:
thx for the hint. i'll have a look later this week.

justussc:
i managed to get a windows setup (only tried 32bit py2.7 until now) and it looks like also originell/jpype master is crashing already on loading the VM. i did not get to start actually debugging the problem, but maybe you could push your ant fix to master such that one can see, that it is already broken there.

marscher:
Thank you so much for investigating this! I've pushed the ant fix to 
orginell/master

marscher:
The windows tests have unfortunately been broken (I'd guess by myself) for a long time. Three tests are failing but they are unrelated to this changes. So I'm going to merge this now.

justussc:
thanks for merging. earlier this week i think i found one crash in the windows tests. it's actually a bug in  jp_primitivetypes_autogen.cpp where we call reserve on a vector instead of resize. but i'll have a look.

PR #178, opened 2016-03-05
fix appveyor tests and bug in jp_primitivetypes_autogen.cpp

justussc:
all these `{s|g}etArrayRange` functions look suspicious, but here a fix for the currently crashing appveyor tests, as i was already looking at these. 

does anybody know where the code is that was used to autogenerate `jp_primitivetypes_autogen_cpp`?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,13 +17,13 @@ environment:
       PYTHON_ARCH: "64"
       JAVA_HOME: "C:\\Program Files\\Java\\jdk1.8.0"
 
-    - PYTHON: "C:\\Python34_32"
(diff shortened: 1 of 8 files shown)


marscher:
@DevilWolf maybe knows how it was generated. However it has already been changed in some places to add numpy slice assignment/getting on java arrays. This should have been done in a separated unit. However the author (me) admits that he was too lazy to do so.

@justussc Thanks for fixing this ugly bug with memory management, which has been introduced quite some time ago. And all this other stuff (fixing anaconda versions, and the path test thingy). Really appreciated :+1: 

But just a tiny request: could you please replace this resize/reserve statement in the auto-gen unit?

PR #207, opened 2017-02-12
Mark destructor that throws as noexcept(false) for >=C++11

Peter-Levine:
In modern C++, destructors default to `noexcept(true)` and anytime it would throw, it calls `terminate()` instead.  When compiling in C++11 or later, such destructors have to be marked `noexcept(false)` to enable throwing of exceptions.  And compile fails if built with CXXFLAGS="-Werror=terminate".  See [dev-python/jpype-0.6.2 : native/.../jpype.h:30:53: error: throw will always call terminate() [-Werror=terminate]](https://bugs.gentoo.org/show_bug.cgi?id=608920).

### DIFF
--- a/native/common/include/jp_field.h
+++ b/native/common/include/jp_field.h
@@ -38,7 +38,7 @@ public :
 	/**
 	 * destructor
 	 */
-	virtual ~JPField();
+	virtual ~JPField() NO_EXCEPT_FALSE;
 	
 public :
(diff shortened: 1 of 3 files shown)


marscher:
Thank you for this contribution. We should probably enforce the standard for compilation or test against newer versions to catch upcoming problems.

Issue #211, opened 2017-03-03
Multithreading not working


baztian:
Thought #210 should fix #203 as well so I ran a small test baztian@b9469c87f0a819134fc6f409c114c1514f6977c4  (using python 2.7.12). The results were a bit weird though.
I'm getting the following error if I run the complete testsuite.

    # A fatal error has been detected by the Java Runtime Environment:
    #
    #  SIGSEGV (0xb) at pc=0x00007ff32ec7b079, pid=9746, tid=0x00007ff319573700
    #
    # JRE version: Java(TM) SE Runtime Environment (8.0_111-b14) (build 1.8.0_111-b14)
    # Java VM: Java HotSpot(TM) 64-Bit Server VM (25.111-b14 mixed mode linux-amd64 compressed oops)
    # Problematic frame:
    # C  [_jpype.so+0x29079][thread 140682070533888 also had an error]
      JPJavaEnv::NewString(unsigned short const*, int)+0x29
    #
    # Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
    #
    # An error report file with more information is saved as:
    # /home/bbowe/Sources/jpype/hs_err_pid9746.log
    #
    # If you would like to submit a bug report, please visit:
    #   http://bugreport.java.com/bugreport/crash.jsp
    #
    Aborted

The test hangs forever if I run the test directly via `python test/testsuite.py jpypetest.threads`.
The test almost always runs if I uncomment the attachThreadToJVM/detachThreadFromJVM calls. But one time I got the following error:

    INFO:jpypetest.common:Running testsuite using JVM /usr/lib/jvm/java-8-oracle/jre/lib/amd64/server/libjvm.so
    #
    # A fatal error has been detected by the Java Runtime Environment:
    #
    #  SIGSEGV (0xb) at pc=0x00007f54dac61a24, pid=9853, tid=0x00007f54c1696700
    #
    # JRE version: Java(TM) SE Runtime Environment (8.0_111-b14) (build 1.8.0_111-b14)
    # Java VM: Java HotSpot(TM) 64-Bit Server VM (25.111-b14 mixed mode linux-amd64 compressed oops)
    # Problematic frame:
    # C  [_jpype.so+0x43a24]  JPMethod::findOverload(std::vector<HostRef*, std::allocator<HostRef*> >&, bool)+0x84
    #
[...]

PR #217, opened 2017-05-13
stop JPJavaEnv::GetEnv returning (JNIEnv*)NULL

mikea1729:
GetEnv can fail and return a NULL JNIEnv which crashes in many places
downstream.  This seems to happen when threads are thrashing.

Instead, detect JNI_EDETACHED and do AttachCurrentThread()
then retry GetEnv.  If it fails again RAISE(JPypeException).

### DIFF
--- a/native/common/include/jp_javaenv.h
+++ b/native/common/include/jp_javaenv.h
@@ -83,7 +83,7 @@ private :
 	jobject referenceQueue;
 	bool convertStringObjects;
 
-	JNIEnv* getJNIEnv();
+	JNIEnv* getJNIEnv(bool tryAttach=true);
 
 public :	
 	static void load(const string& path);
@@ -132,7 +132,7 @@ public :
 	jthrowable ExceptionOccurred();
 	
 	jint DetachCurrentThread();
-	jint GetEnv(JNIEnv** env);
+	jint GetEnv(JNIEnv** env, bool tryAttach);
 	jint ThrowNew(jclass clazz, const char* msg);
 
 	jint Throw(jthrowable th);
--- a/native/common/jp_javaenv.cpp
+++ b/native/common/jp_javaenv.cpp
@@ -253,6 +253,15 @@ jint JPJavaEnv::GetEnv(JNIEnv** env)
 	// This function must not be put in GOTO_EXTERNAL/RETURN_EXTERNAL because it is called from WITHIN such a block already
 	jint res;
 	res = jvm->functions->GetEnv(jvm, (void**)env, JNI_VERSION_1_2);
+	if(res == JNI_EDETACHED)
+	{
+		AttachCurrentThread();
+		res = jvm->functions->GetEnv(jvm, (void**)env, JNI_VERSION_1_2);
+		if(res != JNI_OK)
+		{
+			RAISE( JPypeException, "Java Subsystem GetEnv() failed");
+		}
+	}
 	return res;
 }
 
--- a/native/common/jp_javaenv.cpp
+++ b/native/common/jp_javaenv.cpp
@@ -45,10 +45,10 @@ JPPlatformAdapter* JPJavaEnv::GetAdapter()
 jint (JNICALL *JPJavaEnv::CreateJVM_Method)(JavaVM **pvm, void **penv, void *args);
 jint (JNICALL *JPJavaEnv::GetCreatedJVMs_Method)(JavaVM **pvm, jsize size, jsize* nVms);
 
-JNIEnv* JPJavaEnv::getJNIEnv()
+JNIEnv* JPJavaEnv::getJNIEnv(bool tryAttach)
 {
 	JNIEnv* env;
-	GetEnv(&env);
+	GetEnv(&env, tryAttach);
 	return env;
 }
 
@@ -141,7 +141,7 @@ void JPJavaEnv::DeleteLocalRef(jobject obj)
 {
 	TRACE_IN("JPJavaEnv::DeleteLocalRef");
 	TRACE1((long)obj);
-	JNIEnv* env = getJNIEnv();
+	JNIEnv* env = getJNIEnv(false); // tryAttach=false
 	if (env != NULL)
 	{
 		env->functions->DeleteLocalRef(env, obj);
@@ -153,7 +153,7 @@ void JPJavaEnv::DeleteGlobalRef(jobject obj)
 {
 	TRACE_IN("JPJavaEnv::DeleteGlobalRef");
 	TRACE1((long)obj);
-	JNIEnv* env = getJNIEnv();
+	JNIEnv* env = getJNIEnv(false); // tryAttach=false
 	if (env != NULL)
 	{
 		env->functions->DeleteGlobalRef(env, obj);
@@ -189,7 +189,7 @@ jobject JPJavaEnv::NewGlobalRef(jobject a0)
 
 bool JPJavaEnv::ExceptionCheck()
 {
-	JNIEnv* env = getJNIEnv();
+	JNIEnv* env = getJNIEnv(false); // tryAttach=false
 	if (env != NULL)
 	{
 		return (env->functions->ExceptionCheck(env) ? true : false);
@@ -205,7 +205,7 @@ void JPJavaEnv::ExceptionDescribe()
 
 void JPJavaEnv::ExceptionClear()
 {
-	JNIEnv* env = getJNIEnv();
+	JNIEnv* env = getJNIEnv(false); // tryAttach=false
 	if (env != NULL)
 	{
 		env->functions->ExceptionClear(env);
@@ -232,7 +232,7 @@ jint JPJavaEnv::AttachCurrentThreadAsDaemon()
 
 bool JPJavaEnv::isThreadAttached()
 {
-	return JPEnv::getJava()->getJNIEnv() != NULL;
+	return JPEnv::getJava()->getJNIEnv(false) != NULL; // tryAttach=false
 }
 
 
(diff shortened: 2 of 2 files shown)


mikea1729:
@marscher Thanks for reviewing.  I saw that there was existing work in #211 and so opted for a surgical change to fix the NULL dereference issue without (hopefully) any additional negative impact.  Given the alternative is a JVM crash, I think a potential memory leak may be preferable in the near term.

I was expecting this would need to be assessed with regard to the larger multithreading fixes and was hoping for some insight from you or @baztian or others who understood the impact better.  Thanks for the SO link.

I didn't explicitly add a DetachCurrentThread() because:
1. I'm naive when it comes to the full scope of the JNI and rest of the codebase. ie. does DetachCurrentThread() invalidate the JNIEnv*?
2. My assumption from a minimal review was it is likely all of the GetEnv() callers are already expecting the thread to be attached, since the return value is unchecked in many places.

So, it seems the options are, add a DetachCurrentThread() in my patch right after retrieving the JNIEnv* (if that's safe), or understand the existing codebase and proposed multithreading changes to make more extensive modifications.  (I may not be in the best position to do that).

Any thoughts or other ideas?

marscher:
On 05/14/2017 03:58 PM, mikea1729 wrote:
> @marscher <https://github.com/marscher> Thanks for reviewing. I saw that there was existing work in #211 <https://github.com/originell/jpype/issues/211> and so
> opted for a surgical change to fix the NULL dereference issue without (hopefully) any additional negative impact. Given the alternative is a JVM crash, I think
> a potential memory leak may be preferable in the near term.
probably you're right on this, but we can know, how many threads will be created in this place. In the IPython case, there will be
a one thread for each autocomplete. In a long session this could potentially mess up things.
> 
> I was expecting this would need to be assessed with regard to the larger multithreading fixes and was hoping for some insight from you or @baztian
> <https://github.com/baztian> or others who understood the impact better. Thanks for the SO link.
> 
> I didn't explicitly add a DetachCurrentThread() because:
> 
>  1. I'm naive when it comes to the full scope of the JNI and rest of the codebase. ie. does DetachCurrentThread() invalidate the JNIEnv*?
>  2. My assumption from a minimal review was it is likely all of the GetEnv() callers are already expecting the thread to be attached, since the return value is
>     unchecked in many places.
If we would detach every time in this place, we would also add a lot of overhead. There was another Stack thread about this, but I cant find right now.

I think we should solve the issue of ipython crashing for ipython only (eg. detecting if it is a IPython environment and do the attachment automatically (maybe
without detaching, since ipython is only used for prototyping and not for production). If one really works in a multithreaded environment, he or she has to take
care
of attaching/detaching manually.
> 
> So, it seems the options are, add a DetachCurrentThread() in my patch right after retrieving the JNIEnv* (if that's safe), or understand the existing codebase
> and proposed multithreading changes to make more extensive modifications. (I may not be in the best position to do that).
> 
> Any thoughts or other ideas?
> 
Actually the best thing would be some kind of RAII in C++, but as you said this would involve a lot of refactoring and I do not have the time to do it.

Once I've started to do the same thing on Pythons side by using a context manager, which calls detach automatically, when the block goes out of scope (with even
more overhead involved). I've not measured the impact on performance though. JPypes testsuite isn't a good tool for that, because it does not perform enough
calls. A real world application would be better suited.

mikea1729:
Looks like you may not have seen the two "**UPDATE**" sections since I added them after the initial comment.

Detaching the thread within the same code segment isn't an option.  It would invalidate the `JNIEnv*` which is the entire reason for calling the function in the first place.

And, it looks like *always* attaching actually isn't entirely correct either.  But to be clear, the current patch only attaches when the thread is already not attached, so not "double attaching".  There's at least one place where the result is checked for NULL to determine if an existing thread is attached.  But I think it's probably just that one place or a sufficiently small number that could be handled in a correct way without a massive change.  I still contend that in all of the other cases the additional overhead for attaching should be 0 since the caller is expecting the thread to be attached for correct operation.  So attaching should be acceptable.  And, since there is an expectation the thread is already attached, then that implies there should be already a place where it would be detached.  If not, well that's another bug that already exists.  So I'm not fixing that, but not making it worse either, and still fixing a case where the process would crash.

I do agree that some RAII mechanism is the right long term approach, but there doesn't seem to be enough time to do that now, I wonder if a suitable stop-gap approach is possible?  I don't like the idea of fixing it just for IPython.  Detecting that the caller is IPython sounds inaccurate at best.  It's certainly no better than someone just using my fork if it is in fact "not production", and potentially worse in the general case.

When I have some more time, I'll see if I can adjust the patch to only `AttachCurrentThread()` in the cases where 1) the thread is detached (as in current patch) and 2) the caller expects it to be attached (current patch is too general).  So if I do that right, there should be no extra overhead.  I would still never explicitly detach but as I said above this shouldn't cause a leak for cases I'm adding the attach.  I'll update the PR with an additional commit, so the modification is obvious.  The PR can always be squashed later.

mikea1729:
Looks like my clock is out of sync.  Committed 9480610 after I made the previous comment (I had more time than I thought).  Let me know what you think.

mikea1729:
It's been a few weeks.  Thought I'd ping to see what the status is on accepting this PR.

marscher:
Your observation that jnienv handles are not checked in most places seems correct. I think multi-threading was really intended to be handled from Pythons side (Attach-/DetachThread), since it is hard or even impossible to handle this properly in the native code (without hurting performance).

I agree that your patch solves a null deref, but introduces a memleak, because the thread will never be detached again. Probably one can even crash the jvm, by creating tons of Python threads (denial of service). So I can not accept it in this state for the sake of security and stability.

Maybe you're right and we should detach the thread again immediately (and sacrifice some performance). It is at least better to do this in the native rather than the Python code. We can do some benchmarking to see how large the impact really is.

mikea1729:
> Your observation that jnienv handles are not checked in most places seems correct. I think multi-threading was really intended to be handled from Pythons side (Attach-/DetachThread), since it is hard or even impossible to handle this properly in the native code (without hurting performance).

I'm not sure I'm following you here about the performance implications.  If the code needs the JNIEnv, it must be attached as per the JNI API.  Code that doesn't need it wouldn't attach.  I see no performance issue with native code making this determination and doing it if and only if it is necessary.

> I agree that your patch solves a null deref, but introduces a memleak, because the thread will never be detached again. Probably one can even crash the jvm, by creating tons of Python threads (denial of service). So I can not accept it in this state for the sake of security and stability.

Here we disagree.  I've tried to state this multiple times and either I'm being unclear or you're just not understanding my comments.  Let me break this down...again:

> So I can not accept it in this state for the sake of security and stability

The basis for this statement is false.  Here's why: As you restated above I create a memleak where there was a null deref.  The memory leak is ONLY in cases where the system would have crashed.  Read that sentence again.  Therefore, the patch makes this codebase MORE stable, not less.  I'm not sure how many different ways I can say this.

> Probably one can even crash the jvm, by creating tons of Python threads (denial of service)

Actually, without my patch the system crashes because of the null deref when there are many threads.  WITH my patch the system will NOT crash because of a null deref, and MAY later crash due to memleak, MAY.  This is a strictly better outcome with the patch than without.

Finally, I contend that as implemented, I do not *actually* leak any more memory in practice then without my patch applied, because the only times I attach, is when the code is already expecting the thread to be attached.  In these cases, if the rest of the codebase is correct, something will detach the thread.  If not, then I've created a memory leak where there was otherwise a complete crash.

> Maybe you're right and we should detach the thread again immediately (and sacrifice some performance). It is at least better to do this in the native rather than the Python code. We can do some benchmarking to see how large the impact really is.

I did not suggest detaching "right away".  The detach would have to happen much later keeping more state and require refactoring the code.  If you read my original explanation more closely, you'd see that you cannot detach the thread immediately after obtaining the JNIEnv* because detach invalidates it.

I would really like to see this PR accepted.  I believe it makes the codebase more robust, and I don't see any evidence otherwise.  At this point I've spent more time explaining this patch then writing it.  If you can actually explain why you think a memory leak is worse than a crashing application I'd love to hear it and maybe I'd be willing to spend more of my time implenting a fix.  But as I see it I've done more than enough as a casual user to make this project better.

marscher:
We all like our PRs to merged without much hassle (since this seems to be your first one, I stay calm and friendly). However this simple script crashes already after 10 iterations on your "fixing" branch:
``` python
import threading
import jpype

jpype.startJVM(jpype.get_default_jvm_path())

def threaded_access():
    jpype.java.lang.String("foo")

for i in range(1000):
    threading.Thread(target=threaded_access).start()
    print(i)
```
So leaking resources is not really an option. Sorry. The original author of the software intended to handle multiple threads by the provided Python api, because every other case would involve "tracking" the life time of the Python threads in order to be able to detach them from the JNI properly (https://jpype.readthedocs.io/en/latest/userguide.html#threading). Maybe one can handle this tracking on Pythons side as well (eg. by using the threading context manager or more cleverly, have a hook function which calls attach and tracks the lifetime. There is a function in the threading module to achieve that https://docs.python.org/2/library/threading.html#threading.setprofile)

Any ways thank you very much for your precious time and this discussion. I think it was valuable (the threading hook idea seems promising).

mikea1729:
For the record, your test included above fails on 8d479b9 (tag: v0.6.2) and 07845b0 (master when I started this PR), and much of the time it fails AT the bug that this resolves.  If anything, you've proved this PR is necessary (but not sufficient...).

After applying my patch, it fails elsewhere.  I spent some time trying to resolve the remaining problem(s) your test script finds, but I think there are multiple related concurrent accesses of state within JPMethod and/or JPMethodOverload and I don't have the cycles to reverse engineer the codebase to figure out a reasonable fix.  Plus the lack of synchronization makes me think there may be problems throughout, and time may be better spent on a more thorough redesign for multithreading.

PR #220, opened 2017-05-29
Cygwin

Thrameos:
Minor modifications to compile under cygwin.


marscher:
Hey, thanks for this contribution. Could you please add Cygwin to the Appveyor build matrix? It seems like it is pre-installed [there](https://www.appveyor.com/docs/build-environment/#mingw-msys-cygwin), but I have no experience in using it. 
Why did you have to declare __int64? It seems to be like a compiler internal thing and it not used in jpype itself. 
And I'm curious why you overwrite the Python internals macro  PyUnicode_FromFormat with the non-unicode aware function. As far as I know mingw can handle unicode as well.

setup.py
```diff
@@ -74,6 +74,11 @@ def find_sources():
     platform_specific['define_macros'] = [('WIN32', 1)]
     jni_md_platform = 'win32'
 
+elif sys.platform == 'cygwin':
```
marscher: this could have been added as an or clause to windows, right?

Thrameos: sys.platform on cygwin returns cygwin, but the compiler defines __WIN32__.  As most of the windows stuff uses __WIN32__. There has to be some modifications.

Thrameos:
Still struggling to get the tests working.  The primary issue is that java requires windows style classpaths and cygwin only produces posix.   There is no "good"  way (as in method that does not require external packages) to make the conversion.   I am trying to add a more portable method, but will require some development time.

Thrameos:
All tests now pass on cygwin under python3.6 and python2.7.   It required a new getClassPath method to handle conversions properly.  One test required modification as Java uses windows paths not cygwin.  I added a condition to fix the problem.

setup.py
Thrameos: I will defer on the PyUnicode,  I was struggling to find a set of patterns that worked with the compiler.  It seems like that was a left over.  I will try to get Appveyor working, but I don't know much about it.

Thrameos: Looking at the PyUnicode.  I didn't change it.   I was just cleaning up the warnings that I was getting about redefined macros.   It is likely that the statement that followed should have been conditional.

Thrameos: Still waiting for additional review.

Thrameos:
Something went pretty drastically wrong on the pull request.   I think some testing code regarding a cleanup which was not intended was in it.  Please advise how to fix so that I can resubmit.

Thrameos:
Okay this one has been superseded.

PR #224, opened 2017-06-13
Imports

Thrameos:
This is a proposed system to support safer imports of java classes.  Unfortunately it only works for Python 3.6 as the support hooks for 2.7 are not good enough.   It prevents accidental use of non-imported classes which improves type safety and makes python code more robust to code refactors of the java code.  We have been using this in module for a month, but have not used it on many alternative platforms.  It supports some slight exotic features such as placing python code into netbeans projects.  We are working on directly being able to store a python package to support the use of java classes in the jar, but that support is not ready for distribution.   I have included the basic imports system with hooks and the corresponding documentation.  Testing scripts will still need to be developed.

### DIFF
--- a/jpype/imports.py
+++ b/jpype/imports.py
@@ -1,0 +1,314 @@
+# -*- coding: utf-8 -*-
+# *****************************************************************************
+#   Copyright 2017 Karl Einar Nelson
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#          http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# *****************************************************************************
+
+# Optional jpype module to support:
+#   import <java_pkg> [ as <name> ]
+#   import <java_pkg>.<java_class> [ as <name> ]
+#   from <java_pkg> import <java_class>[,<java_class>*]
+#   from <java_pkg> import <java_class> [ as <name> ]
+#   from <java_pkg>.<java_class> import <java_static> [ as <name> ]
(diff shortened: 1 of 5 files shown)


jpype/imports.py
```diff
@@ -0,0 +1,314 @@
...
+def _initialize():
+    global _exportTypes
+    global _initialized 
+    global _modifier
+    _initialized = True
+    _JMethod = type(_JClass('java.lang.Class').forName)
+    _modifier = _JClass('java.lang.reflect.Modifier')
+    _exportTypes = (property, _JavaClass, _JImport, _JMethod)
+
+_jinit(_initialize)
```
marscher: If your code is only python 3.6 compatible, you should not register this by default as it would then be used by all different Python versions.

marscher:
thanks this looks really useful. Can you add a simple test as well? eg. the examples you've given in the documentation (great!) would be absolutely sufficient.

jpype/imports.py
Thrameos: It is currently placed as an optional module.  To use it "import jpype; import jpype.imports".  If the support hooks are missing it would give a ImportError.  If you have an alternative mechanism I would be happy to transform it.

marscher:
Ah alright :+1:

I also can not spot any Py-3.6 exclusive feature usage. Maybe we can 
make it compatible easily?

Thrameos:
Working on the test cases now.   Just need to figure out the assertion system.   The 3.x dependency is on the importlib.  I tried for a day to figure out a similar hook, but couldn't find anything similar without a monkey patch.   I can't be sure if it really requires 3.6 as the hooks appear at different times during 3.  I suspect that after 3.4 it will work.

Thrameos:
I have merged my local production copy back into this branch and added tests.  Currently the 5 new test show as pass on the jpype.imports system.  Unfortunately, my production system is cygwin, so I am having to back port changes to keep a clean copy until we get a clean cygwin test pattern.  If you have some way to get importlib functionality on 2.7 (barring a monkey patch), I can backport the jpype.imports code.

Thrameos:
As suspected it passes checks on 3.4 on.  I had to remove nightly as it has crash that is not related to my patches.  Please review and advise.  I can remove any functionality that you find too exotic to support long term.

marscher:
LGTM 👍

PR #226, opened 2017-06-15
Boxed

Thrameos:
- Convert boxed types to python types automatically.   Uses type hacking to make boxed types inherit from python types.  Has some major changes to implicit rules so may have some edge cases.
- Minor patch to split MTRACE and TRACE so that can used diff between python2 and python3 to resolve differences in type resolution.  
- Does not replace _JWrapper.   
- Breaks one test case by allowing java.lang.Float(1) and java.lang.Double(1) to succeed.
- Exports JavaObject and JavaClass as stub types for use with isinstance().

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,5 +45,6 @@ install:
   - ant -f test/build.xml
 
 script:
+  - jdk_switcher use oraclejdk8
   - python -c "import jpype"
   - python test/testsuite.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,5 @@ install:
   - ant -f test/build.xml
 
 script:
-  - jdk_switcher use oraclejdk8
   - python -c "import jpype"
   - python test/testsuite.py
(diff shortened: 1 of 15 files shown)


Thrameos:
Hmm.  One set of tests failed, but not on my development system.  I will need to find a system to replicate before I can advance.

Thrameos:
It appears that we are issuing the wrong exception on jp_primitives for range errors on conversion.  We are giving TypeError, but python issues OverflowError under similar situations.

marscher:
please rebase on master instead of merging it into the branch.

marscher:
I dont really understand what you are doing. The history is even more messed up now.

Thrameos:
Okay I hope the history it back.  I had to erase it back through time.  I may need another day before these patches are ready again.  Lets see if the tests pass.

Thrameos:
No sure how, but my history fixes seem to have taken the testing system out.

Thrameos:
This set of patches should be ready.  The test system caught up a few hours later.

marscher:
Thanks again. Do you mind adding a note about this cool new feature to doc/changelog.rst?

Thrameos:
Sure I can add it.  Should I do the same for "with".

marscher:
LGTM

PR #227, opened 2017-06-16
Support for varargs.

Thrameos:
Adds support for varargs on static, methods and constructors.  Resolved a fatal bug in the main branch with argument type cache.   Hopefully this is the last major patch I need for my production.


marscher:
thank you so much!

PR #228, opened 2017-06-17
Cygwin + Appveyor patterns

Thrameos:


Thrameos:
Getting closer.  Need to get the nosetest to run properly.

Thrameos:
Not sure why nose is not running the tests.   Once I can figure that out then all that is required is to add the windows back into the matrix.

.gitattributes
```diff
@@ -0,0 +1,4 @@
+* text=auto
+
+*.sh text eol=lf
```
marscher: this will enforce lf, right? Actually I'd like to keep the previous setting, because this setting will show all files as changed, but they are not.

appveyor.yml
```diff
@@ -4,49 +4,85 @@ environment:
     # /E:ON and /V:ON options are not enabled in the batch script intepreter
     # See: http://stackoverflow.com/a/13751649/163740
     CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\appveyor\\run_with_env.cmd"
-    ANT_HOME: C:\ant
+    ANT_HOME: "C:\\ProgramData\\chocolatey\\lib\\ant\\apache-ant-1.10.1"
+    NUMPY_: "numpy x.x"
 
   matrix:
+    - PYTHON: "python3"
```
marscher: so the cygwin matrix will use cygwins python?

.gitattributes
Thrameos: We can drop that change.  I was suspecting the reason for an error I was getting was that even though I have LF only .sh files into my system it was coming out as CRLF on the test matrix.  

We need to make the change to enforced at some point.  But you are correct that  having it show full file conflicts is a problem right now.  I hate the "EOL hell" where files don't appear in a specified type each time it is pulled as it adds additional variables to testing.  The correct EOL settings require certain global settings appear on PC and MAC users machines and there is no way to guarantee that those settings are proper in the test matrix.  Thus long term we are better to fix all the files.

appveyor.yml
Thrameos: Yes this is correct.  It took a while for me to get a pattern that gave cygwin the right version and kept windows the same.  It turned out the problem was that I already had mock preinstalled on my system and nosetests had a silent failure.  I kept ripping my hair out because I could not replicate the problem.

I still have a problem with the proposed patch.  I would like to autodetect the version of ant and the cygwin python version numbers.  Cygwin appends very specific version numbers like nosetests-3.6  which is great from the prospective that it we know what we have gotten, but means we have to change the version whenever cygwin updates.  I believe I can autodetect with a cygcheck | grep | sed pattern, but I was too tired to put the change in and wait 30 minutes for appveyor to run again.

marscher:
On 06/19/2017 01:18 PM, Karl Nelson wrote:
> *@Thrameos* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In .gitattributes
> <https://github.com/originell/jpype/pull/228#discussion_r122680268>:
> 
>> @@ -0,0 +1,4 @@
> +* text=auto
> +
> +*.sh text eol=lf
> 
> We can drop that change. I was suspecting the reason for an error I was
> getting was that even though I have LF only .sh files into my system it
> was coming out as CRLF on the test matrix.
> 
> We need to make the change to enforced at some point. But you are
> correct that having it show full file conflicts is a problem right now.
> I hate the "EOL hell" where files don't appear in a specified type each
> time it is pulled as it adds additional variables to testing. The
> correct EOL settings require certain global settings appear on PC and
> MAC users machines and there is no way to guarantee that those settings
> are proper in the test matrix. Thus long term we are better to fix all
> the files.
That is all true. I'd also prefer to do the conversion at some point. I
think git can also handle the automatic conversion on the client side.
Eg. files, which contain clrf will be lf locally, but are committed with
the original line ending.

marscher:
On 06/19/2017 01:24 PM, Karl Nelson wrote:
> *@Thrameos* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In appveyor.yml
> <https://github.com/originell/jpype/pull/228#discussion_r122681284>:
> 
>>  
>    matrix:
> +    - PYTHON: "python3"
> 
> Yes this is correct. It took a while for me to get a pattern that gave
> cygwin the right version and kept windows the same. It turned out the
> problem was that I already had mock preinstalled on my system and
> nosetests had a silent failure. I kept ripping my hair out because I
> could not replicate the problem.
sorry for all the hassle. Actually we should not use nosetests any more,
since it is a dead project.
> 
> I still have a problem with the proposed patch. I would like to
> autodetect the version of ant and the cygwin python version numbers.
> Cygwin appends very specific version numbers like nosetests-3.6 which is
> great from the prospective that it we know what we have gotten, but
> means we have to change the version whenever cygwin updates. I believe I
> can autodetect with a cygcheck | grep | sed pattern, but I was too tired
> to put the change in and wait 30 minutes for appveyor to run again.
are the binaries for python 2 and 3 in same directory?

Thrameos:
Yeah.  Cygwin installs everything in /usr/bin so that means nosetests, pip, python, and easy_install all have a version appended.  They put python, python3, and python3.6, python2, python2.7 links, but the subpackages like nose are all linked with the full version (ie easy_install-3.6).  

Do you have a replacement for nosetests we should be using?  I am not a great python programmer so I don't know the story on the tools.  I mainly use a mixed MATLAB/java system but they are still back on JDK 7 and when the JDK 7 link stopped working I suddenly got a bunch of complaints from users that I had to deal with.  Switching our tools to JDK 8 with Python/Jpype seems like a better option that trying to work around the upgrade MATLAB to JDK8.

marscher:
On 06/19/2017 02:16 PM, Karl Nelson wrote:
> Yeah. Cygwin installs everything in /usr/bin so that means nosetests,
> pip, python, and easy_install all have a version appended. They put
> python, python3, and python3.6, python2, python2.7 links, but the
> subpackages like nose are all linked with the full version (ie
> easy_install-3.6).
> 
> Do you have a replacement for nosetests we should be using?
pytest is state of the art, IMHO.

> I am not a
> great python programmer so I don't know the story on the tools. I mainly
> use a mixed MATLAB/java system but they are still back on JDK 7 and when
> the JDK 7 link stopped working I suddenly got a bunch of complaints from
> users that I had to deal with. Switching our tools to JDK 8 with
> Python/Jpype seems like a better option that trying to work around the
> upgrade MATLAB to JDK8.
So jdk7 is not deprecated, but not used by your users any more?

Thrameos:
Oracle disabled the download link for jdk 7 a few months back.  It was out of service life for like 2 years, but the disable was recent.  Unfortunately, my group of summer students all showed up and the computer support said something like "you need to download JDK 7 because we haven't yet set up a security policy for JDK 8 yet."  That would all be fine except the only people who can download JDK 7 is the person with the support contract for Oracle.  Hence, the lead project programmer is burning 5-8 hours each time a student arrives on the phone with computer support trying to explain why they need JDK 8 for half the tools (netbeans) and JDK 7 also (for Matlab), and that the computer support needs to get the JDK7 because the download link points to a nice "we don't support JDK 7 anymore link."  If I am going to burn that amount of time I may as well help out Jpype.

[http://www.oracle.com/technetwork/java/javase/downloads/jdk7-downloads-1880260.html](url)

Bottom line, Oracle is not happy that developers are way out of date on their Java version so it is time to get up to date.  Heck Java 9 is about to ship and some people still have JRE 6 on their machines.

[http://www.infoworld.com/article/3176593/java/java-9-gets-a-release-date-july-27.html](url)

Thrameos:
Oracle has a very interesting definition of deprecated.  They never deprecate something unless it is so broken that it must not be used.  Thus they have gone to a strongly discouraged by making it so you can't get it without a support contract.  The last one available is the one travis is using is one with a known bug.  Release 131 is what I am testing locally but it is not available publicly.  My computer group still has not pushed JDK 7 r141 which is the latest.

Thrameos:
Sorry this is going to get a bit messy.  I need to do a production run and will be merging the head back into this branch.  Hopefully, I don't mess up the pull request in the process.

marscher:
what a mess :P Are there some files which only contains line ending changes? Please omit those. Also I just wonder why you took the time to change the copyright of jpypetests init file, did you re-wrote it from scratch?
What is the actual change you wanted to propose in the first place? Can't you just rebase on master?

Thrameos:
Okay let me checkout and repatch on a clean branch.  I needed  a run last night and had to pull branches in. on a merge.

Thrameos:
Looking at my records the test/jpypetest/__init__.py was copied from another file which I use for autopulling modules.  I added the copyright to the file that was copied.  I didn't see the existing file.   It appears the existing file was empty.

PR #229, opened 2017-06-18
Bulletproof

Thrameos:
This patch deals with segfaults that can happen if a jpype call is made prior to starting the JVM.   Each entry point was tested to see if it properly issues a RuntimeError if that entry point could have produced a segfault.  Test bench for verifying was added, but as it must not be run with the other tests we may need to improve the runTestSuite to handle more than one test set.  

Also contains a minor patch for Boolean types as there were cases of accidental promotion which would be incorrect for a strongly typed language such as Java.  

Like the other patches this is an independent patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,9 @@ sudo: false
 addons:
   apt:
     packages:
+    - oracle-java8-installer
     - ant
+
 env:
   # one build with and without numpy support (opt numpy out)
   - NUMPY="--install-option=--disable-numpy"
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,13 @@ python:
   #dist: trusty
 sudo: false
 
+jdk:
+  - oraclejdk8
+
 addons:
   apt:
     packages:
-    - oracle-java8-installer
+    - oracle-java8-set-default
     - ant
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ python:
   - 'pypy3'
   - "nightly"
 
+dist: trusty
 sudo: false
 
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ python:
   - "nightly"
 
   #dist: trusty
-sudo: false
+sudo: required
 
 jdk:
   - oraclejdk8
@@ -52,5 +52,6 @@ install:
   - ant -f test/build.xml
 
 script:
+  - jdk_switcher use oraclejdk8
   - python -c "import jpype"
   - python test/testsuite.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,11 +47,12 @@ matrix:
       env: NUMPY="--install-option=--disable-numpy"
 
 install:
+  - ls /usr/lib/jvm/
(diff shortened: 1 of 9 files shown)


Thrameos:
Darn.  Appears that some kind of race condition was introduced.  Any clue which function is likely a problem?

Thrameos:
It appears this bug is coming from a problem in the jvm we are testing with (7.0_76-b13).  It may disappear if we move to a later jvm on the travis build.
http://bugs.java.com/view_bug.do?bug_id=7052197

Switching to travis build trusty which is more current.  Otherwise we need to add an "apt get" for oracle-8 in the build pattern.

marscher:
since we do not have a stacktrace, we can really say which bug this is? I just wonder then why it only crashed on Py34. I'd prefer not changing the image.
Can one actually override the fault handler of the jvm by preloading the libc debug library?
eg in travis.yml:
matrix:
python=3.4
LD_PRELOAD=/usr/lib/debug/libc.so

marscher:
BTW: i really like this fault-checking 👍

Thrameos:
As a race condition I think that we could repeat the test many times and not get the failure to appear in the same one.  Everything I could find indicated that it was a problem in Jdk 7 which was fixed.  We could test this by switching back to original image and see if the race condition repeats.

Thrameos:
As suspected it is a race condition in JDK 7.  Thus the problem does not replicate.  Lets try the alternative solution and see if we can upgrade the JDK.  JDK 7 is no longer support by Oracle and thus not something we should be building against.  It would be fine if we tested versus 7 release 131 but many people can't get that one.  I only have it because my site is an oracle support client.

marscher:
Ah I didn't know that jdk7 is deprecated, so it is better to use the recent one.

Thrameos:
Well that was a failure.  It installed jdk-8 but still ran the tests with 7.   Unfortunately, travis has different images by language.  We are using the python one which is correct, but the java is not kept up to date on it.

Thrameos:
Crud.  Still didn't switch JRE.   I guess I should try some of the other patterns.

https://github.com/travis-ci/travis-ci/issues/5161

marscher:
maybe you would have to invoke update-alternatives --config java or something like that?

Thrameos:
Does that need sudo switched on?

marscher:
On 06/19/2017 02:23 PM, Karl Nelson wrote:
> Does that need sudo switched on?
locally it does, but I don't know for travis.

Thrameos:
Lets try some random patterns from...

[https://docs.travis-ci.com/user/languages/java/](url)

Thrameos:
Nope.  Those patterns don't work for python image.  Next...
[https://github.com/travis-ci/travis-cookbooks/tree/master/community-cookbooks/java](url)

Thrameos:
Okay going to have to give up on this for now.    None of the patterns appear to do the very simple thing of updating the JAVA_HOME.  The one race condition failure did not reappear in 5 runs in a row so I am pretty sure it consistent with the know Java 7 "just in time" compiler bug on shutdown.  (As in recompile a method just in time for the jvm to close so we will segfault).  I added you as a contributor to the fork in case you wanted to try to poke at it.

PR #231, opened 2017-06-18
Closed behavior for objects and arrays.

Thrameos:
In order to reduce the problems with typos and moved members, I recommend overriding the `__setattr__` on classes and objects.  The change still allows unrestricted access to set methods and private variables that begin with underscore, but will issue `AttributeError` if other members are set.   This prevents a major source of errors while still allowing augmentation of java objects for frameworks.

This patch is independent of all others submitted.


Thrameos:
#$!^$*!@!  Messed up the rebase again!  Okay need to wait until I figure out how to rebase back to master properly.

Thrameos:
Okay should be properly rebased again.

PR #232, opened 2017-06-19
Patch to fix Class.forName()

Thrameos:
This was a fairly difficult modification as customizers do not support changing fields.  I introduced an enhanced customizer as best I could.  This allows direct adjustment of fields.   This addresses the defective forName()

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,10 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 - **Next version - unreleased**
 
+  - java.lang.Class<>.forName() will now return the java.lang.Class.  
+    Work arounds for requiring the class loader are no longer needed.
+    Customizers now support customization of static members.
+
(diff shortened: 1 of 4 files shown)


Thrameos:
Shelving this one for now until I get proper support of java.lang.Class.  No sense in allowing something which does not yet point to the correct object type.

marscher:
I'm kinda puzzled by the fact that removing the topological sorting for member lookups did not break anything. I think this was introduced to properly support multi-inheritance.

Thrameos:
I don't think that I changed the sorting or at least not intentionally.  The mro got moved from 202 in the previous to 204 in the revised.    The move up (relative to the customizers) was so that the customizer has access to all of the fields.  I can move it back to the original if you feel that it should be an unconditional setting.

marscher:
ah my fault :)

Thrameos:
Okay.  Though I still recommend holding this patch until I can finish work so that forName() returns a java.lang.Class.

PR #238, opened 2017-06-24
Support for "with" using java.io.Closeable.

Thrameos:
One of my team members requested better integration between java resources.   The java "try-with-resources" can be mapped to the python's "with."   There was also a minor patch as two locations in _jp_method.cpp were emitting the same error message which made it hard to disambiguate which condition was hit.


test/jpypetest/autoclose.py
```diff
@@ -0,0 +1,35 @@
...
+class AutoCloseCase(common.JPypeTestCase):
+
+    def setUp(self):
+        common.JPypeTestCase.setUp(self)
+
+    def testCloseable(self):
+        AutoClose = jpype.JClass("jpype.closeable.AutoClose")
+        self.assertFalse(AutoClose.closed)
+        with AutoClose() as myFile:
+            myFile.print_("hello")
```
marscher: does that also handle exceptions in the context?

marscher:
That's a very useful addition, thanks!

test/jpypetest/autoclose.py
Thrameos: I will need to check to make sure it has the correct behavior.   The intent is to properly handle python or java exceptions but I realize there may be a case in which close generates an IOException which could cause issues for python.  Lets hold off until I can make a test case.

Thrameos:
Okay the issue here is that I am not sure that I can get the same behavior for python 2 and python 3.  

In python 3, I can properly raise an exception this is generated in the close method.  And I can conditionally eat a nested exception to prevent problems.  

In python 2, they don't always clear the last exception frame, thus it would be context dependent as to whether or not the exception would be issued.  If I eat it then it is always eaten and thus no way to detect an exception on close.  It I do issue it, it will clobber the first exception which is always bad. 

The best I can do is have proper behavior in Python3 and eat all exception on close in python 2.  As exceptions on close are strongly discouraged this would be best solution.  Maybe something with a stronger understanding of python can give a better answer.  I can really understand how to properly handle the nested otherwise I would let Python3 issue the nested as that is more indicative of what is going on.  On the other hand I think all that people care about is that close was attempted rather than if it succeeds.

Thrameos:
I am somewhat perplexed on the python 2.6.   I am tempted to just put a skip on the two failed tests as I see no mechanism to handled the rather rare case of close issuing an exception.

PR #244, opened 2017-07-07
Exception during multithreaded execution of proxies

matrixx567:
During the execution of a registered proxy in a multithreaded application I get the following exception.

```
java.lang.RuntimeException: Python exception thrown: AttributeError: 'NoneType' object has no attribute 'getName'
 	at jpype.JPypeInvocationHandler.hostInvoke(Native Method)
 	at jpype.JPypeInvocationHandler.invoke(JPypeInvocationHandler.java:10)
	at com.sun.proxy.$Proxy18.handleMessage(Unknown Source)
```

The proxy is called from multiple threads using Java's ExecutorService with a fixed pool size. If I only run a single thread at once Jpype works perfectly but if I change to more than one thread execution at a time Jpype crashes.

During the execution of some specific test cases the exception turns into an access violation.
```
#
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x000007fef86a23c5, pid=7180, tid=1056
#
# JRE version: Java(TM) SE Runtime Environment (8.0_91-b15) (build 1.8.0_91-b15)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.91-b15 mixed mode windows-amd64 compressed oops)
# Problematic frame:
# C  [_jpype.cp35-win_amd64.pyd+0x23c5]
#
# Failed to write core dump. Minidumps are not enabled by default on client versions of Windows
#
# C:\DATA\Software\jpype\hs_err_pid7180.log
#
#   http://bugreport.java.com/bugreport/crash.jsp
```

## Solution

The return value of the called proxy method can't be converted back to a Java object.
The constructors of the corresponding Java class were null.

**Call Stack**
- jpproxy.cpp (95): canConvertToJava
- jp_primitivetypes.cpp (19): newInstance
- jp_class.cpp (469): invoke_constructors
**Error:** m_constructors in JPCLass was null

The problem can be simply solved by completely loading the Java class before it is registered to Jpype. (swapping `postLoad()` and register)

```java
JPClass* res = new JPClass(name, cls);
// Finish loading it
res->postLoad();
//Register it here before we do anything else
javaClassMap[name.getSimpleName()] = res;
```

**System settings:**
Python 3.5
OS: Windows 7 , 64 bit
vm_info: Java HotSpot(TM) 64-Bit Server VM (25.91-b15) for windows-amd64 JRE (1.8.0_91-b15)

### DIFF
--- a/native/common/jp_typemanager.cpp
+++ b/native/common/jp_typemanager.cpp
@@ -84,12 +84,12 @@ JPClass* findClass(const JPTypeName& name)
 	cleaner.addLocal(cls);
 
 	JPClass* res = new JPClass(name, cls);
-	
+	// Finish loading it
+	res->postLoad();
 	// Register it here before we do anything else
 	javaClassMap[name.getSimpleName()] = res;
-	
-	// Finish loading it
-	res->postLoad();		
+
+
 
 	return res;
 	TRACE_OUT;
--- a/native/common/jp_typemanager.cpp
+++ b/native/common/jp_typemanager.cpp
@@ -12,8 +12,8 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-   
-*****************************************************************************/   
+
+*****************************************************************************/
 #include <jpype.h>
 
 namespace {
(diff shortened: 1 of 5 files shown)


marscher:
hey man, very sophisticated test! Thanks for taking care of this. Should have merged this one quicker, sorry for that. It should also be time for packing this stuff soon.

Issue #247, opened 2017-07-20
PropertyCustomizer


Thrameos:
While working on modifying to expose the class structure in python I ran into a curious bug.  The java.lang.String had an inaccessible property java.lang.String.class.  It is not clear where it came from as it did not appear from either the static_fields nor the members in _jclass.py.  After a bit of debugging I tracked it to the the PropertyCustomizer.  For some reason it is registering getClass as a Bean pattern and attempting to produce a property.  The second thing that is odd is that same property does not appear on java.lang.Object.  The same was true of Exception.  Tracking it through code the PropertyCustomizer is only called on classes after a certain point.   

Based on this analysis I think we should consider the following corrective actions.

1.  Move property.initialize() higher in the initialize() routine as it appears to be the case that we may create classes that need to be customized before the customizer is registered.  
2.  Add a keyword check in the PropertyCustomizer or alternatively move the keyword check lower in the _jclass so that we guarantee that keywords are cleaned up.
3.  Add an explicit check for "class" in the PropertyCustomizer as we really don't want to expose "getClass" as "class".   
4.  Define a long term project to improve the Customizer registration such that Customizers are applied to existing classes so that order of registration and object creation does not affect the class members.  I have run into this a few times in which I was adding a math to double[] and found that which library got loaded affected whether or not my customizer was called.  As I noted when patching forName, there seems to be a number of enhancements needed for customizers to be useful for all cases.  I am not sure I have a good enough picture of the requirements yet.
5. If we don't plan work of 4, then we should consider checking the order of operations in initialize().   I believe that PropertyCustomizer is the only currently defined Customizer that affects all classes.   Though I think that 4 would be the most robust method.
6.  The keywords are being explicitly defines in _pykeywords but the list does not match the list in "import keywords"  things like "True" and "False".   Consider if we would be better to use the pythons keywords.  The downside is that if 2.7 and 3.* have different keywords then we will have a different name binding.  The upside is that we won't have to maintain the _pykeywords if something is added.  Was "import keywords" introduced after jpype was started or was there an explicit reason to maintain a separate keyword list.

I have started a patch for items 1-3.  We need to decide a plan for 4 and 5.   I will need more info for 6.

Thrameos:
This items has mostly been incorporated in main branch.  Further customizer work is planned.

Issue #237, opened 2017-06-21
Expose Class object through getClass()


Thrameos:
I am going to take a shot at removing the automatic conversion of java.lang.Class objects to Python class objects.  The code that I am working on makes pretty heavy use of java attributes and reflection to access data.   This means getting into the java.lang.Class directly.   Also with access to the Varargs we can now directly pull methods from the java.lang.Class which means one can work around problematic method overloads.   A "side" benefit is with access to the class one can get private and protected members through reflection and flip the accessability flag if debugging at the protected access level is needed.

Currently, jpype wraps outgoing class objects as the Python class.   This means there is no way to access the java class proper.   The wrapping looks redundant because if one has an python/java object, they can access the python object class from the dictionary.  Having a second method provides no purpose that I can see.

The alternative would be to augment the __javaclass__ which is a _jpype.JavaClass.  However, that means more startup overhead and more C++ code.    I looked at code which would allow Python code to plug into the _jpype.JavaClass but that means adding an additional base class with an added dict and more setup.   

The main effect is that the meaning of forName() and getClass() would change.  There would be a jpype specific method in the java.lang.Class to get the Python class if that path is needed (Class.getPythonClass()).  

The modification is to remove _class from the preload of jp_typemanager.cpp.  However, there are several internal methods that are expecting the Python type.  (JObject() in particular)  Those methods segfaulted after the change, but should not be hard to run down.

Thrameos:
I have made reasonable progress on this task, but I am hitting a place that will need some design input.   In Java classes can get their ```Class<Foo>``` from a class with ```Foo.class``` and from an object with ```foo.getClass()```.   In the wrapper we have ```Foo.__javaclass__``` which gives the internal wrapper class and ```Foo.__class__``` which gives the python class.   I wish to add support to get the ```Class<Foo>``` from the python class and the python object.   The object version is pretty easy as it should obviously be ```foo.getClass()```.   Getting the python class to produce the class is more challenging.  We can't expose it as ```Foo.class``` because of a keyword conflict.   It is difficult to expose as ```Foo.getClass()``` because we would need a different method called when called from the object and class version.  Thus the most logical alternative is ```Foo.class_```.   However having 3 different class fields in each object seems like a recipe for user confusion.  

Does it make sense for a Class to have
1. ```Foo.__javaclass__``` the jpype wrapper object (private)
2. ```Foo.__class__``` - standard python method to get the python type.
3. ```Foo.class_``` - access to java.lang.Class<Foo>
[...]

PR #248, opened 2017-07-21
Support for java.lang.Class<>

Thrameos:
* This patch exposes java.lang.Class as Foo.class_ for types and foo.getClass() for objects.  
* It allows full use of reflection to access annotations, private fields and private methods.   
* It reorders the customizers so that PropertyCustomizer is applied to java.lang.Object and java.lang.Exception.
* It unifies a large number of mostly copy and paste methods under setResource.
* It removes the redundant class JPClassType.
* It patches a warning on comparison of unsigned to signed
* It simplifies a handful of comparison tests.

See issues #247 and #237 for details.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,17 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 - **Next version - unreleased**
 
+  - Support of java.lang.Class<> 
+    - java.lang.Object().getClass() on Java objects returns a java.lang.Class 
+      rather than the Python class
+    - java.lang.Object().__class__ on Java objects returns the python class 
+      as do all python objects
+    - java.lang.Object.class_ maps to the java statement 'java.lang.Object.class' and
+      returns the java.lang.Class<java.lang.Object> 
+    - java.lang.Class supports reflection methods
+    - private fields and methods can be accessed via reflection
+    - annotations are avaiable via reflection
+
   - Java objects and arrays will not accept setattr unless the 
     attribute corresponds to a java method or field whith 
     the exception of private attributes that begin with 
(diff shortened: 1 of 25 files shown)


marscher:
I like this unification of setResources very much! Tell me, when it is mergable.

Thrameos:
It should be good to go.   There is the basic question if the mapping of java ```gov.foo.Bar.class``` should map to  python ```gov.foo.Bar.class_```.  The tree of python classes, java classes, and internal wrappers is overly complex but I don't see a good way to simplify it as of yet.  I can redo java.lang.Class<> bindings piece if you have another direction that you think would be better.  

Other than philosophy on project direction, everything in this patch is working and I put the test unit in place.  I needed access to annotations and the ability to call private methods.  

My next project is one of the following (in no particular order)
* Update the forName patch to be work with head
* Rework customizers so that they can be applied to classes after the class definition was created.  (Current have a method of patching the initializers in our start up script, but my remote users are more likely to trip on this one.)
* Try to get a dynamic class implementation that would allow python to implement a java class or interface.  This would potentially expose bugs in the threading as it would then be possible to launch a thread from within java.  I have the specifications of how to pull this off from the JNI layer using Proxy hooks.
* Figure out what the string conversion code is doing so that java strings and python strings are the same like the boxed types.  I stumbled through some of that code on this consolidation, but I haven't figured out how it connects.
* Finish the date patch.  (My coworker is pulling his hair out trying to get java dates to play nice with matplotlib.)
* Take other shot at seeing if I can get the stale pypy branch back in.  I took a shot but the ubuntu kernel bug prevented progress.
* Debug what is wrong with the nightly build.  (I think this is a simple change to the start of the module.)

Brandon would also like to know if when we have reached a milestone for a release.   My list of features are still pretty long, but we have made a fair about of progress.  I will be out for the next month on travel.  But if you have a release date target, I can go through and make sure the changelog reflects the changes from the last.

PR #250, opened 2017-07-23
Work on fixing leak.

Thrameos:
This is a work in progress pull request.  Please disregard until all tests are working.  All accessors that use resources now require a cleaner to prevent making leaks.  However, it does not pass tests because at least one of the fixed sections required leaky behavior.

### DIFF
--- a/native/python/py_method.cpp
+++ b/native/python/py_method.cpp
@@ -85,6 +85,7 @@ PyJPMethod* PyJPMethod::alloc(JPMethod* m)
 
 PyObject* PyJPMethod::__call__(PyObject* o, PyObject* args, PyObject* kwargs)
 {
+	JPLocalFrame frame(32);
 	TRACE_IN("PyJPMethod::__call__");
 	try {
 		PyJPMethod* self = (PyJPMethod*)o;
@@ -128,6 +129,7 @@ void PyJPMethod::__dealloc__(PyObject* o)
 
 PyObject* PyJPMethod::__str__(PyObject* o)
 {
+	JPLocalFrame frame;
 	PyJPMethod* self = (PyJPMethod*)o;
 	stringstream sout;
 
@@ -138,6 +140,7 @@ PyObject* PyJPMethod::__str__(PyObject* o)
 
 PyObject* PyJPMethod::isBeanAccessor(PyObject* o, PyObject* arg)
 {
+	JPLocalFrame frame;
 	try {
 		PyJPMethod* self = (PyJPMethod*)o;
 
@@ -156,6 +159,7 @@ PyObject* PyJPMethod::isBeanAccessor(PyObject* o, PyObject* arg)
 
 PyObject* PyJPMethod::isBeanMutator(PyObject* o, PyObject* arg)
 {
+	JPLocalFrame frame;
 	try {
 		PyJPMethod* self = (PyJPMethod*)o;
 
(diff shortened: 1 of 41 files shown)


Thrameos:
Getting closer.  Seems to have a problem with Proxy.

Thrameos:
More progress again, but now it is working locally and failing on the test builds.  Thus I will need to replicate it before making progress.   It is all the usual issue that the more aggressive the cleanup the more likely that something which is needed was cleaned.

marscher:
I really suggest adding a test script, which returns != 0 on a leak and
do a git bisect to find the bug.

Thrameos:
Unfortunately, my cygwin development platform does not support valgrind or any other leak checker.  So I can't do a git bisect.

marscher:
The guy who reported the leak issue posted a snippet, which you could
use to check this systematically. So you dont need valgrind at all.

Thrameos:
I will give it a try.  Though I suspect that his leak is not the same.   The leak he is reporting was in the python reference counting as far as I can tell.  I am trying to patch a leak in the java global reference table.   I did a full review of the java counting and found several places where I may have introduced a bug due to problems with obeying conventions.  Thus I am just trying to change the conventions to make it hard to leak.  Unfortunately, when I plugged that leak I find that portions of the existing code did not run because they were using the java reference leaks.

Thrameos:
Do you show his leak as failing on the master?   I ran it and is shows no issues.

Thrameos:
Okay this is a slightly different approach.  The problem is that the current code is using the JPCleaner to try to keep the local references from building up.  But that depends on every thing that produces a LocalReference to make it to the cleaner.   A single missing marking means that resources pile up because we are calling from the jni side and local references only get cleaned when returning to java from a native call.  

The alternative approach is to wrap all calls from python with PushLocalFrame and PopLocalFrame.  This makes returning to python the same as returning to java and thus all local references are cleaned appropriately.   I took a short at today, but still need to deal with the edge cases where the cleaner was affecting the control flow by delaying destruction.   Hopefully this will solve all long term resource leaks.

Thrameos:
Making progress.  Two checks are failing because the resource is being killed before they are given global resources.  Still some issues on various compilers.  On the plus side I have run a number of leak tests and am not finding leak issues anymore.  Will try to get the leak checks to be python tests, but getting the condition to pass and fail properly will take some work.

Thrameos:
It appears that JValue(String) was depending on leaking resources.   I have fixed that problem.   I will try to make some tests and then deal with the compiler issues that I may have introduced.

marscher:
big thanks!

Thrameos:
This should be getting close.   Finding a test that doesn't false alarm is challenging.  I added some leaks and it caught them properly.  

One side effect is that execution speed has improved quite a bit.  The leaking resources cause a big hit because DeleteLocalRef is a linear search function.  So the more that leaked the slower it ran.   I believe that is fixed.

Thrameos:
Okay this one should be ready for review.

Thrameos:
Will work on resolving the conflicts shortly.

Thrameos:
Okay should be back to a working state.  There is a minor chance that a global reference leak reappears due to a bad edit in the rebase.  However, I audited it again and don't see any unbalanced calls except for the global resources.   I believe it to be safe to apply.

native/python/py_method.cpp
```diff
@@ -190,6 +195,7 @@ PyObject* PyJPMethod::getName(PyObject* o, PyObject* arg)
 
 PyObject* PyJPMethod::matchReport(PyObject* o, PyObject* args)
 {
+	JPLocalFrame frame;
```
marscher: here we have both a LocalFrame and a Cleaner, ain't that redundant?

marscher: I see... :)

marscher:
Thank you so much for all your effort. I like the push/pop local frame style much more than the jpcleaner approach, since it avoids the need to track individual objects. LGTM

PR #253, opened 2017-07-28
Cleanup

Thrameos:
Placeholder for work to clean up primitive table and improve performance on conversion of types.  This will be rebased to the master after the leak match is merged.   Please ignore until then.


Thrameos:
This one is going to take some time to untangle.  Remains on hold until the head catches up to leak.

marscher:
mhm, now I've broken cygwin compilation. The compiler just returns 127 and no proper message. Do you know whats going on here @Thrameos?

Thrameos:
Not to worry.  This was a low priority clean up patch.  I will clean up the errors once I get everything back to head.  We don't need this one for a release.

Thrameos:
Making some progress on this one.  Need to finish some of the clean up work before it is ready.  It is now properly rebased to the current master.

Thrameos:
The failure makes no sense at all.  If seems more like something changing with the appveyor environment rather than a change in our code.   I will need to track the problem down.  It does not replicate on my system.

Thrameos:
This branch is currently dead due to a second refactorization attempt.

PR #257, opened 2017-08-24
Support for java.lang.Iterable

robbert-ef:
Added support for interface `java.lang.Iterable` (JDK >= 1.5) to make them iterable in Python.


Thrameos:
This one is now corrected in devel branch.  We should close this request when we are ready for the next release.

Thrameos:
Fixed in release 0.7

PR #264, opened 2017-09-22
Describe exception when tracing is enabled

tcalmant:
When the TRACING compile flag is enabled, use the JNI ``ExceptionDescribe()`` method to have more information about what happens.


marscher:
Thanks!

Issue #241, opened 2017-06-25
Safety for 32b and 64b on Windows


Thrameos:
I am not sure on licensing issues from stackoverflow, but can we add some code like this to the startJVM for windows?

Ported from https://stackoverflow.com/questions/1345632/determine-if-an-executable-or-library-is-32-or-64-bits-on-windows

```python

import jpype
import struct
import sys

def _checkJVMArch(jvmPath):
    IMAGE_FILE_MACHINE_I386=332
    IMAGE_FILE_MACHINE_IA64=512
    IMAGE_FILE_MACHINE_AMD64=34404

    is64 = sys.maxsize > 2**32
    with open(jvmPath, "rb") as f:
        s=f.read(2)
        if s!=b"MZ":
            raise RuntimeException("JVM not valid")
        else:
            f.seek(60)
            s=f.read(4)
            header_offset=struct.unpack("<L", s)[0]
            f.seek(header_offset+4)
            s=f.read(2)
            machine=struct.unpack("<H", s)[0]

            if machine==IMAGE_FILE_MACHINE_I386:
                if is64:
                    raise RuntimeError("JVM mismatch, python is 64 bit and JVM is 32 bit.")
            elif machine==IMAGE_FILE_MACHINE_IA64 or machine==IMAGE_FILE_MACHINE_AMD64:
                if not is64:
                    raise RuntimeError("JVM mismatch, python is 32 bit and JVM is 64 bit.")
            else:
                raise RuntimeError("Unable to deterime JVM Type")

# Success if correct
_checkJVMArch(jpype.getDefaultJVMPath())

# Fail if wrong
_checkJVMArch(r"c:\Program Files (x86)\Java\jre7\bin\client\jvm.dll")

```

marscher:
thats a cool hack we should introduce also for linux (bsd?) and osx.

PR #270, opened 2017-12-01
Added support to check for architecture differences betweeen jvm and python

Thrameos:
Patch for #241 

I still don't have access to make the patterns for Mac or Linux, but the problem bit me on Windows so I figured it was time to put the check in place.  

This patch adds a new call to the jvm finder that checks if the jvm is valid prior to returning it.  The check procedure can throw a JVMNotSupportedException if the jvm and the python architecture don't match so that the finder will continue to look for a valid jvm.  It will now properly throw a failure if the found jvm is not supported but it doesn't report the exception string that would be useful for debugging.  

We can add checks for other architectures and have them implement the check method.  

This patch is a work in progress as I have not checked out the other issues that were causing appveyor failures.

### DIFF
--- a/jpype/_jvmfinder.py
+++ b/jpype/_jvmfinder.py
@@ -110,6 +110,14 @@ class JVMFinder(object):
                         yield path
                         break
 
+    def check(self, jvm):
+        """
+        Check if the jvm is valid for this architecture.
+
+        This method should be overriden for each architecture.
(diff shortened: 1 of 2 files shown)


marscher:
Thanks for that. The Appveyor failure is due to a redefintion of jlong (jpype.h and shipped jni.h have different typedefs). I wonder why we need to define jlong in jpype.h in the first place.

PR #272, opened 2017-12-04
Ignore unrecognized

Thrameos:
Change in default behavior to force unrecognized options to issue an exception.

### DIFF
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -60,14 +60,17 @@ def _initialize():
 def isJVMStarted() :
     return _jpype.isStarted()
 
-def startJVM(jvm, *args):
+def startJVM(jvm, *args, ignoreUnrecognized=False):
     """
     Starts a Java Virtual Machine
 
-    :param jvm:  Path to the jvm library file (libjvm.so, jvm.dll, ...)
-    :param args: Arguments to give to the JVM
+    Args:
+      jvm (str):  Path to the jvm library file (libjvm.so, jvm.dll, ...)
+      *args (str[]): Arguments to give to the JVM
+      ignoreUnrecognized (Optional[bool]): option to jvm to return an 
+        error if a jvm argument is invalid.  (Default False)
     """
-    _jpype.startup(jvm, tuple(args), True)
+    _jpype.startup(jvm, tuple(args), ignoreUnrecognized)
     _initialize()
 
     # start the reference daemon thread
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -4,6 +4,9 @@ Changelog
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
 - **Next version - unreleased**
+  - Changed startJVM() to report errors on unrecognized options by default.
+    Use `ignoreUnrecognized=True` to get previous behavior.
+
   - Java reference counting has been converted to use JNI 
     PushLocalFrame/PopLocalFrame.  Several resource leaks
     were removed.


marscher:
the new signatures syntax is only supported by Python-3. If we put the keyword argument before args, it will break the api.

Thrameos:
Oh python2 why do you have to treat me so badly.   Is this any better?

marscher:
Can we test this? Eg. in a subprocess, because we can only start the jvm once per process. If this involves too much time right, now we can post-pone it, but it would be nice having this.

Thrameos:
I ran each of the cases by hand for my machine, but I am not sure how to make this into a test because we don't have anything that creates a process then tests.  I suppose it is possible, but I don't know the pattern.

jpype/_core.py
```diff
@@ -80,7 +80,7 @@ def startJVM(jvm, *args, **kwargs):
     # Classpath handling
     args=list(args)
     if 'classpath' in kwargs:
-        args.append('-Dclasspath=%s'%(kwargs['classpath']))
+        args.append('-Djava.class.path=%s'%(kwargs['classpath']))
```
marscher: 👍

marscher:
I added a test as PR in your fork, can you merge this?

Thrameos:
Sure.   But my computer is out for a few days with a broken fan.

Thrameos:
Should jvm=None also set the classpath to getClassPath()?  I was concerned that may cause problems with existing (though I couldn't find a case other than `startJVM(None, '-Djava.class.path=foo'))` but it would make `jpype.startJVM()` do the right thing in most cases.

marscher:
you mean in case class path is also None?

Thrameos:
Yeah.  I was hoping to get to something like

```python
def startJVM(jvm=None, *args, **kwargs):
    """
    Starts a Java Virtual Machine.  Typical usage
    startJVM(getDefaultJVMPath(), classpath=getClassPath())

    Args:
      jvm (str):  Path to the jvm library file (libjvm.so, jvm.dll, ...)
        default=None, will use jpype.get_default_jvm_path() and default classpath (jpype.getClassPath())
      *args (str[]): Arguments to give to the JVM
      classpath (Optional[string]): set the classpath for the jvm.
        This will override any classpath supplied in the arguments
        list.
      ignoreUnrecognized (Optional[bool]): option to jvm to ignore
        invalid jvm arguments.  (Default False)
    """
    if jvm is None:
        jvm = get_default_jvm_path()
       classpath = getClassPath()

```

Thrameos:
BTW which is the preferred method `get_default_class_path` or `getDefaultClassPath`?  We have both, but I assumed the later was preferred as it matched the rest of the tools.

marscher:
On 08.12.2017 20:04, Karl Nelson wrote:
> BTW which is the preferred method `get_default_class_path` or `getDefaultClassPath`?  We have both, but I assumed the later was preferred as it matched the rest of the tools.
> 

actually camelCase is Java style, but have it in a lot of spots. 
Preferably we would convert these some day.

You should also check if the class path is None or not set, before 
overwriting the argument.

marscher:
Ain't it easier for the user if the class path is a list of paths and we concatenate it internally with os.path.path_sep?

marscher:
Ah you've done that already in _classpath 👍

Thrameos:
Yes.  I had to add an API for that because cygwin has the horrible situation where the os.path.path_sep does not work for the jvm.  As it hasn't been in a release version, you may want to review and apply any changes needed.

Thrameos:
>  You should also check if the class path is None or not set, before overwriting the argument.

I am a bit fuzzy on how to accomplish this.   I guess we need to make sure that our code satisfies each of the use cases properly.

Case 1:
`startJVM()` - user gave no options, thus we should use default classpath and default jvm.

Case 2:
`startJVM(None, args)` - User deliberately gave no jvm.  This was previously invalid so no existing code has it.  We would have to scan if the class path was set before overriding.  This is a bit of a pain to search the list.  

Case 3:
`startJVM(classpath="foo")`  - the user is wants the default jvm but has a classpath in mind.   Thus we would need to check if classpath is None before overriding.

Case 4:
`startJVM(None, args, classpath="foo")` - Okay this should fall back to case 3 logic.

Reasonable?

Thrameos:
Okay I ran it through a torture test.  But I need to fold it into the test system.

```python

import jpype
from jpype import java

# Default JVM and default Classpath
#jpype.startJVM()
#jpype.startJVM(None)
#jpype.startJVM(None,'-Dfred=sally')

# Default JVM and Specified classpath
#jpype.startJVM(None, '-Djava.class.path=foo')
#jpype.startJVM(classpath="foo")
#jpype.startJVM(None, classpath="foo")
#jpype.startJVM(None, '-Djava.class.path=bar', classpath="foo")
#jpype.startJVM(None, '-Djava.class.path=bar') # gave bar
#jpype.startJVM(None, '-Djava.class.path=bar', classpath=None) # gave bar

# Default JVM and No classpath (explicitly)
#jpype.startJVM(None, classpath=None)
#jpype.startJVM(classpath=None)

# Specified JVM and specified Classpath
#jpype.startJVM(jpype.getDefaultJVMPath(), classpath="foo")
#jpype.startJVM(jpype.getDefaultJVMPath(), '-Djava.class.path=bar', classpath="foo")
#jpype.startJVM(jpype.getDefaultJVMPath(), '-Djava.class.path=bar') # gave bar
#jpype.startJVM(jpype.getDefaultJVMPath(), '-Djava.class.path=bar', classpath=None) # gave bar

print(jpype.java.lang.System.getProperty('java.class.path'))
```

Thrameos:
Rebased to master.  WIP.  Will be squashed before ready for consideration.

ConstantineLignos:
A project I work on ran into an issue where we had to make our own build of JPype just to set `ignoreUnrecognized=False`, so this MR would've been great for us. Is this likely to be merged soon? Is there anything I can do to help?

Thrameos:
I believe this was sitting here because the testing suite showed a fail (likely due to other problems) and I had indicated that I was planning to try to augment the test bench to make sure all of the patterns gave a defined and reasonable behavior.  If you would like to help you can freshen the pull request by rebasing it to the head master and updating the test bench.   It just fell off the radar because noone else had needed it.

PR #282, opened 2018-02-06
Fix bug in the check architecture command.

Thrameos:
There was a case in which the None was passed to the check architecture resulting in a failure.  Rather than forcing all check architectures to be robust against None I added a check for None before calling the check.


marscher:
good spot! Thanks.

PR #284, opened 2018-02-16
Refactoring to support pickle.

Thrameos:
This is a work in progress pull, that is not ready for integration.  I am trying to add support for pickling a java object using serialization.  The key requirement is that pickle needs to be able to find `jpype._jclass.<JavaClassName>` as an attribute of the module so that it can find the custom pickle `__getstate__` and `__setstate__`.  But as our modules are created on the fly this does not work.  I pushed together concepts from `JPackage` and the `_CLASSES` dictionary to create a `jpype._jclass.classes` directory object which hold a cache of all the known class types and is capable of performing the lookup as needed.  However, I don't have time to complete the debugging to get the custom `__getstate__` and `__setstate__`, so that will need to wait for my next time window.   

Trying to push things around to get the pickle functions called, showed that a bunch of stuff we are manually loading into the tables would be fine to simply place in the base classes of each of the respective types.  I likely have broken a bunch of stuff in the refactor by doing so.  Nothing is yet tested on Python 2 which may have be the driving force for some of the existing design.  If things look like they are heading in a bad direction, please feel free to comment.

### DIFF
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -14,16 +14,19 @@
 #   limitations under the License.
 #
 #*****************************************************************************
-import collections
-import sys
+import collections as _collections
+import sys as _sys
 
 import _jpype
 from . import _jclass
 from . import _jwrapper
 
-if sys.version > '3':
-    unicode = str
-    irange = range
+if _sys.version > '3':
+    _unicode = str
+    _irange = range
+else:
+    _unicode = uncode
+    _irange = irange
 
 _CLASSES = {}
 _CUSTOMIZERS = []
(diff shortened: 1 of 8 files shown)


Thrameos:
I was able to work in a feature that I have needed for a while.  Currently we can only register a customizer prior to the creation of a class.  Customizers registered afterward would not ever happen.  This lead to the very unhappy situation that using customizers was very order dependent.  Create in the wrong order and the code would suddenly stop working.  This patch fixes that defect.  Still not able to pickle the java classes, but at least some useful improvements were made.

Thrameos:
I will push in a revised copy.  Most tests pass now.  Just issues with the proxy.   Still trying to get pickle to load.  The save routine is pushing a valid state for objects.  But I will need to get all the types fixed.  Arrays may be more of an issue.

Thrameos:
Hmm, Not sure how this pull picked up the Java 1.8 tests I was working on.   I will try to separate them because those are two distinct issues.  I don't yet know how to get Java 1.8 on travis.

Thrameos:
Putting this one back in the pickle jar for now.  Another attempt will be made aimed for release 0.8

Issue #285, opened 2018-02-17
JArray for JLong going very wrong


Thrameos:
I was trying to wrap up the modifications that I made to get pickle working but I was hitting a series of errors so bad the jvm crashed.  I tried a number of methods and eventually rolled back to the master figuring I could find it by rolling forward to locate the error.  Only I got pretty much the same issue going back to the master.

```
jpype.JArray(jpype.JLong,1)([1,2,3,4,5,6,7])[:]
=> array([1, 0, 2, 0, 3, 0, 4])
```

The slice operator is copying the longs into memory which are ints.  The array returned is half the elements (the other half clobbered some piece of memory).  It just so happened that it hit something vital that lead the crash.  We have a test for JLong conversion but it appears to not catch the problem.  

I scanned the patches for anything can could have changed the corresponding section of code, and turned up nothing.  I am now unsure this path ever worked properly.  It may be a windows issue but it would be good to verify.

Thrameos:
Well that would do it....

```

NPY_LONG¶

    Equivalent to either NPY_INT or NPY_LONGLONG, depending on the platform.
```

```
PyObject* JPLongType::getArrayRangeToSequence(jarray a, int lo, int hi) {
    return getSlice<jlong>(a, lo, hi, NPY_LONG, PyLong_FromLong);
}
```

marscher:
So there is a lengths mismatch between jlong and npy_long on 32 bit?

marscher:
I have to admit, that I tested this code only on Linux and hoped my unit tests are strong enough to cover problems on other platforms too. Can you provide a test, which fails on your platform please?

Thrameos:
The current test fails.  But fails so hard the the system segfault 10 tests later.  I will send the patch.  Just changing to NPY_INT64 and adding the extra forward decl fixed it.

Thrameos:
What I do not understand is how it could have ever passed.  But it appears to have worked before.  I will have to check if when the code or test changed.

Thrameos:
The test code is 2 years since the last change and the bug is 4 years since last change.  I know the test passed when I started working on the project thus it must be able to pass incorrectly under some situation.  But I have no clue how that could have ever been the case.  Alternatively something in numpy changed that caused the issue.

I am removing all code that used NPY_LONG and switching to NPY_INT64.  

Pre patch
```
$ python3 -m nose test.jpypetest
[...]

PR #286, opened 2018-02-20
Fix for errors on windows platform with long array.

Thrameos:
Fix for issue #285


marscher:
Thanks! It makes more sense to use the long type of java in this case.

Issue #274, opened 2017-12-13
Java 8 lambdas do not work


Thrameos:
Okay an interesting flaw was found when we tried to return functions that gave BiConsumer<> instances.  If we produced an anonymous instance it worked properly, but if we returned a lambda the resulting Python class produced an error.  Unfortunately the test bench is currently expected to compile with 1.7 so I can't yet produce tests that should the issue.   

I will try to revise the tests to move to 1.8.  Once that is complete I can demonstrate the problem which should lead to a pull with a correction.

Thrameos:
Sorry real life is still interfering.  Will be on this one once if falls back into critical path on my project.

Thrameos:
Completed as a branch for #334

PR #287, opened 2018-02-21
Work on lambda bug fix

Thrameos:
This is a work in progress pull working to correct Issue #274.  The initial pull will replicate the bug and hopefully test the CI changes to get to 1.8.


marscher:
Thanks! Just a question, wouldn't changing the source/target comapt hinder us to test with older java versions?
At the moment the situation is that we test a lot of python versions, but only one or two hard coded java versions. This is kind of suboptimal. I should change the build matrix at least on travis to include multiple java versions. Eg. set language to java and use miniconda to drag in multiple Python versions. However miniconda oldest Python release supported is 2.7. I don't think there are much people left stuck to 2.6, but we never know.

Thrameos:
Yes.  I am a bit worried about that.  However, Java 1.7 is now no longer available from Oracle.  It went past service life and then was removed from releases with security patches.  It can still be obtained with a service contract.  However, I have a service contract and have found it challenging obtain because outdated releases are only given to the site service coordinator. The question is should we be testing against a platform which is no longer available?

More broadly it would be great if our test matrix supported different versions of java.  I suspect we would need a different build.xml for each platform.  However I am not a travis build expert.  Even if we switch to 1.8 for now,  we once 1.9 is released we are just going to be back into this boat again.

Now that I have this lambda bug replicated I thought I could find the spot that was giving the problem, but the diagnostics I am getting from the exception are minimal.  Hopefully I can find the source.

Thrameos:
Okay so the problem is happening in the JPTypeManager.  For some reason it is attempting to find the return type class by name and lambdas don't have valid names.  I think there was a flaw in the C++ layer that was assuming the one can't directly get the java class object from a object type and thus it is forced to look it up by name.  If we bypass the name lookup from by using the fact that many of  the jtype are infact polymorphic I may be able to dodge the problem.  Unfortunately, it is a bit of a heavy lift as the C++ layer depends on some of these assumptions.

marscher:
There is openjdk, which is pretty much independent from Oracle, isn't it?
Maybe there is a clever way to have only one build.xml. Multiple files are harder to handle.

Thrameos:
We could give parameters to ant and push it in command line.  `ant -DJAVA_VERSION=1.8`   It is an okay solution, but will make it a bit harder to call directly as it will expect the variable to be passed  in.   The harder problem will be making ant and python skip those tests that aren't valid for the Java version.

marscher:
On 21.02.2018 20:29, Karl Nelson wrote:
> Okay so the problem is happening in the JPTypeManager.  For some reason it is attempting to find the return type class by name and lambdas don't have valid names.  I think there was a flaw in the C++ layer that was assuming the one can't directly get the java class object from a object type and thus it is forced to look it up by name.  If we bypass the name lookup from by using the fact that many of  the jtype are infact polymorphic I may be able to dodge the problem.  Unfortunately, it is a bit of a heavy lift as the C++ layer depends on some of these assumptions.
> 

I think for the same reason inner classes are not supported. Probably 
one should rewrite the whole cpp layer with pybind11 and design it 
properly with future Java features in mind.

marscher:
On 21.02.2018 20:33, Karl Nelson wrote:
> We could give parameters to ant and push it in command line.  `ant -DJAVA_VERSION=1.8`   It is an okay solution, but will make it a bit harder to call directly as it will expect the variable to be passed  in.   The harder problem will be making ant and python skip those tests that aren't valid for the Java version.
> 
> 

I think setting JAVA_HOME correctly will fix the java version in ant

Thrameos:
Okay I moved the test harness to be version dependent so that we can add tests that require a specific version.  The ant pattern now looks at the java version so that it builds the tests for the current version and any prior.  And the tests now can have skipping based on the version.  However, we have a problem in that the skip is being evaluated before the JVM so we are getting the wrong version string.  Once that is fixed the test harness code you were working on should be able to run the full test matrix.

Thrameos:
The unit test skipping on java version now works.  Now if I could figure out how to get lambdas to work I would be all set.

marscher:
I also think, that the hashing method seems to be the best choice here.

Thrameos:
I am still trying to figure out why the JP types don't match the tree in JNI.
[https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/types.html](url)

A lot of stuff with the class is messed up because of this.

marscher:
Can you be specific on which types are not matching?

Thrameos:
In JNI jclasses are in fact object instances.  Thus you can operate on them as objects to access their methods.  JP has two separate trees and thus there are lots of times functions assume that you can only operate the class methods or the object methods.  It makes for a bunch of unnecessary helpers like JPTypeName.   Arrays instances as also objects and can be acted as specializations of objects.  But they are treated as a completely separate type, thus the paths for working with objects and arrays diverges which I often find has no difference other than the naming, or something which could have just been a specialization.   

JPYPE C++ Class Tree
image

JNI Class Tree
image

Class is on the bottom of one tree and doesn't connect at all with object.  Arrays are on parallel levels rather than below the type they come from.  Base classes exist but most code is using the concrete types.  There is 1-1 correspondence between jclass, JPTypeName, and JPType but the relationship is spread all over the hierarchy.  Things needlessly covert from strings to names to jclass to JPClass and back and forth just to to handle the miss factorization.  

Don't worry this is just my usual rant when I look down at this level of code.   There is just a lot of needless code that would have gone away if the tree were designed with more with JNI inheritance model in mind.  It is like the memory management section that made its own frame system when jni already provided one.  There was nothing particularly wrong with spinning a memory model if we need it, but it was clear that the original author wasn't experienced enough with jni to know about the built in tool so a lot of reinvention happened.  (Please no offense, I did the same thing when I wrote a JNI bridge class library back in 1995).  It just takes a while to remove reinventions and straighten out the tree especially when something is ingrained like we have here.    I have fixed this a lot at the python level where we can now work with Objects, Bridge Classes and Java Classes, but to do it I had to go back through the tables using forname.  This despite the fact that I have the jclass and the jclass is already a jobject.  

Thus everytime I go to this level I try to fix at least part of the model, but usually find there were too many assumptions resting on some concept.  95% of it can be fixed quickly, but there was one special case required tear your hair out level rework that prevents moving forward.   The Lambda fix is the same, one special case where the tree is insisting on passing a JPTypeName were a jclass would be fine, but every object implements it as a virtual method, the number of changes to get that one case will take 100 times more than everything else.

Thrameos:
Do I need also mention that there is hardly a line of comments in the headers to tell me what is the difference between something like getName() and getObjectType()?  😡

Thrameos:
For example of why something easy is really hard to fix.

```
                if (simpleName == "java.lang.Object")
                {
                        if (JPEnv::getHost()->isInt(obj))
                        {
                                JPTypeName tname = JPTypeName::fromType(JPTypeName::_int);
                                JPType* t = JPTypeManager::getType(tname);
                                res.l = frame.keep(t->convertToJavaObject(obj));
                                return res;
                        }
```

Here we want to convert an int to a boxed type.  So we take an enum and put it through something to get the typename.

```
JPTypeName JPTypeName::fromType(JPTypeName::ETypes t)
{
        return fromSimple(nativeTypes[t].c_str());
}
```

Okay so it turned the enum to "int" then called fromSimple...

```
JPTypeName JPTypeName::fromSimple(const char* name)
{
        string simple = name;
        string componentName = simple;
        ... blah... blah ... blah  (huge long function)
}
```

Okay so now was have a JPTypeName which contains a string "int". So now we just need the type converter....

```
JPType* getType(const JPTypeName& t)
{
        TRACE_IN("JPTypeManager::getType");
        TRACE1(t.getSimpleName());
        map<JPTypeName::ETypes, JPType*>::iterator it = typeMap.find(t.getType());

        if (it != typeMap.end())
        {
                return it->second;
        }
        ...
        TRACE_OUT;
}
```

Okay so we went from type name "int" to fetch a predefined class JPIntType instance.  

We went through 3 layers just to get a predefined class instance using 3 tables of string lookups just to get a predefined constant.  That is a lot of layers of an onion to trace, debug, and maintain.  Not to mention very inefficient.  But this is the level of tracing I have to do to unwrap a  single assumption because one special case is using a name only lookup rather than the jclass corresponding to the boxed type.

marscher:
If you are willing to redesign it the c++ backend, while preserving the public api, I'd be totally happy with that. As you mentioned the type wrapping is very detailed and goes through layers which are not necessary. I bet one can write this with a couple of templated c++ classes very easily. I agree that removing extra layers would also be a benefit for performance (replacing virtual calls with generated template code, removing extraneous abstraction layers for types). I like your idea of just mimicking the JNI type hierarchy as it is clean and simple. 

We can use at least cxx11 for now, as I'd be also tempted to use pybind11 for the python binding in the future.

Thrameos:
I carpet bombed the C++ layer removing extra class layers and tried my best to remove JPTYPENAME as the matching criteria.  There is still a few places where it was needed for the logic.  It also had some hidden behavior for handling nulls that I need to maintain. It is still heavier than I like.   

The new hierarchy has JPClass as the base type from which we have JPObjectClass,  JPArrayClass, and JPPrimitiveType.  
JPStringClass is a specialization of JPObjectClass.  There is one table for all C++ class bridges.  Dynamic cast is used if we need a particular type.  3 global tables died.  3 sections of cut and paste code died. A peice of code that did some expensive parsing passed it down three layers and did nothing with it when it got there also died.  There will be a new module call the promotes JPObjectClass to an object instance so we don't need to call for Name to get the class.  

Gutting the extra Jni layer and the host layer is still too heavy a lift.   Was the common layer intended for something other than python?   The only function currently is holding Jnisms from pythonisms.  Though perhaps that is a good thing.

If I succeed in getting that running again,  we can remove JPField with minor rewrite.  Everything can be done through the class members as we now have access to the class object.    We just need to make that code obsolete in the python layer.

marscher:
I think @DevilWolf intended to use these type abstraction layers for binding other languages than Python (eg. Ruby). But for the Python binding use case we do not need it. Nowadays we really should stick to pybind to glue to the jni wrappers.

Thrameos:
Okay making good progress.  But there are a lot of moving pieces.

Thrameos:
Everything now compiles and lambdas appear to work.  Unfortunately, I still have a lot of breakage regarding JPTypename and its abuses.  Mostly problems with translating primitive array types to the proper class structure.  Once all those test pass it will be good to review.

Thrameos:
Still working.  Thrameos/jpype#22

Thrameos:
Trying to get back to finishing up the patch.  The key issue is I broken the initialization sequence for the Java hook class.  I can't find sources for those classes so I am scouring the internet to see if some ancient copy has the source for those class files.  

Unfortunately, my father died after a 3 month battle with liver cancer last week.  So been feeling a bit too down to make much progress.

marscher:
On 03/26/2018 05:48 PM, Karl Nelson wrote:
> Trying to get back to finishing up the patch. The key issue is I broken
> the initialization sequence for the Java hook class. I can't find
> sources for those classes so I am scouring the internet to see if some
> ancient copy has the source for those class files.
> 
> Unfortunately, my father died after a 3 month battle with liver cancer
> last week. So been feeling a bit too down to make much progress.

My deepest sympathy for your big loss! Please take all the time you need
to finish things - there is no need to rush. Sometimes getting
distracted by work is good to overcome the pain.

The code for the hook classes should by in jpype/_jcollection.py

Best regards

Thrameos:
Unfortunately, that is not the hooks I was looking for.   The ones I am looking for are the ones stored in native/common/jp_invocationhandler.cpp, native/common/jp_reference.cpp, native/common/jp_referencequeue.cpp. They were created by compiling some currently lost java files.

I see some old copies of these files floating around the internet.

https://github.com/remram44/jpype/tree/master/src/java/jpype
https://github.com/remram44/jpype/tree/master/src/java/jpype/ref

There is a also a corresponding python script that converts the java to source code.  I am guessing at some point all of the code generation steps got cut out of the project and thus the corresponding sources used to produce those sources was cut at the same time.   Unfortunately when I was manipulating the code I broke the loading order sufficient to need to look at the underlying class files.  I have restored the required files in my private branch and will push them it.  

The java sources also give me a much better way to solve the name lookup problem.  The hashcode is incomplete because hashcodes are not unique (thank you java for not guaranteeing that class objects don't relocate due to garbage collection) thus I need to create a map of lists to store class wrappers by hashcode.  If I use the jclass directly to call a java class which I dynamically load, then I don't need to worry about collisions because the class wrappers can be stored as java.lang.Long in the Java layer.  The only downside it that like the reference queue code it will look a bit like magic because we will install our own java class and be using those hook points in C++.  

A prior author pulled all of the java dynamically loaded class stuff into jp_proxy with a note that this code doesn't belong here but is similar to the stuff that is being done here.  Eventually I was able to trace that cryptic comment to figure out some of the class in jp_javaenv were in fact using the dynamically loaded class as entry points.  Thus leading me to realize that a portion of the source code was in fact missing.  Now that I know about what it going on there I can reorganize it so that all the stuff to support those classes are together and make use of dynamic classes to reduce some of the sticky global objects problems.

Thrameos:
Completed rework of thunk code.  Unfortunately on a skii trip so can't update the pull request.  New code makes it through load and initialization but segfaults on a new object.  So still have dangling variables in the rework.

Thrameos:
Having the thunk code helped a lot as I can finally see the workings of the reference queue.  That cleared up a lot if mysterys about the proxy memory management. 

Once I get the last segfaults out I sold be able to start debugging the test bench.  Then I will need to back port any changes since this branch started.

Thrameos:
Enough seg faults are cleared to run some of the test cases.  Sill some mysteries remain.

I can't figure out what is the point of JException and PYEXC.  There doesn't seem to be much purpose for Throwables and Python exceptions to be separated.  I tested in python 3 and found no issue merging them.  It may be an issue for pythom 2.  

I have also experimented with removing  capsule usage.  It appears to be one of the areas where stradding between python 2 and 3 is requiring extra glue code.  Promoting all to formal classes elminates a bunch of differences between the two and allows a lot of stray entry points to be moved to the `__init__` methods.  

If I can isolate the remaining segfault that seems to be crashing inspect I should be able to run the full test suite.

Thrameos:
Making decent progress.  I split the jp_primitives_autogen code as the file is so heavily edited that the comment made no sense.  Further, having it as one giant file made it so I could not difference the files and detect errors between the different types.  With each broken out properly I can now resolve the edits and catch errors proactively.

I think I have this close enough to make a task list. (Not necessarily in order)

1) Update the python layer to match the back end code.
2) Finish reworking the jni wrapper layer to untangle jp_jni, jp_javaenv, jp_env
3) Debug test cases for python 3 development platform.
4) Replace remaining global tables with java thunk code for typemanager.
5) Remove dead code from JPObjectClass, JPArrayClass to remove remaining boot strap issues.
6) Add compatibility hooks if python interface is changed.
7) Test with python 2 and resolve issue.
8) Transfer pull request to jpype github repo so test on all architecture and resolve remote issues using the CI.
9) Consult hair restoration specialist regarding results from step 8.
10) Cross compare jp_primitive classes to resolve user edits in previous autogenerated code.
11) Audit for defensive programming for all entry points and calls to resources pushed by python layer.
12) Deal with interface for Proxy to support `class myClass(JProxy): def __init__(self): jpype.implements([jclass]) ...`
13) Deal with interface for Monitor to support `with jpype.synchronized(jobject): ...`
14) Test the UTF conversion code.
15) Leak check the Proxy and Reference thunk code.
16) Rework JArrayClass lookup to take a jclass and dims rather than using string binding to remove last vestiges of typename code.
17) Replace use of forname with direct call from jclass to eliminate name lookup issues where we already had the resource in memory.
18) Backport in revisions from pickle work to finish internal python class structure.
19) Complete experiment with python module to resolve issues with need for bound and unbound method and fields.  Make static methods available in object instance like it is in java.  Revise class structure accordingly.
20) Resolve unification of JWrapper and JavaObject for primitives and StringWrapper.  Check all type conversion patterns.  With removal of JPObject and unification with JPValue there is no need for separate wrapper classes.  All casting between objects and exact matched can now be resolved directly from the value type.  Replace wrapper classes with the same pattern used with boxed classes so that primitive types and cast objects are not dead ends as they are now.  
21) Disambiguate light at end of tunnel between daylight and oncoming freight train.

Still seems like a heavy lift.  But at least it is looking finite.

Thrameos:
I am closing this request.  All of this code was pulled by hand into the devel copy.   The diagrams etc are still relevant by the cod base is not stale.

PR #288, opened 2018-02-21
Fix for JPackage when accessing python privates.

Thrameos:
Simple patch to disable the incorrect warning message when accessing python private members like `__base__` in JPackage prior to starting the jvm.

### DIFF
--- a/jpype/_jpackage.py
+++ b/jpype/_jpackage.py
@@ -25,7 +25,9 @@ class JPackage(object):
     def __getattribute__(self, n):
         try:
             return object.__getattribute__(self, n)
-        except:
+        except AttributeError as ex:
+            if n.startswith("__"):
+                raise ex
             # not found ...
 
             # perhaps it is a class?


marscher:
why should object.__getattribute__('__base__') raise an attribute error? It should not for the Python wrapper classes?

Thrameos:
`JPackage` should respond just like any other object and give the same output as `object.__getattribute__(name)`.  So if we get an `AttributeError` then we should just pass it on rather than trying to load a java class.  Currently calling `JPackage.__base__`, `JPackage.__module__`, `JPackage.__meta__`, etc is trying to load java classes rather than returning the value or throwing AttributeError.  There is a lot of python mechanics that try to access like

``` 
   try:
      meta = obj.__meta__
     # Okay handle if we have a meta class
     ...
   except:
     # No meta class
     pass
```

The current `__getattribute__` was intercepting these and trying to load java classes which is incorrect.

Thrameos:
perhaps it would be more clear if I wrote it like

```
     def __getattribute__(self, n):
           # If it is a python private call object.__getattribute__
           if n.startswith('__')
              return object.__getattribute__(n)  # This will raise AttributeError if the python private does not exist

          # Otherwise, evaluate it as a package or a class
          try:
              return object.__getattribute__(n) # Get the cached value if it exists
          except:
              ...
```

which is the same as 
```
     def __getattribute__(self, n):
          try:
              return object.__getattribute__(n) # Get the cached value if it exists or the python private
          except AttributeError as ex:
              if n.startswith('__'):
                 raise ex
              ...
```

The second version is more efficient, but it less clear.

Thrameos:
The reason I noticed this was while running with tracing on I saw a call to `<M>JPTypeManager:: findClass : java.__bases__</M>` which seemed really suspect.

marscher:
Thanks for pointing it out. That makes a lot of sense.

PR #289, opened 2018-02-21
test vs. multiple java versions

marscher:
work in progress...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ addons:
   apt:
     packages:
     - ant
-    - openjdk-6-jdk
+   # - openjdk-6-jdk
     - openjdk-7-jdk
     - openjdk-8-jdk
     - openjdk-9-jdk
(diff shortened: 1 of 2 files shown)


Thrameos:
I put some code that may help to get the version testing in the lambda patch.  The key project is that the skip command for unit tests has to be run before the jvm is started, which is really problematic.

Thrameos:
Does the test suite stuff in the lambda patch look useful for your patch?  I can add a 1.6 section as well.

marscher:
I think the stages in harness should be treated like sets (eg. 1.8 will include all of the stuff of the old versions - or are there incompatible changes, that will prevent 1.8 from executing eg. 1.6 code?)

marscher:
the travis build matrix now has 72 jobs, which is way too much. I think we should drop support for Python 2.6 (not supported officially by the Python foundation). I guess there are more options to shrink the size of the matrix, but I need to think about it. 
Without py26 its still 64 jobs...

Thrameos:
Try trimming the with without numpy to just one java version.  Those are separate dimensions.  Also java 1.6 is likely too far from current.

hugovk:
Like Python 2.6, Python 3.3 is also past EOL and no longer supported by the Python Software Foundation. How about dropping that as well?

marscher:
will do that, thanks @hugovk!
We also need a modern ant version to support java 9, which is not installable from apt. Unfortunately apache deletes micro releases as soon as they get replaced by newer versions from their mirrors, so downloading directly from there requires too much baby sitting. We did this in the past on Appveyor (before using chocolatey) and every month or so CI was failing because the used ant version disappeared from the server.

Thrameos:
Is there a wget command that can pull the list of available that we can parse to get the latest.  It would be a pain to set up but if it is the only way to get the ant required that doesn't require babysitting the we should consider it.

marscher:
I do not think so. But these things only change rarely - so it is not that painful as you put it.

marscher:
superseded by #304

PR #300, opened 2018-03-27
Float conversion bug

Thrameos:
A copy and paste error was leaving widening conversions to  float in an incorrect state causing them to convert to 0 rather than expected value.   

This also switched assignments of fields from a RuntimeError to a TypeError when the conversion fails on fields.  This is more consistent with parts of the toolkit.  

The test are currently incomplete as another error in the kit is preventing a complete test run.  There is one unexpected behavior in that calling JInt(1.0) give the byte value from 1.0 in floating point rather than the expected behavior of creating an int with a value of 1, ie.  `int i=(int)1.0;`.   I am not sure if this is another bug or intended behavior.

### DIFF
--- a/native/common/jp_field.cpp
+++ b/native/common/jp_field.cpp
@@ -105,7 +105,7 @@ void JPField::setStaticAttribute(HostRef* val)
 	{
 		stringstream err;
 		err << "unable to convert to " << type->getName().getSimpleName();
-		RAISE(JPypeException, err.str().c_str());
+		JPEnv::getHost()->setTypeError( err.str().c_str());
 	}
 		
 	jclass claz = m_Class->getClass();
@@ -138,7 +138,7 @@ void JPField::setAttribute(jobject inst, HostRef* val)
 	{
 		stringstream err;
 		err << "unable to convert to " << type->getName().getSimpleName();
-		RAISE(JPypeException, err.str().c_str());
+		JPEnv::getHost()->setTypeError( err.str().c_str());
 	}
 		
 	type->setInstanceValue(inst, m_FieldID, val);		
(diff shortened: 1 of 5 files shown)


native/common/jp_field.cpp
```diff
@@ -105,7 +105,7 @@ void JPField::setStaticAttribute(HostRef* val)
 	{
 		stringstream err;
 		err << "unable to convert to " << type->getName().getSimpleName();
-		RAISE(JPypeException, err.str().c_str());
+		JPEnv::getHost()->setTypeError( err.str().c_str());
```
marscher: was the exception not propagated correctly or why is this change needed?

test/jpypetest/values.py
```diff
@@ -0,0 +1,85 @@
...
+    def setUp(self):
+        common.JPypeTestCase.setUp(self)
+        self.Fields = jpype.JClass('jpype.values.FieldsTest')
+
+# Int
+    def testIntFromInt(self):
+        self.Fields.intField = 1
+        self.assertEquals(self.Fields.intField,1)
+
+    def testIntFromInt(self):
```
Thrameos: Cut and paste error.  Will fix.

native/common/jp_field.cpp
Thrameos: The previous exception was just a generic RuntimeError.  I believe that it should properly be TypeError as per the python description.   We could live with either way but when setting up the testing framework I was left scratching my head as to what exception type should have been.  I originally wrote it as TypeError then found it failed because it was actually a RuntimeError.  I thought it best to make it clear the difference between trying to shove the wrong type of data into a strongly typed language versus trying to put the right type of data into a read only field.

PR #304, opened 2018-03-28
Cleanup testing

marscher:
This will build our matrix (so far without pypy) with conda-build variants. Will test first how things work out.
Switched from Nose to PyTest.

All test deps (ant, java, etc.) are handled by conda build.

- [x] enable ccache
- [x] enable coverage https://github.com/codecov/example-cpp11/blob/master/.travis.yml
- [ ] pypy is on conda-forge now, check if we can use it as variant. (https://github.com/conda-forge/pypy2.7-feedstock/issues/1)
- [ ] appveyor should work except for Cygwin (eg. split into one conda-build job and several cygwin jobs).


Thrameos:
You may want to disable the appveyor part of this build while testing or the CI system will be seriously backed up.

marscher:
good point.

marscher:
Since travis gmbh refused to add a super useful feature (https://github.com/travis-ci/travis-ci/issues/239), we should switch to CircleCI, which has junit.xml report highlighting. It also faster in most cases.

@originell can you please add circleci? BTW an jpype organization would make our life much easier.

originell:
@marscher do you mean a github organization or circleci?

marscher:
A github orga

marscher:
But it'd be fine if you can just add CircleCI in the meantime. I don't know the implications of moving repos around

originell:
Ah sure. Let's transfer this. Any name suggestions? (`jpype` is taken)… Naming things is hard!

marscher:
maybe jpype-python?

marscher:
@Thrameos Do you eventually understand why I get this 255 exit code immediately?

Thrameos:
Yes I have a guess.  The reason that the whole if statement was on one line is that I couldn't find a line continuation statement before so it is trying to execute the command `IF DEFINED CYGWIN (` and hitting a syntax error.  I assume that if you add a line continuation (or indentation, I am not clear what triggers it) so that it is clear the whole block is evaluated that it may work.

Thrameos:
Looking at the block you are probably better to just put the contents of the two if statements in scripts and leave it as the simple if CYGWIN run this script else run this script.   I am sure we can call the cinst and such in the cygwin sh script and the conda commands can be in the powerscript.

Thrameos:
I think you need to specify the interpreter in order for scripts to get executed from within the yml.   That is why it was calling the %CYGSH% and ps commands there.

marscher:
so I would call cmd.exe script.cmd? Would this pattern inherit all of the environment variables?

Thrameos:
Not sure if this helps.  I haven't used cmd.exe.

https://help.appveyor.com/discussions/problems/10014-false-build-fail-status

Thrameos:
Does this pull request need to be updated and finished for 0.7 release?

marscher:
I think we can safely postpone it.

Thrameos:
See #436 for PyTest conversion

PR #309, opened 2018-04-06
Fix bug with ipython imports

Thrameos:
A certain module in ipython scientific package is requiring all modules to use __path__ as either NoneType or a list.  Another module is requiring that path does is not None.  Changed the __path__ to an empty list to remove the interaction between jpype and ipython module.

### DIFF
--- a/jpype/imports.py
+++ b/jpype/imports.py
@@ -142,7 +142,7 @@ class _JImport(object):
     # Module requirements
     __doc__ = None
     __loader__ = None
-    __path__ = "<java>"
+    __path__ = []
     __package__ = "java"
 
     def __init__(self, name):


Thrameos:
I will have to get that from my coworkers.  The bug only occurred on Mac and unfortunately I have a PC so I was never able to replicate it myself.

desertfrog:
The original issue is crash and this error:

```
  File "/Users/monterial1/anaconda3/lib/python3.6/site-packages/pyximport/pyximport.py", line 253, in find_module
    fp, pathname, (ext,mode,ty) = imp.find_module(fullname,package_path)

  File "/Users/monterial1/anaconda3/lib/python3.6/imp.py", line 271, in find_module
    "not {}".format(type(path)))

RuntimeError: 'path' must be None or a list, not <class 'str'>
```

marscher:
Thanks for the stack trace. LGTM.

desertfrog:
It appears that both None and [] works for me.

PR #310, opened 2018-04-12
Print method sigs when overloaded match could not be found.

rbprogrammer:
In the exception raised when when an overloaded method could not be matched I really wanted the ability to see what the possible method signatures were.  I added some extra text to that exception text.

Sorry for the many changes.  I have Atom configured to remove trailing spaces.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -3,33 +3,36 @@ Changelog
 
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
-- **Next version - unreleased**
+- **0.6.4 - unreleased**
+  - Added extra verbiage to the to the raised exception when an overloaded
+    method could not be matched.  It now prints a list of all possible method
+    signatures.
 - **0.6.3 - 2018-04-03**
-  - Java reference counting has been converted to use JNI 
+  - Java reference counting has been converted to use JNI
     PushLocalFrame/PopLocalFrame.  Several resource leaks
     were removed.
 
-  - java.lang.Class<>.forName() will now return the java.lang.Class.  
+  - java.lang.Class<>.forName() will now return the java.lang.Class.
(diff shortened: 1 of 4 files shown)


doc/CHANGELOG.rst
```diff
@@ -3,50 +3,53 @@ Changelog
 
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
-- **Next version - unreleased**
+- **0.6.4 - unreleased**
+  - Added extra verbiage to the to the raised exception when an overloaded
+    method could not be matched.  It now prints a list of all possible method
+    signatures.
```
rbprogrammer: Sorry for the many changes in this file.  This block with this comment is the only real addition.  The other changes are from trailing whitespaces that got removed.

native/common/jp_method.cpp
```diff
@@ -123,7 +123,14 @@ JPMethodOverload* JPMethod::findOverload(vector<HostRef*>& arg, bool needStatic)
...
 		std::stringstream ss;
-		ss << "No matching overloads found for " << getName() << " in find.";
+		ss << "No matching overloads found for " << getName() << ", options are:" << std::endl;
+		for (std::vector<OverloadData>::iterator it = m_OverloadOrderCache.begin();
+				it != m_OverloadOrderCache.end();
+				++it)
+		{
+				ss << it->m_Overload->getSignature();
+				ss << std::endl;
+		}
```
rbprogrammer: Similar to CHANGELOG.rst, this is the only real change in jp_method.cpp.  The other changed lines in this file are from the removal of trailing whitespaces.

doc/CHANGELOG.rst
marscher: please keep the "next version" string, since I think that the bumpversion tool looks for this when we make a release.

rbprogrammer: Oh I see.  I reverted that line.

PR #318, opened 2018-05-26
Work of jp_method bug

Thrameos:
This  patch corrects a bug in which assignment was being used to copy a resource leaving the object operating on a non-existent global reference. It would then free the global reference freeing other resources resulting in errors.  However, at this was mostly at destruction time, it was rarely exercised.

The patch has the side effect of allowing construction of a JArray from an np.array when initializing.  np.array was not registering as a sequence and thus not hitting the correct code path.

The patch has the side effect of improving error reporting for methods.  (which is what I was trying to do when I found the bug).

```
import jpype
import jpype.imports

import java.lang.String
import java.lang.Double
import numpy as np

try:
    java.lang.String(5.0)
except Exception as ex:
    print(ex)

try:
    f=java.lang.String("foo")
    f.substring(6.0)
except Exception as ex:
    print(ex)

try:
    java.lang.String.copyValueOf(6.0)
except Exception as ex:
    print(ex)

try:
    x=np.array([0],dtype=np.int)
    f=java.lang.String("foo")
    f.substring(x[0])
except Exception as ex:
    print(ex)
```

now gives
```
No matching overloads found for constructor class java.lang.String(float), options are:
        public java.lang.String(byte[],int,int)
        public java.lang.String(byte[],java.nio.charset.Charset)
        public java.lang.String(byte[],java.lang.String) throws java.io.UnsupportedEncodingException
        public java.lang.String(byte[],int,int,java.nio.charset.Charset)
        public java.lang.String(byte[],int,int,java.lang.String) throws java.io.UnsupportedEncodingException
        public java.lang.String(java.lang.StringBuilder)
        public java.lang.String(java.lang.StringBuffer)
        public java.lang.String(byte[])
        public java.lang.String(int[],int,int)
        public java.lang.String()
        public java.lang.String(char[])
        public java.lang.String(java.lang.String)
        public java.lang.String(char[],int,int)
        public java.lang.String(byte[],int)
        public java.lang.String(byte[],int,int,int)
 at native\common\jp_method.cpp:130
No matching overloads found for class java.lang.String.substring(float), options are:
 at native\common\jp_method.cpp:130
No matching overloads found for class java.lang.String.copyValueOf(float), options are:
        public static java.lang.String java.lang.String.copyValueOf(char[])
        public static java.lang.String java.lang.String.copyValueOf(char[],int,int)
 at native\common\jp_method.cpp:130
No matching overloads found for class java.lang.String.substring(numpy.int32), options are:
 at native\common\jp_method.cpp:130
```

Implementation notes
- it was necessary to rework the caching mechanism on the jp_method, but this needs to be tested to make sure that it still properly handles ambiguous overloads.
- it was necessary to cause methods to be lazy loaded by the class as there is a race condition between loading methods from java.lang.Object during initialization.  I suspect that nothing was actually using this function as methods may be pulled later when the python class is created.  But I have not sorted that out.
- it was challenging to properly suppress the this pointer when printing the error message.  I changed the copy of the Bound version to help sort that out.

### DIFF
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -90,6 +90,12 @@ class _JavaArrayClass(object):
         else:
             raise AttributeError("%s does not have field %s"%(self.__name__, attr), self)
 
+def _isIterable(obj):
+    if isinstance(obj, collections.Sequence):
+        return True
+    if hasattr(obj,'__len__') and hasattr(obj,'__iter__'):
+        return True
+    return False
 
 def _jarrayInit(self, *args):
     if len(args) == 2 and args[0] == _jclass._SPECIAL_CONSTRUCTOR_KEY:
@@ -100,7 +106,7 @@ def _jarrayInit(self, *args):
             .format(len(args) + 1))
     else:
         values = None
-        if isinstance(args[0], collections.Sequence):
+        if _isIterable(args[0]):
             sz = len(args[0])
             values = args[0]
         else:
--- a/native/common/include/jp_class.h
+++ b/native/common/include/jp_class.h
@@ -39,17 +39,8 @@ public :
 	
 	JPField*                getInstanceField(const string& name);
 	JPField*                getStaticField(const string& name);
-	JPMethod*				getMethod(const string& name);
-	vector<JPMethod*>		getMethods() const
-	{
-		vector<JPMethod*> res;
-		res.reserve(m_Methods.size());
-		for (map<string, JPMethod*>::const_iterator cur = m_Methods.begin(); cur != m_Methods.end(); cur++)
-		{
-			res.push_back(cur->second);
-		}
-		return res;
-	}
+	JPMethod*		getMethod(const string& name);
+	vector<JPMethod*>	getMethods();
 
 	jclass getClass()
 	{
@@ -96,12 +87,12 @@ private :
 
 private :
 	bool                    m_IsInterface;
-	JPClass*				m_SuperClass;
-	vector<JPClass*>		m_SuperInterfaces;
+	JPClass*		m_SuperClass;
+	vector<JPClass*>	m_SuperInterfaces;
 	map<string, JPField*>   m_StaticFields;
 	map<string, JPField*>   m_InstanceFields;
 	map<string, JPMethod*>	m_Methods;
-	JPMethod*				m_Constructors;
+	JPMethod*		m_Constructors;
 };
 
 
--- a/native/common/include/jp_hostenv.h
+++ b/native/common/include/jp_hostenv.h
@@ -148,6 +148,8 @@ public :
 	virtual void printReferenceInfo(HostRef* obj) = 0;
 	virtual bool isByteBuffer(HostRef*) = 0;
 	virtual void getByteBufferPtr(HostRef*, char**, long&) = 0;
+
+	virtual const char* getTypeName(HostRef*) = 0;
 };
 
 #endif // _JPHOSTENV_H_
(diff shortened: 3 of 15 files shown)


Thrameos:
I added similar logic to the reporting of ambiguous methods.  It passes checks on my system properly.

There were a large number of hidden assumptions in the method code.  The JVM passes multiple copies of the same method and the order that they appeared on the final list when used mattered on the detection of ambiguous methods.  The previous code was using sets and maps to try to solve the graph problem.  I switched it to a simpler graph algorithm and documented it so that it will be more clear if anyone wonders in there again.

Thrameos:
@marscher I know that previously there was debate if the error reporting was supposed to be this level of verbose or not, but given that #310 is now in I thought it best to clean up the reporting to make it more human readable.  

The global reference bug it at least as far as I could tell was real.  I logged all the calls to global release and at caught it releasing more object that it allocated.  Fortunately, it was only on the destruction point and nothing used the trashed JVM after that so it is not likely to hit us.  I suppose if we ever got code that unloaded and reloaded the JVM we would be in trouble.

I probably should not have rewritten the graph algorithm because I really could not understand the original.  I did end up with a robust solution, but there was still something in the original code that managed to put the boxed objects above the primitive types in the ordering that I was not able to replicate.  I am not sure if it was intended or a side effect of the previous logic having to do with the fact the previous code was using an alphabetized list.   

I also really wanted to get stuff to lazy load on the methods.  I tried and failed on that front.  There really is no reason for the system to load up every class that appears in a method until after the python class requests the list.  But currently that is the model.  I may try again in another patch to shift all that work to a helper function rather that part of the class wrapper.  Then we will break all these start up dependencies.

Thrameos:
@marscher Any action on this pull request?  It has already fallen behind once and been updated.  

I backed out the array changes when I discovered a larger set of issues with array types.  But this patch is sort of a prerequisite to finishing that one as good error reporting is needed to debug the other.  Having to merge them fix and then reapply the applicable changes to the other is getting tedious.

Thrameos:
@marscher Darn EOL when in first, had to happen one of these days.  Okay I will have to revise and resubmit.

Thrameos:
@marscher This patch is ready for review.

Thrameos:
Not to worry.  My long term plan is to move all of the major logic from within C++ to either the Python side or to a java library leaving the C++ just for memory management and applying argument matching.  

When  your logic was written we didn't have boxed types exposed quite as directly or variable argument support and we hadn't yet customized boxed types to be derived from python primitives.  Thus I am pushing a lot more tests because we have a lot more edge cases.  The previous logic was still holding but it was really depending on the order of order of certain exact matched coming up and at least some of those relationships were unspecified.   I think part of what is missing is that boxed types and primitives have to be given an explicit relationship so that they can be sorted properly.  Because some relationships were not fully specified, it seems like it was literally the implementations use of string maps that was imposing a resolution.  When I converted it to use a list rather than a map to hold resources the order of the insertion was governed by the order that java provided rather than fixed alphabetically exposing the fact there were not test cases for the unspecified relationships.  Thus byte and java.lang.Byte would appear in the final ordering differently than short and java.lang.Short.  But the order could then swap if it was the second argument.  Thus the test cases would have to know to look for that issue by testing in both the first and second argument position.   I have been trying to remove all string based resolutions in place of direct logical comparisons.  

Thanks for the pointer into java specs.  It will help when I push the logic into the jpype jar.

Thrameos:
This request has been merged into the development patch.  I am leaving it open in case we want to cut a release with the earlier code base.

PR #323, opened 2018-06-04
And now for something completely different with lambdas

Thrameos:
@marscher Okay I hate to bother after that last horrible pull request but I have a different idea on how to fix the lambda issue.  (Please don't merge this, it is just for review)

Unfortunately, the lambda problem continues to impact my library support, so I need to take another shot at fixing it.  The direction I tried last time was a struggle as jumping from the system that used TypeName and HostRef to using jclass and PyObject* required too big of leap for me to put my feet down.  Though I tested regularly with small codes running the full nose test during transition was too much as it could never complete during the transition and I paid for it with a deep core bug that I could never resolve.  

Thus I am going to propose work on a different solution entirely pulling from what I learned last time.  Rather than struggle with C++/CPython Module/Python with a turn around time of about 15 minutes per test.  I propose moving as much of the jpype core into java as possible. This would leave just the type conversions in the C++ layer.

I did a prototype (not connected to anything) so you can review.  Is this a viable direction?

### DIFF
--- a/thunk2/README
+++ b/thunk2/README
@@ -1,0 +1,35 @@
+This is a set of experimental changes to fix the issues with Java 8 lambdas.
+
+The current jpype system is using names to distinguish classes, but lambdas
+don't have names that the class loader can find.  Thus any lambda chokes the 
+system badly.  We tried to switch this to jclass but these are handles that
+are not unique.  Comparing them with java doesn't work because it is both
+slow and won't work in a map.  We tried hashcodes but those aren't unique
+either.  Thus there is no good solution down the current path.  And the
+things are likely to just get worse by Java 9.
+
+The following proposal is to do the opposite. Rather than forcing C++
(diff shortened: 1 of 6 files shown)


Thrameos:
Although the urgency on this particular design is lower now that I have corrected the issue with lambdas, I am going to try to resync this concept with the devel once I have the compiler pull in.  There are a number of minor issues with the method presidence that are currently outstanding.  Trying to fix them from within C++ is a pain and we have lots of redundancy in that we create a separate method overload for each class even if the particular class never overrides that method.  If we let a Java handle the type manager tasks I can share a great deal more resources and trim out another good sized chunk of the C++ code.  Things get a lot easier if I can call the Java order resolver to decide if one method should shadow another.

marscher:
sounds like a good approach. I'm  sorry for the huge delay. The memory compiler is on devel now.

PR #324, opened 2018-06-05
Test of thunk building

Thrameos:
This is a test of the setup.py that constructs the thunk code automatically.   I am having serious problems with nose because it loves to get in and add stuff to the path automatically which is forcing my directory structure to never include a directory "java".   No luck in PyTest thus far.


marscher:
I do not get why you need to move everything around. This messes up the project history a lot  - please do not sacrifice the history for the sake of using an unmaintained testing software (nose). What is the problem with pytest?

Thrameos:
I was trying to get all the source back under src to match a more standard configuration (or at least what I understand it to be.)  I used "git mv" so it should have no changes to the annotations.  I can resubmit another pull with the same directory structure as before.  This was just an experimental pull to see what is broken on using the build matrix.  I am going to run into an issue though if I use the custom class loader.   We will have 4 separate source directories.  "native/common", "native/python", the bootstrap class loader, and the jar sources.  I can just put them all under native if you prefer.  

I did split out the setup.py because I needed to override the build process to insert the thunk building.  That means a lot of class overrides.  

The repo I was referencing from to find the lost code that was being used to make all the java code had a different structure.

https://github.com/remram44/jpype/blob/master/src/java/jpype/JPypeInvocationHandler.java

I am a bit unclear as to where the reorg happened as both are initial imports.

As for PyTest I just get "no tests run" after 5 minutes a waiting no matter what options I hit.  Thus is it currently not viable for me.  Thus I have to use nose to develop.  Unfortunately 2AM is rapidly approaching so I have to call it a night.

marscher:
On 05.06.2018 10:35, Karl Nelson wrote:
> I was trying to get all the source back under src to match a more standard configuration (or at least what I understand it to be.)  I used "git mv" so it should have no changes to the annotations.  I can resubmit another pull with the same directory structure as before.  This was just an experimental pull to see what is broken on using the build matrix.  I am going to run into an issue though if I use the custom class loader.   We will have 4 separate source directories.  "native/common", "native/python", the bootstrap class loader, and the jar sources.  I can just put them all under native if you prefer.  
> 
> I did split out the setup.py because I needed to override the build process to insert the thunk building.  That means a lot of class overrides.  
> 
> The repo I was referencing from to find the lost code that was being used to make all the java code had a different structure.
> 
> https://github.com/remram44/jpype/blob/master/src/java/jpype/JPypeInvocationHandler.java
> 
> I am a bit unclear as to where the reorg happened as both are initial imports.
> 
> As for PyTest I just get "no tests run" after 5 minutes a waiting no matter what options I hit.  Thus is it currently not viable for me.  Thus I have to use nose to develop.  Unfortunately 2AM is rapidly approaching so I have to call it a night.
> 
Using git mv does not preserve history:
https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history#2314757

So please avoid moving things around if it is not inevitable.

Pytest does not like the current testsuite approach. It looks for files
starting with a test_ prefix.

So if you invoke pytest like this:
cd test
py.test jpypetest/*.py

It collects all of the test classes properly.

Thrameos:
I am all for unix line ends.  I may work on windows but I have forced everything to be unix because 80% of my team is on macs.  (given I have a 5 man team you can work out the math on that one)

I am currently using netbeans for all the work on the java native stubs and then copying the files back and forth between the netbeans and jpype as I debug the typemanager class.  I also have all my current extension classes like the int[] and double[] math routines in netbeans.  It would be best if I could get a netbeans project location in the jpype tool chain so that I am not switching back and forth as every time I forget to copy a file back I do testing on invalid source.   That way I just open the netbeans project under jpype and everything links to the actual sources a level above.

(Got to run to drop kids off.  Back in 30m to finish conversation.)

marscher:
On 05.06.2018 16:12, Karl Nelson wrote:
> I am all for unix line ends.  I may work on windows but I have forced everything to be unix because 80% of my team is on macs.  (given I have a 5 man team you can work out the math on that one)
> 
> I am currently using netbeans for all the work on the java native stubs and then copying the files back and forth between the netbeans and jpype as I debug the typemanager class.  I also have all my current extension classes like the int[] and double[] math routines in netbeans.  It would be best if I could get a netbeans project location in the jpype tool chain so that I am not switching back and forth as every time I forget to copy a file back I do testing on invalid source.   That way I just open the netbeans project under jpype and everything links to the actual sources a level above.
If it makes your life easier to store this folder, please go ahead - but
I do not know a good location for it. I guess Netbeans expect it to be
at the project root?
> 
> (Got to run to drop kids off.  Back in 30m to finish conversation.)
> 
>

Thrameos:
Netbeans in its native form likes to have a bunch of files at the root.

```
project_root/
   build.xml
   src/
   build/
   nbproject/
   dist/
   test/
```
As this is just going to be a place for editing and internal testing I was planning on pushing it a level deeper in the hierarchy and having it linked back up a level.  I just need to know what to call it so that it doesn't clutter things up too much.

It creates its own build.xml file which is needed for it to build its copy of the jar (which will not have the native calls).  Thus i worry about the structure.

Perhaps it would be best if I explained where stuff is going and what is needed.  As explained in #323 I am planning on gutting out the C++ layer by pushing everything I can into support calls in pure java.  This will eliminate a lot of tedious jni code and improve speed.  Rather than creating a jp_class an having it do 100 jni calls to get the each method and each field and probe the names of each method, class, super class interface, etc, it will just ask the java extension "please give me all the stuff I need".  The java layer figures out everything and calls back to C++ to allocate resources with calls passing large lists of arguments (which according to IBM is how they intended for it to be used).  The side benefit is that when we inherit a class the method wrappers and the overloads are shared, thus we also cut the memory foot print.  

Okay so what does this mean for structure... we have a few different java source trees.  For ease in building they will need to be separate trees.
- bootstrap - this is a classloader (explained later) (1)
- pure java - this is all the stuff that doesn't involve a native call that is going to get be built into jpype (2)
- native - this is everything that calls the C++ layer that will be built into jpype (3)
- stubs - this is a shadow of all classes in native but with the methods replaced by stubs  for the test harness (4)
- test - a test harness that uses a jar build of (1),(2) and (4) to verify functionality of the core. (5)
- utilities - support classes that are not used in jpype module to do things like turn a stub file into a native file or turn a class file into a C++ stub.  (6)

So you can see there are a few builds that have to happen (2)(3) have to go into a jar which along with (1) will all get compiled into _jpype.  (1)(2)(4) have to go into a second jar that is tested by (5) to make it so that a bug in the operation of (2) doesn't have to be discovered by running through everything.  And (6) uses the jar to do code generation.  I may be able to use python to do the function of (6) but that would mean having both the jar that a live copy of jpype needs and the copy that I am building stubs for in memory.  And I hate JarJar. (okay not really I used JarJarLinks all the time, JarJar Binks is another issue entirely but I digress)

The problem is that jpype is already a pretty disorganized place.  We had way to many layers to the onion in because of the ruby.  And the CI is forcing a lot of stuff on the top level.  So I was thinking that (1),(2),(3) belong under native/ some place and (4)(5) belong in the netbeans managed subtree.   But as you are the conductor and I am the band it would be best if I got some approval so it is clear what I am doing and why so we don't end up with just more mess. 

The classloader is the last thing that I have yet to explain.  We are currently doing everything through DefineClass.  This is all fine, but it prohibits a class from loading that has a circular dependency in the imports.   Thus if A imports B and B imports A even implicitly by package then if I load one I will get `ClassNotFoundException`.  Thus I either need to write my code in a rather strange fashion with one hand tied behind my back, or use an actual classloader to resolve the issues.  After all if it is a problem for me to write (and I have been doing jni stuff since 1994) then it will be a virtual bar for anyone else.

If you can suggest names and locations for each of the pieces (1),(2),(3) then it would save me some effort.

marscher:
On 05.06.2018 17:13, Karl Nelson wrote:
> Netbeans in its native form likes to have a bunch of files at the root.
> 
> ```
> project_root/
>    build.xml
>    src/
>    build/
>    nbproject/
>    dist/
>    test/
> ```
> As this is just going to be a place for editing and internal testing I was planning on pushing it a level deeper in the hierarchy and having it linked back up a level.  I just need to know what to call it so that it doesn't clutter things up too much.
> 
> It creates its own build.xml file which is needed for it to build its copy of the jar (which will not have the native calls).  Thus i worry about the structure.
> 
> Perhaps it would be best if I explained where stuff is going and what is needed.  As explained in #323 I am planning on gutting out the C++ layer by pushing everything I can into support calls in pure java.  This will eliminate a lot of tedious jni code and improve speed.  Rather than creating a jp_class an having it do 100 jni calls to get the each method and each field and probe the names of each method, class, super class interface, etc, it will just ask the java extension "please give me all the stuff I need".  The java layer figures out everything and calls back to C++ to allocate resources with calls passing large lists of arguments (which according to IBM is how they intended for it to be used).  The side benefit is that when we inherit a class the method wrappers and the overloads are shared, thus we also cut the memory foot print.  
This sounds like a sane change and saving memory and calls is always a
good thing.
> 
> Okay so what does this mean for structure... we have a few different java source trees.  For ease in building they will need to be separate trees.
> - bootstrap - this is a classloader (explained later) (1)
> - pure java - this is all the stuff that doesn't involve a native call that is going to get be built into jpype (2)
> - native - this is everything that calls the C++ layer that will be built into jpype (3)
> - stubs - this is a shadow of all classes in native but with the methods replaced by stubs  for the test harness (4)
> - test - a test harness that uses a jar build of (1),(2) and (4) to verify functionality of the core. (5)
> - utilities - support classes that are not used in jpype module to do things like turn a stub file into a native file or turn a class file into a C++ stub.  (6)
I do not really understand why (6) is needed. Why do you need to
generate C++ code? This pattern sounds like Apache JCC to me. JCC is
static, so you need to generate cpp wrappers for each jar prior using
it. This would be a deal breaker for Jpype, because it manages to do
everything dynamically. But probably I just got you wrong? :)

> 
> So you can see there are a few builds that have to happen (2)(3) have to go into a jar which along with (1) will all get compiled into _jpype.  (1)(2)(4) have to go into a second jar that is tested by (5) to make it so that a bug in the operation of (2) doesn't have to be discovered by running through everything.  And (6) uses the jar to do code generation.  I may be able to use python to do the function of (6) but that would mean having both the jar that a live copy of jpype needs and the copy that I am building stubs for in memory.  And I hate JarJar. (okay not really I used JarJarLinks all the time, JarJar Binks is another issue entirely but I digress)
> 
> The problem is that jpype is already a pretty disorganized place.  We had way to many layers to the onion in because of the ruby.  And the CI is forcing a lot of stuff on the top level.  So I was thinking that (1),(2),(3) belong under native/ some place and (4)(5) belong in the netbeans managed subtree.   But as you are the conductor and I am the band it would be best if I got some approval so it is clear what I am doing and why so we don't end up with just more mess. 
I agree that the code base is a mess. Thank you very much for taking the
shot of cleaning it up. As for the Java part, I think it does not hurt
anybody, if you turn this subtree into a NB project. However code should
be organized such that it is independent of the used IDE. But it only
affects the testing part (4)(5), right? So please go ahead and change it
like you desire.

Do you want to keep the packaging with Python/setuptools? Because this
is messy as well. If most of the new code relies on Java, we should also
switch to a modern Java build tool, shouldn't we?
> 
> The classloader is the last thing that I have yet to explain.  We are currently doing everything through DefineClass.  This is all fine, but it prohibits a class from loading that has a circular dependency in the imports.   Thus if A imports B and B imports A even implicitly by package then if I load one I will get `ClassNotFoundException`.  Thus I either need to write my code in a rather strange fashion with one hand tied behind my back, or use an actual classloader to resolve the issues.  After all if it is a problem for me to write (and I have been doing jni stuff since 1994) then it will be a virtual bar for anyone else.
Python it self can not deal with cyclic import dependencies. I do not
think it is possible if we treat Java packages like Python packages to
resolve this.
> 
> If you can suggest names and locations for each of the pieces (1),(2),(3) then it would save me some effort.
> 

Since you have a much better understanding of what you're going to do, I
suggest that you choose these names.
But for (1)(2) I would simply go for a dir named "java", put a README.md
inside and then everybody knows what to expect inside.

Thrameos:
Item (6) is just a time saver for me.  when I have to write out code that says

```
jmethodID my_method = field.GetMethodID(my_class, "mymethod", "(jjL;java/lang/class;j[jL;org/jpype/typemanager/TypeManager;j[j[Lorg/jpype/typemanager/MethodOverload;');  
```
(whoops sorry that was wrong because the 8th argument was supposed to be an integer, sorry no work for you, take 20 minutes to rerun with trace to see that ClassNotFoundException was thrown during bootstrap causing segfault.)

If that happened ever single time that I update an interface call by hand I am likely to go nuts (or just never change it even when it is needed.)  It is much better to just have a simple method load the class by hand and use reflection to get that horrible string and print to the screen.  I am not actually doing code generation, just making life somewhat more sane.  But that will be in the netbeans so noone will see it.  But trying to run without something to help with the jni bookkeeping is likely not sane.   This "stub" generator won't in anyway affect the build or use process.

> But for (1)(2) I would simply go for a dir named "java", put a README.md
inside and then everybody knows what to expect inside.

The problem with calling a directory "java" is that nose likes to pick up source directories and place it in the python path and because of the search order of python the statement "import java.lang.String" will get the java directory and not the java package it all goes bust in the test. I completely agree with you that if I can find a way to make that stuff "java" it would make the most sense.  That is why I really want to get PyTest running and the test structure reorged so I can punch nose.

Also we can't have both the classload in the jar.  We have to have two memory thunks. Thus org/jpype/bootstrap/ByteArrayClassLoader.class and org.jpype.jar need to be compiled to build/lib and the jar contents need to go to build/classes.   I suppose I could have the build.xml be clever with exclude patterns to make one file compile to one place and the rest compile someplace else.  I will see how much trouble that will be.

As to the classload, this isn't really a python issue.  Java has no problems with circular references in java files so long as the classloader is there to sort them out.  But if I have to write my java code such that every file is completely self contained then it is going to get very restrictive.  At least in python you only need to worry about circular at the package level.  The java restrictions without a class loader will be much worse.

> Do you want to keep the packaging with Python/setuptools? Because this
is messy as well. If most of the new code relies on Java, we should also
switch to a modern Java build tool, shouldn't we?

I took a good 6 hours last night trying to understand what was the concept of distutils.  Afterward I wanted to gouge my eyes out with a rusty spoon.  That said, I wrote a set of patterns to make setup.py call `ant` to create the required jars and then a bin to `char _foo[] = {0x01, 0xa4, ... }` thunk generator that is required for _jpype.  The actual amount of stuff we will be doing with java is confined to that.  98% of the build process remains creating the _jpype extension, packaging, and installing.  If python ever gets a sane build distribution system I would say switch in a heart beat but for now we have to live with it.

The best I could do for setup.py was to split it into a module so that each part of the process has the appropriate command with limited scoped variables and avoids the global mess that it was growing into.   It was looking like an unstructured mess.  I guess you are much better at python than I as I couldn't follow it at all.  Of course I can't call this reorg a success until the CI says "no errors found".

Thrameos:
> I see. This would probably involve some copy paste code at worst?

The class loader is really just a JPEnv::init() thing.  In order to get all the thunks I want to load I either have to load each class by hand using JNI DefineClass (with horrible restrictions) or I have DefineClass for my custom classloader then use that to load the whole jar into memory.  This cuts out a huge chunk of code that I would have to write.    

I would use a code generator to give me the strings to do some cut and paste work when I plan to change a signature.  I could even just have the code generator write a .h file with const char* for each of the methods that I use.  This is sufficiently internal that noone would see it unless they were doing hard core development.

> sorry about this particular reality of Python...

I find setuptools a bit shocking.  A system with different idiom in each command class and overrides as the default method of doing anything.  Once you go down that road you can't ever change it without breaking everyone's code every place, because you have forced your class guts to be interacted with.

I have everything working fine on my 5 separate copies of pythons (windows 2.7. standalone windows 3.6 standalone, windows 3.6 conda, cygwin 2.7, cygwin 3.6) but not a single one in the CI worked because something is different in the setuptools.  I would be hard pressed to design a worse system if I tried.

> I think for gcc you need to add the --std=cxx11 flag to make it compile.
Dunno about the install-option handling yet.

I was offended by the need for casting each byte of the thunk.  Thus I tried something different but discovered C++ does not guarantee that "char", "unsigned char" or "signed char" are ever 8 bits or even compatible.  Forcing the use of cxx11 is likely also not a solution.  I have reverted it to the horrible casing method, but now that thunks are build in the build process I guess I can ignore it.  When we had large files of constants checked as sources it was offending.

Thrameos:
@marscher Assuming this passes CI this pull request is complete.

The work in this pull was to fetch to versions of the thunk code from the archives and reenable them so that we can actually see what is being done in the java native sections.  I then integrated those thunk pieces into the build process so that we always get code that is up to date. To make it possible to develop the thunks I then reworked the setup.py script to split it into small pieces.  I remove all the global variables and made each command a separate unit.  A few extra targets got added 
- build_java - compiles java code into thunk classes
- build_thunk - compiles thunk classes into thunks for _jpype
- test_java - compiles the test harness

In theory if I can get setup.py test command to work we can have the whole build, test, install run from setup.  But that needs more work to pull off.

The existing global option was made into a formal setup.py option so it now shows up on the --help message and a new global for the location of ant was added (--ant=).  This had the odd side effect of requiring the master switch to pip to be --global-option rather than --install-option. 

It will require ant during the build process as a result because we now are not shipping the built source.  There is likely a side effect that we will need to specify the java version to be the lowest supported in the native/build.xml so we build compatible java classes. 

Last this pull fixed an obnoxious bug that the JPypeReferenceQueue did not start itself as daemon and thus was making shutdown impossible without just pulling the plug on JVM.  I haven't tested if this makes DestroyJVM work but it is certainly part of the process to make that code work properly.

Thrameos:
This one is currently in.

PR #326, opened 2018-06-06
Eol

Thrameos:
This is a pull request to normalize all files to eol in the repo after the thunk pull request is merge.   It may or may not affect the blame depending on if `-w` is applied.  It is not clear to me if the annotations shown on github are with or without -w.


marscher:
I guess to make CI pass :+1:

Thrameos:
Sorry that was my fault.  I thought you would put the thunk in first and then I would rebase the EOL, but you pulled the EOL first.  I should have been more clear.

marscher:
On 06.06.2018 22:18, Karl Nelson wrote:
> Sorry that was my fault.  I thought you would put the thunk in first and then I would rebase the EOL, but you pulled the EOL first.  I should have been more clear.
> 
We can also undo things, if your life gets easier then!

PR #327, opened 2018-06-06
Numpy array

Thrameos:
This pull request exposes a bug in the array slice routine in which it assumed all primitives with the same type are compatible and thus tries to use the direct copy method.  This leads to rather predictable failures.


Thrameos:
This pull request is not ready to go.  It does correct the basic issue that memory was pushed around inappropriately but it requires the additional step of make numpy types be recognized by the canConvertToJava.  Without this everything will take the bypass and nothing will hit the optimized copy path.

Thrameos:
This one needs to be finished.  Not sure yet why is does not work on *nix.

Thrameos:
Hmm.  Seems that somehow I lost part of this pull request.   

I did further analysis of the numpy problem.  The use of a memory view without using the numpy structures to first determine how to access the data is fundamentally flawed.   I started implementing a replacement under devel, but it is a pretty heavy lift because the NUMPY array api allows so many different memory organizations.   My implementation will certainly be better than allowing incorrect data to be passed in so long as the size of the data types is the same, but there are a lot of edge cases.  I am not sure even how I would create a misbehaving, byte order reversed, view of data with a stride.  Unfortunately, that is what would be required to hit the test cases to ensure we do everything correctly.  

After it is fixed in devel we can look to back port if needed.

marscher:
I do not think it is possible to map non contiguous arrays to Java. Therefore we should just raise, if the array does not full fill this requirement.

Thrameos:
Seems like this is the problematic patch.   Shall we revert it?

PR #332, opened 2018-06-21
Fix #331: Don't call Py_Initialize() in PyInit__jpype() to support Python 3.7

### LINKED ISSUE #331: JPype is incompatible with Python 3.7: Fatal Python error: _Py_InitializeCore: main interpreter already initialized
Since Python 3.7, calling Py_Initialize() twice now fails with a fatal error:

```
Fatal Python error: _Py_InitializeCore: main interpreter already initialized
```

But PyInit__jpype() calls Py_Initialize() at native/python/jpype_python.cpp:161:

https://github.com/originell/jpype/blob/490eb6511a0a0b57be2191cbbf3a90bac45e5879/native/python/jpype_python.cpp#L150-L157

It doesn't make sense to call Py_Initialize() in the initialization function of a module, since Python must be initialized to able to... import jpype :-)

The fix is just to remove the call.

vstinner:
Since Python 3.7, calling Py_Initialize() twice triggers a fatal error.

Simply remove the Py_Initialize() call from PyInit__jpype(), since it is just useless.

### DIFF
--- a/native/python/jpype_python.cpp
+++ b/native/python/jpype_python.cpp
@@ -153,7 +153,6 @@ PyMODINIT_FUNC PyInit__jpype()
 PyMODINIT_FUNC init_jpype()
 #endif
 {
-	Py_Initialize();
 	PyEval_InitThreads();
 	  
 #if PY_MAJOR_VERSION >= 3


vstinner:
cc @hroncok

vstinner:
I tested manually https://github.com/konlpy/konlpy test suite with this JPype fix: Python doesn't crash anymore! My PR fixed my bug.

vstinner:
I also reported the issue upstream: https://bugs.python.org/issue33932

vstinner:
cc @marscher 

Note: I'm a CPython core developer.

marscher:
I trust you and my CI, was just waiting for it to pass. Thank you very much for your patch!

marscher:
I guess we need to cut this jpype release soon...

Thrameos:
Does the PyEval_InitThreads() belong in the module initialization as well?   The CPython documentation wasn't clear if this was a module thing or just part of the python startup.

vstinner:
> Does the PyEval_InitThreads() belong in the module initialization as well?

You have to keep this call.

We enhanced the documentation in Python 3.7:
https://docs.python.org/dev/c-api/init.html
"Note: The following functions should not be called before Py_Initialize(): (...) PyEval_InitThreads()."

Moreover, we did another change in Python 3.7:
" void PyEval_InitThreads(): (...) Changed in version 3.7: This function is now called by Py_Initialize(), so you don’t have to call it yourself anymore."

On Python 3.6 and older, it's a good thing to call this function to prevent race conditions when spawning threads in C which then uses the Python API.

vstinner:
>  Changed in version 3.7: This function is now called by Py_Initialize(), so you don’t have to call it yourself anymore.

Oh, and if you are curious, I wrote an article about this specific change :-) https://vstinner.github.io/python37-gil-change.html

marscher:
That's very cool! Congrats for nailing this nasty one down :sunglasses:

Thrameos:
@marscher We likely need to hit the known defect on the arrays considering everything with the same size as being directly convertible before cutting a release.  But it will be a week or two before I can get back to that.

vstinner:
While removing the Py_Initialize() is safe and fix a bug in Python 3.7rc1, FYI I fixed the future Python 3.7.0 final to allow again to call Py_Initialize() twice. But Nick Coghlan still plans to deprecate this behaviour, in the future calling Py_Initialize() twice will fail (as Python 3.7rc1).

PR #347, opened 2018-10-09
Bug varargs

Thrameos:
This patch fixes a bug introduced during the refactorization having to do with empty argument lists.  A number of java methods such as `Paths.get()` require at least one argument with optional additional arguments.  There was no check for this pattern in the varargs test bench.  In the case of an empty optional argument it generated a null list rather than a list with 0 elements.  This caused certain methods in java to fail.

I rebased my development branch to match the head, but this appears to have introduced a few null commits as the squash point introduced by github did not match exactly.  I don't think that that will introduce any problems.

The missing flag is now set properly and I added appropriate testing to ensure this bug does not reappear.


marscher:
LGTM

PR #350, opened 2018-10-15
Unicode support.

Thrameos:
This patch completes the work on the UTF8 support and incorporates Rene Bakker's UTF8 testing code.  All problems in the encoder should now be fixed.  (though I still object to the fact that Java decided to call UTF-16 surrogate codes encoded as UTF-8 as anything standard).  

Most of the patch is pretty clean but there were some white space errors cropping up from the autopep8 that slipped in that were cleaned up by this patch hence a few random whitespace changes.   There is one FIXME tag which has some addition clean up which is not worth doing at the current time regarding pulling Python.h into the common source.  I was not able to give Rene credit because the patch history was a major mess between the master and devel branch.  The fraction between Python2 and Python3 over str vs unicode was replicated.  

There are still 3 more pulls before I get to my development head, but those are having problems with the conflict resolution that broke something when merging.   Assuming I can get those back up to working we should be getting closer to a release candidate.   Sorry for the heavy review assignments.

### DIFF
--- a/jpype/_jobject.py
+++ b/jpype/_jobject.py
@@ -96,6 +96,9 @@ class JObject(object):
     def __str__(self):
         return self.__javavalue__.toString()
 
+    def __unicode__(self):
+        return self.__javavalue__.toUnicode()
+
     def __hash__(self):
         return self.hashCode()
 
--- a/native/common/include/jp_javaframe.h
+++ b/native/common/include/jp_javaframe.h
@@ -355,6 +355,7 @@ public:
 	const jchar* GetStringChars(jstring a0, jboolean* a1);
 	void ReleaseStringChars(jstring a0, const jchar* a1);
 	jsize GetStringLength(jstring a0);
+	jsize GetStringUTFLength(jstring a0);
 } ;
 
 /** JPClass is a bit heavy when we just need to hold a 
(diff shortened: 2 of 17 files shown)


Thrameos:
Test fail on `ImportError: cannot import name OrderedDict` for 2.6.   Appears to be a function introduced in 2.7.   I will have to work around it.

marscher:
including python.h is not so much of a big deal in a cpp unit I'd guess. The fix and tests look good to me, although I have to admit that these bitshifting operations for range definitions are confusing. Do you mind adding a few more comments there to indicate whats going on there?

Thrameos:
I added some additional documentation on the process.  The original java documentation is clear as mud because they completely neglect to mention that the valid code point range has an exclusion in the middle.  Hopefully the diagrams will make the adding, shifting, masking process more clear.

We already have some Python.h dependencies in the common section to deal with numpy.  I would rather avoid it but it is additional work, and I don't think that the common will ever be shared with Ruby again.

Thrameos:
I should mention that the entire jp_encoding.cpp file could have been avoided, but it would have required additional JNI calls for each encoding and decoding, as we would have to request a conversion to and from byte[].  I debated going that route, but felt that it would just cause additional complexity as it could introduce additional java exceptions on a path that we need for handling exceptions.  Thus it would have required more thunk code to ensure fault tolerance.

marscher:
Thanks again!

PR #351, opened 2018-10-17
Revised customizers

Thrameos:
This patch replaces the current customizer system with simple classes with annotations to denote the methods to use for customizing.   

This pull may require additional documentation or additional work to fit with the style of the module.  The previous system was not documented.

### DIFF
--- a/jpype/_jboxed.py
+++ b/jpype/_jboxed.py
@@ -29,9 +29,11 @@ else:
 
 __all__ = []
 
-# Boolean is special because in python True and False are singletons,
-# thus it is not possible to make a wrapper properly act as a bool
+
+@_jcustomizer.JImplementationFor("java.lang.Boolean", base=True)
 class _JBoxedBoolean(int, _jobject.JObject):
+    # Boolean is special because in python True and False are singletons,
+    # thus it is not possible to make a wrapper properly act as a bool
     def __new__(cls, *args):
         if len(args) != 1:
             raise TypeError("Invalid arguments")
@@ -45,12 +47,21 @@ class _JBoxedBoolean(int, _jobject.JObject):
     __eq__ = int.__eq__
     __ne__ = int.__ne__
 
+    if _sys.version_info < (3,):
+        __cmp__ = int.__cmp__
+    else:
+        __lt__ = int.__lt__
+        __gt__ = int.__gt__
+        __le__ = int.__le__
+        __ge__ = int.__ge__
+
     def __str__(self):
-        if int(self)==0:
+        if int(self) == 0:
             return str(False)
         return str(True)
 
 
+@_jcustomizer.JImplementationFor("java.lang.Byte", base=True)
 class _JBoxedByte(int, _jobject.JObject):
     def __new__(cls, *args):
         if len(args) != 1:
@@ -65,7 +76,16 @@ class _JBoxedByte(int, _jobject.JObject):
     __eq__ = int.__eq__
     __ne__ = int.__ne__
 
+    if _sys.version_info < (3,):
+        __cmp__ = int.__cmp__
+    else:
+        __lt__ = int.__lt__
+        __gt__ = int.__gt__
+        __le__ = int.__le__
+        __ge__ = int.__ge__
 
+
+@_jcustomizer.JImplementationFor("java.lang.Short", base=True)
 class _JBoxedShort(int, _jobject.JObject):
     def __new__(cls, *args):
         if len(args) != 1:
@@ -80,7 +100,16 @@ class _JBoxedShort(int, _jobject.JObject):
     __eq__ = int.__eq__
     __ne__ = int.__ne__
 
+    if _sys.version_info < (3,):
+        __cmp__ = int.__cmp__
+    else:
+        __lt__ = int.__lt__
+        __gt__ = int.__gt__
+        __le__ = int.__le__
+        __ge__ = int.__ge__
 
+
+@_jcustomizer.JImplementationFor("java.lang.Integer", base=True)
 class _JBoxedInteger(int, _jobject.JObject):
     def __new__(cls, *args):
         if len(args) != 1:
@@ -95,7 +124,16 @@ class _JBoxedInteger(int, _jobject.JObject):
     __eq__ = int.__eq__
     __ne__ = int.__ne__
 
+    if _sys.version_info < (3,):
+        __cmp__ = int.__cmp__
+    else:
+        __lt__ = int.__lt__
+        __gt__ = int.__gt__
+        __le__ = int.__le__
+        __ge__ = int.__ge__
 
+
+@_jcustomizer.JImplementationFor("java.lang.Long", base=True)
 class _JBoxedLong(_long, _jobject.JObject):
     def __new__(cls, *args):
         if len(args) != 1:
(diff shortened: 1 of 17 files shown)


marscher:
The code is much more pythonic, thanks. Can you sort out the problem with Python2, then we are ready to merge?

Thrameos:
Yeah.  I will sort it out.  It appears to be a difference in how the name binding works between python2 and python3.

In python 3, this is acceptable. 

```
class _JList(object):
    def __jclass_init__(cls):
        # Replace addAll method with the one from _JList
        type.__setattr__(cls, 'addAll', _JList.addAll)
        type.__setattr__(cls, 'removeAll', _JList.removeAll)
```

in python2 it wants the `__jclass_init__` to read as 

```
class _JList(object):
    def __jclass_init__(cls):
        # Replace addAll method with the one from _JList
        type.__setattr__(cls, 'addAll', _JList.__dict__['addAll'])
        type.__setattr__(cls, 'removeAll', _JList.__dict__['removeAll'])
```

I am thinking that this level of monkey patching is way too much, so perhaps I should have it use a helper class.

```
class _JList(object):
    # Helper is a proxy for getting, setting, or replacing items in the class dictionary.
    # it can't have map syntax as it needs to do cross version dictionary patching.
    def __jclass_init__(cls, helper):  
        # Replace addAll and removeAll method with the one from _JList
        helper.copy(JList, ('addAll', 'removeAll'))
```

This eliminates the need to getting the syntax of the setattr right and lets us perform required checks so we can prevent problems.  But it doesn't seem very pythonic.  Any suggestions?  

The `__jclass_init__` is supposed to be the post loader that gets called after the class is created to monkey patch out java methods that are problematic.  Maybe it should be `__jclass_new__`?

marscher:
can't you avoid using the internal dict by directly patching the method in the class?
```
class _JList(object):
    def __jclass_init__(cls):
        # Replace addAll method with the one from _JList
        cls.addAll = JList.addAll
```

marscher:
I like the proxy solution. My proposal does not work.

Thrameos:
I wished your solution would work.  However, there are two parts that prevent it.   JClass structures are closed to prevent accidentally attempting to set field or method as a user.  I have been burned too many times by code that looks like this...

Java code:
``` 
public MyClass
{
   public int someGlobal = 1;
}
```

Python code:
```
MyClass = JClass('MyClass')
MyClass.someGloble = 2
```

As java classes are closed and accessing or setting a variable which does not exist it should be an error in jpype.  Further this discourages setting something on an instance which would technically work, but as passing an object to and from java returns a new java object proxy, any extra data would be lost as it only exists on the original wrapper.  Thus we have overridden the `__getattribute__` and `__setattr__` on all jpype classes to render them closed to modifications.

So setting something on the closed class requires an extraordinary action like `type.__setattr__`.  

The second issue is that python2 and python3 mean different things when the call `v=Foo.bar`.  In python2 it returns a method which is bound the to the class `Foo`.  Thus it is unless to put into the dictionary of another class.  In python 3, it returns a function which can be used anywhere.  

Having rendered the classes closed and then having to deal with syntax differences between versions the only clean solution would be the proxy.  

I will try to come up with a reasonable syntax, but please feel free to make suggestions on the syntax when I push it in.  

So here are some options to consider.   

Option 1) Use the jpype namespace to put helper functions for any class patching.
```
@JImplementationFor('java.util.List')
class JList(object):
   def __jclass_init__(cls):
        jpype.JClassCopyMethods(cls, JList, ('addAll', 'removeAll))
```
I don't like this as it pollutes the namespace badly.  It won't be clear where the functions can be used and for what purpose.

Option 2) Hand the user a proxy class with magically monkey patching powers. 
```
@JImplementationFor('java.util.List')
class JList(object):
   def __jclass_init__(cls, dproxy):
        dproxy.copy(JList, ('addAll', 'removeAll))
```
This at least contains where we can do the modifications.  And we have the ability to change the implementation of the proxy if we need to have compatibility with different versions or implementation of python.  But then we have document the capabilities for an object which is pretty well hidden.

Option 3) Limit the scope of the power.  Instead of giving the user an all powerful monkey patching handle, I could limit it to do just a few specific actions which are keyed off of global fields.   This would be much like module exports.
```
@JImplementationFor('java.util.List')
class JList(object):
   # these methods will hide all java implementations for this class and all subclasses. 
   __jclass_overrides__ = ('addAll', 'removeAll')
```
This reduces what we can do to the class to a very limited and controlled set of actions.  We can always grant them more powerful functions later if we need them.  The documentation is limited to just those actions we need to support.

Option 4)  Use annotations to mark functions with attributes we can used to trigger the specific actions we need.
```
@JImplementationFor('java.util.List')
class JList(object):
   @JMethod(sticky=true)
   def addAll(obj, values):
      pass

   @JMethod(sticky=true)
   def removeAll(obj, values):
      pass
```
The result is more consistent with the other annotations we are doing, but this means more stuff in the global namespace.  

Do you have any preferences between the options?

Thrameos:
@marscher any opinions on the direction?

Thrameos:
Okay I am going to try for the annotation based method.  I will use `@JOverload` as we are going to need it for Proxies and Extends as well.  So using the same keyword for all three will make sense.  There are no places where customizers, proxies, and extends are on the same object.

Thrameos:
Made some more progress on this one, but apparently I have another issue with python 2.  Will have to wait until I have access to machine with multiple versions installed.

Thrameos:
Okay puzzled a portion of this one out.  There is an instability in the mro algorithm.  No matter what I tried in terms of the overrides it was randomly failing.  I am going to switch it to something which pushes Interfaces down relative to explicitly specified base classes.  Assuming that works the pull request will be ready.

Thrameos:
@marscher this pull request should be ready for review.  Once merged, I will use this to finish the proxy patch.

jpype/_jboxed.py
```diff
@@ -45,12 +47,21 @@ def __new__(cls, *args):
     __eq__ = int.__eq__
     __ne__ = int.__ne__
 
+    if _sys.version_info < (3,):
+        __cmp__ = int.__cmp__
```
marscher: could you please add a comment here, why you do not use cmp in python3?

marscher:
I like the decorator approach, it is clean and simple. Thanks a lot! Please merge on your own behalf.

jpype/_jboxed.py
Thrameos: I removed it from Python 3 because it is not being used.   At least according to the guides I could find that special method does nothing in Python 3.  I had it in both versions but I got a bug that showed up only in Python 2 and then did some research to discover the reason the bug didn't trigger in Python 3 was that nothing was calling it.  

https://portingguide.readthedocs.io/en/latest/comparisons.html

I don't really have an issue placing the `__cmp__` method in Python 3, but there is no guarantee that `int` will have a `__cmp__` at points in the future. 

Do you think that a comment to this effect is required in the code?

marscher:
On 02.11.18 19:11, Karl Nelson wrote:
> Thrameos commented on this pull request.
> 
> 
> 
>> @@ -45,12 +47,21 @@ def __new__(cls, *args):
>      __eq__ = int.__eq__
>      __ne__ = int.__ne__
>  
> +    if _sys.version_info < (3,):
> +        __cmp__ = int.__cmp__
> 
> I removed it from Python 3 because it is not being used.   At least according to the guides I could find that special method does nothing in Python 3.  I had it in both versions but I got a bug that showed up only in Python 2 and then did some research to discover the reason the bug didn't trigger in Python 3 was that nothing was calling it.  
> 
> https://portingguide.readthedocs.io/en/latest/comparisons.html
> 
> I don't really have an issue placing the `__cmp__` method in Python 3, but there is no guarantee that `int` will have a `__cmp__` at points in the future. 
Thanks a lot for the reference.
> 
> Do you think that a comment to this effect is required in the code?
No, now it is documented here.
> 
>

PR #353, opened 2018-10-17
[appveyor] upgrade chocolatey

marscher:


marscher:
looks like jdk8 pkg is broken in chocolatey. Does anybody have a clue where to report this?

ConstantineLignos:
[Ran into this PR while trying to sort out tests for https://github.com/jpype-project/jpype/pull/272.]

I haven't had to use it before, but there's info on the reporting process here: https://chocolatey.org/docs/package-triage-process#the-triage-process

I think there was an update to the jdk8 package right around the time you say it was broken, so it could be fixed already. I'm not sure whether your comments were from before or after the latest, but other people seem to have reported issues at https://chocolatey.org/packages/jdk8 around the same time.

Thrameos:
I identified the issue.  The ant is now 1.10.5, but the appveyor script has ANT_HOME set to 1.10.1.  It should just be a matter of changing the ANT_HOME in the appveyor.yml.  Not sure why we require ANT_HOME, but I think it did not work without it.

Thrameos:
Is it possible to get the cinst to put ant in a particular location rather than the default?  I looked and it was not clear to me if cinst supports `--install-path` or some other option.   I am not clear why it is that the cinst is not putting the ANT_HOME in but without it setting it manually it does not appear to be set at all.  Unfortunately with the appveyor CI not working well I was leaving changes to these scripts for now.

PR #356, opened 2018-10-23
JArray fails on Classes and strings in 0.6.3

Thrameos:
Somehow during some work on the JArray a bug was introduced that removed two required paths for the JArray constructor.  issubclass does not work on strings and thus the first path failed.  The test for a class instance was also incorrect.  It appears that we had some holes in our test coverage that was not checking this pattern.

I corrected the logic and added appropriate tests for each of the patterns.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ environment:
     # /E:ON and /V:ON options are not enabled in the batch script intepreter
     # See: http://stackoverflow.com/a/13751649/163740
     CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\appveyor\\run_with_env.cmd"
-    ANT_HOME: "C:\\ProgramData\\chocolatey\\lib\\ant\\apache-ant-1.10.1"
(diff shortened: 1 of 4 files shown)


Thrameos:
The build fail appears to be a failure in cinst ant.  I will try to make the failure more explicit.  Not sure how to address the ANT_HOME problem.

marscher:
Maybe one can pass a static install location and point the env variable to it?

Am 25. Oktober 2018 17:52:40 MESZ schrieb Karl Nelson <notifications@github.com>:
>The build fail appears to be a failure in cinst ant.  I will try to
>make the failure more explicit.  Not sure how to address the ANT_HOME
>problem.

Thrameos:
The odd thing is that the ant choco install file sets the ANT_HOME, so the issue appears to be that the variable is set in one place and then does not appear when we go to run.  Either that or the original setting of ANT_HOME was not required.  I am going to do a few test pulls with different settings.  But I am not really sure when appveyor is running or not.  The travis build is running every time a push occurs but the appveyor is only appearing occasionally.   It may be a configuration problem.

PR #360, opened 2018-10-26
Object field bug for devel.

Thrameos:
This is a pull request for a set of tests regarding object fields.  This patch simply adds the tests but does not correct the issue.


Thrameos:
Automatic type conversions to object are now working properly.

PR #361, opened 2018-10-26
Object field bug for master.

Thrameos:
This adds a set of tests for the use of an object field.  It simply adds tests and does not correct the problem at the current time.  Not sure if I will fix this on master.

### DIFF
--- a/test/harness/jpype/values/FieldsTest.java
+++ b/test/harness/jpype/values/FieldsTest.java
@@ -9,4 +9,5 @@ public class FieldsTest
   public static int intField;
   public static float floatField;
   public static double doubleField;
+  public static Object objectField;
 }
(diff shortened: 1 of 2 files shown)


marscher:
Maybe it is a good time for synchronizing devel and master again?

Thrameos:
I have about 4 more pulls before devel is up to a release point.  I would like to get customizers, np,array fix, ducktype primitive support, and the classloader code.  There are a few things on master that will also need to be brought into devel.  Assuming everything works we can pull devel to master and cut a release. 

I have just finished the jpype.compiler support which allows dynamically generated java classes to be imported.  It isn't quite able to extend a class from java fully with a python object, but it does allow the user to define a class that overrides and then directs methods to python manually.

PR #365, opened 2018-10-30
Compiler

Thrameos:
This is a work in progress pull request for the classloader and inline compiler.  I was also stuck on an airplane for 5 hours so I added a quickguide to using JPype.  There is still some material that needs to be finished up.   Assuming you are happy with the direction, I can finish the remaining items and squash it.


native/java/org/jpype/classloader/JPypeClassLoader.java
```diff
@@ -0,0 +1,130 @@
...
+public class JPypeClassLoader extends ClassLoader
+{
+  TreeMap<String, byte[]> map = new TreeMap<>();
+  static private JPypeClassLoader instance;
+
+  public static JPypeClassLoader getInstance()
+  {
+    if (instance == null)
+    {
+      JPypeClassLoader.instance = new JPypeClassLoader(getSystemClassLoader());
```
marscher: just out of curiosity, can people hook in here, if they want an URL class loader?

native/java/org/jpype/classloader/JPypeClassLoader.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private JPypeClassLoader(ClassLoader parent)
+  {
+    super(parent);
+  }
+
+  public void importClass(String name, byte[] code)
+  {
+    name = name.substring(0, name.length() - 6).replaceAll("/", ".");
```
marscher: what happens here, if the the class is not found? Or name.length is < 6?

native/java/org/jpype/classloader/JPypeClassLoader.java
```diff
@@ -0,0 +1,130 @@
...
+   * Loads a class from the class loader.
+   *
+   * @param name is the name of the class with java class notation (using dots).
+   * @return the class
+   * @throws RuntimeException on a fail.
+   */
+  @Override
+  public Class findClass(String name)
+  {
+    byte[] data = map.get(name);
```
marscher: if name is not in the map, it will throw something else (eg. not a runtime exception), right?

native/java/org/jpype/classloader/JPypeClassLoader.java
```diff
@@ -0,0 +1,130 @@
...
+    if (data == null)
+    {
+      for (Map.Entry<String, byte[]> entry:map.entrySet())
+      {
+        System.out.println(entry.getKey());
```
marscher: debug leftover?

native/java/org/jpype/ref/JPypeReference.java
```diff
@@ -1,32 +1,25 @@
-package jpype.ref;
+package org.jpype.ref;
 
 import java.lang.ref.PhantomReference;
 import java.lang.ref.ReferenceQueue;
 
 public class JPypeReference extends PhantomReference
 {
-  private long mHostReference;
+  long mHostReference;
```
marscher: why change visibility here?

testMemoryCompiler.py
```diff
@@ -0,0 +1,182 @@
...
+#   a pointer to the python object and a reference to hold scope.
+#
+# Issues:
+# - Generating a code segment to cover all the cases will be difficult.
+# - Exceptions need to be passed through, thus constructors with exceptions
+#   require complete specification.
+# - Extending a generic will be challenging as reflection does not provide
+#   everything in terms of names at least in some versions.
+# - Writting code stubs that apply for many versions of java may be an issue.
+#   I doubt that I can make it work for 1.6.
```
marscher: Is java6 still being used a lot? We can think of cutting old pig tails then.

testMemoryCompiler.py
```diff
@@ -0,0 +1,182 @@
...
+def JExtends(javaClass, *args, **kwargs):
+    def f(cls):
+        JExtendsDefineClass(javaClass, cls)
+        return cls
+    return f
+
+from java.util import ArrayList
+
+@JExtends(ArrayList)
+class MyList(object):
```
marscher: pretty cool stuff! :+1:

doc/quickquide.html
```diff
@@ -0,0 +1,558 @@
+
```
marscher: is still already part of the rendered documentation? Do you mind converting this via pandoc to markdown as it is easier to read in an editor? If the formatting will be screwed, we can just keep the html.

native/java/org/jpype/classloader/JPypeClassLoader.java
Thrameos: Yeah.  There is a problem in the build system such that a change in java is not always changing the thunk code.  I need to resolve that as it is causing me to trace down errors that don't actually exist.

native/java/org/jpype/ref/JPypeReference.java
Thrameos: It was requiring multiple accessor to change something which was just a package level variable.   Thus to save space I trimmed out anything which should not be seen on the outside.  The entire JPypeReference class is likely to be changed into a private internal class.

testMemoryCompiler.py
Thrameos: Not as far as I am aware.  I am a bit worried about the effects of Java 9 and the changes to generics.

doc/quickquide.html
Thrameos: I just did a quick hack to make the docs in py.  I didn't have access to the internet so I couldn't look up options.  I will try to see if pandoc can be created instead of html.

Thrameos: I looked a bit at converting this with pandoc.  I haven't used it before.  It may be possible though markdown does not have great table support.  I will give it a try when I have a good connection.

native/java/org/jpype/classloader/JPypeClassLoader.java
Thrameos: Fixed.

native/java/org/jpype/classloader/JPypeClassLoader.java
Thrameos: Fixed.

native/java/org/jpype/classloader/JPypeClassLoader.java
Thrameos: Fixed.

native/java/org/jpype/classloader/JPypeClassLoader.java
Thrameos: Documentation added.

PR #368, opened 2018-11-02
Transferring changes from master.

Thrameos:
This update is to pull in selected changes from master that were not applied to devel.


Thrameos:
I added back some of the tests from main.  Discovered a minor bug which was resolved.

PR #369, opened 2018-11-02
Tests for duck typing of primitives.

Thrameos:
Adds a bunch of tests having to do with type conversion.  This is part of the currently out of date duck typing patch for primitives.  Most of these tests fail currently.  I will revise it with the corresponding code, but not until the head catches up because the type conversion code touches code that from the other pulls preventing merging for now.

### DIFF
--- a/native/common/jp_baseclasses.cpp
+++ b/native/common/jp_baseclasses.cpp
@@ -62,27 +62,15 @@ JPMatch::Type JPObjectBaseClass::canConvertToJava(PyObject* pyobj)
 	}
 
 	// Let'a allow primitives (int, long, float and boolean) to convert implicitly too ...
-	if (JPPyInt::check(pyobj))
-	{
-		JP_TRACE("implicit int");
-		return JPMatch::_implicit;
-	}
-
-	if (JPPyLong::check(pyobj))
-	{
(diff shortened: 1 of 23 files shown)


Thrameos:
This pull request adds duck typing to all uses of primitive types.  This allows native numpy types as well as user defined types to be treated properly when calling java methods.  Part of it uses the concept of "checkIndexable" to distinguish between integer and floating point types so that we can handle types that define both `__int__` and `__float__`.    With this and the array pull request and the numpy array pull request we are much better able to handle numpy.  

There may be minor side effects that result from the change of order when considering types for conversion, though those changes are outside of our current test bench.   I have a large conversion level api which will handle the cases for objects which will handle java `Instant`/`Path` and python equivalent types.  But that will have to come post 0.7.0.

Thrameos:
Ouch that appveyor CI is really slow!  Most of the 20 minutes is being spent on installing the jdk.  Got to be a faster way to do that.

marscher:
I have fixed this on master I'd guess, so you can just copy it over.

Thrameos:
Okay I will try to pull it over from master.

marscher:
it is just the appyeyor.yaml file

PR #372, opened 2018-11-04
Numpy array bug

Thrameos:
This pull request is for fixing a bug with handling of numpy arrays.  Currently the array conversion code when setting from a range assumes that any type that fits into a MemoryView can be directly copied into a java array without any type conversion.  This results in a number of incorrect conversions taking place when handling numpy types.  

This pull request is currently incomplete as it currently just includes the tests that demonstrate the bug.  Once complete this will be ported back to master if needed.

### DIFF
--- a/native/common/include/jp_primitive_common.h
+++ b/native/common/include/jp_primitive_common.h
@@ -94,8 +94,24 @@ public:
 template <typename jarraytype, typename jelementtype, typename setFnc>
 inline bool setRangeViaBuffer(JPJavaFrame& frame,
 		jarray array, int start, uint length,
-		PyObject* sequence, setFnc setter)
+		PyObject* sequence, int npy_type, setFnc setter)
 {
+#ifdef HAVE_NUMPY
+	JPPyObject ref;
+	// Arrays need to be checked
+	// FIXME This will still allow mismatched byte order and 
+	// irregular array types pass.  
+	if (PyArray_Check(sequence) && PyArray_TYPE(sequence) != npy_type)
+	{
+		// If we can't cast then fall back to standard methods
+		if (!PyArray_CanCastSafely(PyArray_TYPE(sequence), npy_type))
+			return false;
+
+		// Otherwise, create a casted array
+		sequence = PyArray_Cast((PyArrayObject*) sequence, npy_type);
+		ref = JPPyObject(JPPyRef::_call, sequence);
+	}
+#endif
 	JPPyMemoryViewAccessor accessor(sequence);
 	if (!accessor.valid())
 		return false;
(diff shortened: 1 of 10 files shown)


marscher:
1. is not an option.
2. we could use the type information for setting too, it should be a template pattern I'd guess.

I think the orignal idea to use memoryviews was
1. Cython uses these as well or an extension of it.
2. As you said, gain ducktyping, eg. everythings which is buffered can be treated without copying around massive data.

So to workaround these conversion errors, we can, if we know we have a numpy array, we can do explicit type checking upon setting.
When it is a raw buffer, we have to assume the programmers are adults and know how to do conversions. I think the Python buffer protocol only uses byte arrays, so this can be directly mapped to ByteBuffer in Java, or if the underlying data type is something else, but mappable to Java, it can be casted within Java, right? So having a Python buffer containing signed int64 eg. can be mapped to a LongBuffer.

https://docs.oracle.com/javase/7/docs/api/java/nio/package-summary.html

Thrameos:
I am fine with this approach. Still playing with the numpy casting methods.

Thrameos:
Hmm.  It appears that if we compile without NumPy and still have MemoryViews compiled it will automatically try to convert even if the type is incompatible.  We really should just require numpy as trying to support these operations without it would lead to bad behaviors.   Optional compiling with and without a package means in effect having two separate distributions which would be incompatible.

Is it possible to require numpy for this package from 0.7.0 on?

marscher:
On 11.11.18 21:17, Karl Nelson wrote:
> Hmm.  It appears that if we compile without NumPy and still have MemoryViews compiled it will automatically try to convert even if the type is incompatible.  We really should just require numpy as trying to support these operations without it would lead to bad behaviors.   Optional compiling with and without a package means in effect having two separate distributions which would be incompatible.
> 
> Is it possible to require numpy for this package from 0.7.0 on?
> 
We have numpy support for the provided binaries. But I still like the
idea of opting out NumPy - although it is available ubiquitous nowadays.

You are absolutely right, that we should avoid different behaviors
depending on the implementation. Maybe the stdlib array package can do
the same job as NumPy. If it is too much of a hassle we can also depend
on it in the next release.

Thrameos:
I merged this one so that I can pull it for the duck typing patch which was more extensive.  This one was much lighter than I expected as all I required was some guard routines at the start of the MemoryView API call.  I don't like the use a string check for "numpy.ndarray"  but that was the only way that I could make it function.   The efficiency is pretty poor when running without numpy as it will do everything using sequence API if the data type is from numpy.  I still recommend that we just make jpype require numpy going forward.  I can do a lot better integration between the two in terms of efficiency if I can guarantee access to numpy types.  There was one type still using the MemoryView API so I can't strip it out.  

There is a lot of edge cases such as numpy arrays with weird strides, byte order issues, or misalignment.  I would fix those as well, but I do not know how to create such bad things using the numpy api.   I think we will have to get someone to submit some additional tests showing what is broken before I can take that on.

PR #377, opened 2018-11-13
Work on compiler.

Thrameos:
This is the first hack at the compiler api.  It still has some rough edges though most of those have to do with the java language rather than anything we control.  Package names are required to create public classes.  Private classes do not produce a default constructor so one must be declared.   Performance is horrible in terms of time as the java compiler has a huge overhead to spin up.  But when you have no other choice but to compile a class to implement something due to how a class was designed in java then this has to be an option.

It also appears to be having a weird interaction with the leak checker on my system.  I don't believe that it causing leaks, but it appears the compiler causes a lot of garbage collection actions that foul up the leak checker.  I may need to put some tear down in the test cases to try to clean up its mess.

I updated the documentation a bit (more needed) and added test cases to make sure it actually works outside of my system.

### DIFF
--- a/jpype/compiler.py
+++ b/jpype/compiler.py
@@ -1,0 +1,98 @@
+# *****************************************************************************
+#   Copyright 2018 Karl Nelson
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# *****************************************************************************
+
+from . import _jclass
+from . import _jexception
+
+__all__ = ['JCompiler', 'JSyntaxError']
+
+class JSyntaxError(Exception):
+    def __init__(self, msg, code, diag):
+        self._code = code
+        self._diag = diag
+        Exception.__init__(self, msg)
+
+    diagnostics = property(lambda self: self._diag, None)
+    code = property(lambda self: self._code, None)
+
+
+class JCompiler(object):
+    """ Front end for java compiler.
+
+    Thic class is used to compiler short pieces of java code to create a
+    dynamic class.  Compiling classes is slow and thus this should only 
+    be used in cases in which the Java API requires extension of a class
+    to implement behavior.
+
+    Calling the compiler will produce a new class which is returned to 
+    the user.  This class can be used like any native class subject to
+    restrictions. As the java class does not belong to any jar it can
+    not be deserialized.
+
+    Restrictions:
+        * The java compiler must be installed on the system.
+        * Public classes must have a package name.  They may not be in 
+        the default package.
+        * Private classes must have a defined public constructor or they cannot
+        be instantiated.  Only public classes have default constructors.
+        * Classes are loaded using the JPype internal memory class loader
+        and thus may not be able to import access classes defined in other 
+        external classloaders.
+
+    Todo:
+        * Compiler options are not exposed.
+
+    """
+
+    def __init__(self, *args):
+        # Use the custom class loader
+        classLoader = _jclass.JClass("org.jpype.classloader.JPypeClassLoader")
+        Class = _jclass.JClass('java.lang.Class')
+
+        # Get a class from jpype class loader
+        cls = Class.forName("org.jpype.compiler.MemoryCompiler",
+                            False, classLoader.getInstance())
+
+        # Convert it to a python class
+        self._memoryCompiler = _jclass.JClass(cls)()
+
+    def available(self):
+        return self._memoryCompiler.isAvailable()
+
+    def compile(self, className, code):
+        """ Compile a class with current compiler options.
+
+        Args:
+            className(str): Class name to be compiled including the package name.  This should be 
+               dot seperated.
+            code(str): java code to be compiled typically represented with a multiline string using
+               regular expression encoding.
+
+        Returns:
+            A java class wrapper for the newly created class or throws an exception on failure.
+
+        """
+        if not self._memoryCompiler.isAvailable():
+            raise OSError("Compiler is not available")
+
+        cls = self._memoryCompiler.compile(className, code)
+
+        if cls == None:
+            dc = self._memoryCompiler.getDiagnostics().getDiagnostics()
+            raise JSyntaxError("Failed to compile.", code, dc)
+
+        return _jclass.JClass(cls)
--- a/native/java/org/jpype/compiler/MemoryCompiler.java
+++ b/native/java/org/jpype/compiler/MemoryCompiler.java
@@ -32,6 +32,12 @@ public class MemoryCompiler
 {
   public DiagnosticCollector<JavaFileObject> diagnostics;
 
+	public boolean isAvailable()
+	{
+    JavaCompiler javac = ToolProvider.getSystemJavaCompiler();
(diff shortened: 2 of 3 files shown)


marscher:
Have you measured that the gc is interfering with performance or is that a conjecture?
Where do I find the "leak checker" code?

Thrameos:
On my test system I got the following failure when running nose manually.  Testing the leak checker by itself and the compiler by itself generated no issues.   Not sure what the interaction is, but it does appear to happen on at least one system.

Running tests individually.
```
$ python3 -m nose test/jpypetest/leak.py
.....
----------------------------------------------------------------------
Ran 5 tests in 4.501s

OK

$ python3 -m nose test/jpypetest/compiler.py
..
----------------------------------------------------------------------
Ran 2 tests in 1.359s

OK
```

Running all tests together.

```
$ python3 -m nose test/jpypetest/
.................................................................................................................................................................................................S...FFF.F...................................SSSSSSSSSSSSS.....................................................
======================================================================
FAIL: testClassLeak (test.jpypetest.transplant_class.<locals>.C)
----------------------------------------------------------------------
  File "/home/nelson85/devel/open/jpype/test/jpypetest/leak.py", line 147, in testClassLeak
    self.assertFalse(self.lc.memTest(classFunc, 5000))
AssertionError: True is not false
-------------------- >> begin captured stdout << ---------------------

  Pass0: 101.580800 -1048.576000  - 227790848 216006656 208729320
  Pass1: 9.011200 524.288000  - 227835904 218628096 211350760
  Pass2: 9.830400 0.000000  - 227885056 218628096 211350760
  Pass3: 8.192000 0.000000  - 227926016 218628096 211350760
  Pass4: 9.011200 104.857600  - 227971072 219152384 211875048
  Pass5: 6.553600 0.000000  - 228003840 219152384 211875048
  Pass6: 3.276800 0.000000  - 228020224 219152384 211875048
  Pass7: 11.468800 0.000000  - 228077568 219152384 211875048
  Pass8: 9.830400 104.857600  - 228126720 219676672 212399336
  Pass9: 13.926400 -104.857600  - 228196352 219152384 211875048


--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: testCtorLeak (test.jpypetest.transplant_class.<locals>.C)
----------------------------------------------------------------------
  File "/home/nelson85/devel/open/jpype/test/jpypetest/leak.py", line 151, in testCtorLeak
    self.assertFalse(self.lc.memTest(lambda: ctorFunc(self.cls), 5000))
AssertionError: True is not false
-------------------- >> begin captured stdout << ---------------------

  Pass0: 11.468800 0.000000  - 228327424 219676672 212399280
  Pass1: 16.384000 104.857600  - 228409344 220200960 212923568
  Pass2: -1402.470400 -104.857600  - 221396992 219676672 212399280
  Pass3: 14.745600 104.857600  - 221470720 220200960 212923568
  Pass4: 6.553600 0.000000  - 221503488 220200960 212923568
  Pass5: 15.564800 0.000000  - 221581312 220200960 212923568
  Pass6: 9.011200 0.000000  - 221626368 220200960 212923568
  Pass7: 10.649600 104.857600  - 221679616 220725248 213447856
  Pass8: 13.107200 -104.857600  - 221745152 220200960 212923568
  Pass9: 12.288000 104.857600  - 221806592 220725248 213447856


--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: testInvokeLeak (test.jpypetest.transplant_class.<locals>.C)
----------------------------------------------------------------------
  File "/home/nelson85/devel/open/jpype/test/jpypetest/leak.py", line 156, in testInvokeLeak
    lambda: invokeFunc(self.string), 5000))
AssertionError: True is not false
-------------------- >> begin captured stdout << ---------------------

  Pass0: 983.040000 0.000000  - 226754560 220725248 213446984
  Pass1: 166.297600 0.000000  - 227586048 220725248 213446808
  Pass2: 13.926400 0.000000  - 227655680 220725248 213446808
  Pass3: 8.192000 0.000000  - 227696640 220725248 213446808
  Pass4: 20.480000 0.000000  - 227799040 220725248 213446808
  Pass5: 13.107200 0.000000  - 227864576 220725248 213446808
  Pass6: 9.011200 104.857600  - 227909632 221249536 213971096
  Pass7: 8.192000 0.000000  - 227950592 221249536 213971096
  Pass8: 11.468800 0.000000  - 228007936 221249536 213971096
  Pass9: 14.745600 0.000000  - 228081664 221249536 213971096


--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: testStringLeak (test.jpypetest.transplant_class.<locals>.C)
----------------------------------------------------------------------
  File "/home/nelson85/devel/open/jpype/test/jpypetest/leak.py", line 143, in testStringLeak
    self.assertFalse(self.lc.memTest(stringFunc, 5000))
AssertionError: True is not false
-------------------- >> begin captured stdout << ---------------------

  Pass0: 9.011200 0.000000  - 228204544 221249536 213970984
  Pass1: -1053.491200 0.000000  - 222937088 221249536 213970984
  Pass2: 8.192000 0.000000  - 222978048 221249536 213970984
  Pass3: 7.372800 0.000000  - 223014912 221249536 213970984
  Pass4: 13.107200 0.000000  - 223080448 221249536 213970984
  Pass5: 11.468800 0.000000  - 223137792 221249536 213970984
  Pass6: 9.011200 0.000000  - 223182848 221249536 213970984
  Pass7: 7.372800 0.000000  - 223219712 221249536 213970984
  Pass8: 5.734400 0.000000  - 223248384 221249536 213970984
  Pass9: 9.830400 0.000000  - 223297536 221249536 213970984


--------------------- >> end captured stdout << ----------------------

```

Thrameos:
Our leak checker (`test/jpypetest/leak.py`) tries to create resources and then measures the effect on the memory footprint of the JVM.  It is not actually using a reference queue to check if a resource has been leaked.   However, on windows it appears that the compiler allocated a number of very large blocks that interact with the garbage collector.  Thus rather than observing a steady allocation of memory we see the GC freeing or reusing these large chunks creating the appearance that our memory operations are producing a leak.

To get an accurate check of a leak we would need to different approach.  This one is just a observational test that depends on assumptions about how the GC will do its work when we force collection.  A reference queue would allow us to monitor the lifespan of an individual object and confirm that it has been collected.  But as we create internal structures that may not be exposed we can't catch those leaks with a reference queue approach.   

The "best" solution I have right now would to place the compiler test and the leak checker into separate processes so that they have there own sandbox.   I don't have a good pattern though for spawning a test from within a test bench.   The "dirty" solution is to place an "x" in front of the compiler test so that it appears after the leak checker.

Thrameos:
@marscher this code appears to be working now.  There were two main issues.  

The first is the java compiler is not available if only the runtime library is found.  This somewhat limits the usefulness of the compiler module.  

The second is that leak checker would trigger a large garbage collect of the javac classes we created during the compiler test.  I managed to find a pattern that loaded the leak checker in a separate jvm.  This resolved the other remaining issue.  This pattern can also be used to test out startJVM thus allowing us to test the behaviors such as when the jvm is not found or the start up options are bad.

Assuming that no major issues are found with this or the other pull requests, we should be able to cut a 0.7 release.

Thrameos:
I am somewhat disappointed that sufficient portions of the compiler are not included in the run time version.  I was hoping to be able to allow for extension of java classes, but that requires overriding methods and directing the overrides to Python.   There may be another way to skin this cat by creating the required byte code directly.  But that is beyond my current skill level.

Thrameos:
Looking around it seems like it may be possible to achieve extension of a Java class using an assembler like lilac.  I am going to investigate some more and see if I can find a suitable assembler to include in JPype for this purpose.   I guess I should split this patch into two pieces.  The compiler piece may need to be scrapped as I am not sure we want to provide a feature with marginal availability.   The fixes to the leak checker on the other hand will improve the stability of the CI and should be kept.

Thrameos:
Rejecting for now.

Issue #375, opened 2018-11-12
getClassPath interprets the path on import


Thrameos:
I get different behaviors depending on the order for 

```
import os
os.environ['CLASSPATH']='mypath'
import jpype
print(jpype.getClassPath())
```

and

```
import os
import jpype
os.environ['CLASSPATH']='mypath'
print(jpype.getClassPath())
```

Apparently the CLASSPATH is  being cached at the start rather than interpreted when the getClassPath is called.

marscher:
Agreed it would add more flexibility, when the paths is read when 
getClassPath is invoked.

PR #380, opened 2018-11-15
Fix for classpath issue.

Thrameos:
Fix for #375

### DIFF
--- a/jpype/_classpath.py
+++ b/jpype/_classpath.py
@@ -4,21 +4,11 @@ import glob as _glob
 
 __all__ = ['addClassPath', 'getClassPath']
 
-_CLASSPATHS = set()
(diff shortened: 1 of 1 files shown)


marscher:
You now allow duplicate items on classpath, is this intended?

Thrameos:
As a path it should preserve order. And at minimum should have a defined order. I suppose that I should unique it to remove duplicates.

marscher:
On 15.11.18 13:35, Karl Nelson wrote:
> As a path it should preserve order. And at minimum should have a defined order. I suppose that I should unique it to remove duplicates.
> 
> 
The order preservation is a perfect argument. I do not think that
duplicates will hurt in any way.

PR #381, opened 2018-11-15
Proxy api rework

Thrameos:
Applying duck typing to our proxy implementation allows for annotations to be used to create proxy class.  This adds `@JImplements` and reuses `@JOverride`.  It does not yet do runtime checking to make sure all methods are implemented and does not yet have support of name mangling.  

This patch also fixes a bug in which java.lang.Object did not accept proxies for conversion.  More testing is required before this is complete.  But I want to put it through the CI to make sure it works on multiple python versions and to discuss how the api looks.

### DIFF
--- a/jpype/_jproxy.py
+++ b/jpype/_jproxy.py
@@ -18,7 +18,7 @@
 import collections
 import sys
 
-__all__ = ["JProxy"]
+__all__ = ["JProxy", "JImplements"]
 
 import _jpype
 from . import _jclass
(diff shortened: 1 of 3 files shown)


Thrameos:
Initial testing passes.   Needs a complete test bench before it is perfect.  I added support to verify all abstract methods are implemented.  Still need to work on handling conflicting overloads, but that is a low priority.

Thrameos:
I think we can ignore the fail on cygwin.  The issue is missing "crypt.h" included in "Python.h" meaning it is a problem with cygwin image.

Thrameos:
This one will be applied after the documentation pull as there is a conflict in the userguide.

PR #385, opened 2018-11-19
Fixing "attribute name must be string"

KOLANICH:


KOLANICH:
>Does this work with unicode names on python2?

IDK, I don't use python 2.

Thrameos:
Okay I will have to try to set up some horrible code in the harness that contains unicode characters.  We can run it through the CI and see what python does with it.  I suspect it is going to barf on both python3 and python2 given that Java allows "letter-like" while python is strictly letters.  

I ran a quick check.   This is valid code in Java.
```
public class Test2
{
  static int Ü=1;
  static int £=2;
  static int ¥=3;
  static int ₧=4;
  
  static void do£()
  {
    System.out.println(Test2.₧);     
  }
  
  static public void main(String[] args)
  {
    Test2.do£();
  }
}
```

But python dies on most of them.
```
#!/usr/bin/python
# -*- coding: utf-8 -*-
Ü=1 # Valid Java/Python 3, invalid Python 2
£=2 # Valid Java, invalid python 3
¥=3 # Valid Java, invalid python 3
₧=4 # Valid Java, invalid python 3
```





Not that this is an issue for your pull request.   Simply an unrelated bug with the differences between identifiers in the two languages.

KOLANICH:
Have you tried to access them with `getattr`?

Thrameos:
They can definitely be part of the dictionary so `globals()` and `getattr` works.  They just can't be accessed except by direct string lookup.

```
globals()['Ü']=1

Ü
Out[85]: 1

globals()['£']=2

£
  File "<ipython-input-87-be4437dba19e>", line 1
    £
    ^
SyntaxError: invalid character in identifier
```

PR #386, opened 2018-11-19
Fixes interference with pickle.py

### LINKED ISSUE #384: JPype importer interferes with https://github.com/python/cpython/blob/2708578736d1aa15685495e9b94b827a8e185a8c/Lib/pickle.py#L94
https://github.com/python/cpython/blob/2708578736d1aa15685495e9b94b827a8e185a8c/Lib/pickle.py#L94

I guess we need to move them into an own namesapace, like `jtypes`

KOLANICH:
Fixes #384

### DIFF
--- a/jpype/imports.py
+++ b/jpype/imports.py
@@ -189,7 +189,7 @@ class _JImport(object):
             return jtype
 
         # If the java class does not exist, throw a ClassNotFound exception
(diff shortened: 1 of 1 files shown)


marscher:
@Thrameos Now it makes sense, dosen't it?

KOLANICH:
Not sure though that it is the right place to use this exception type.

PR #414, opened 2019-05-10
Work on the documentation.

Thrameos:
This patch fixes minor issues with the development branch such as copying of properties for customizers and then documents most of the changes to using jpype in the userguide.  There are still some stray symbols in the module such as `nio` that I have yet to track down.  I also have not yet tested jpype.beans module to see if the previous autoproperty behavior is working.  Hopefully this won't break anything for python 2.7.


Thrameos:
@marscher any objections to the documentation?  There are three unrelated lines but the rest is trying to get the documentation to match our current development head.  Once this pull is in I will merge the Proxy, Beans and a portion of the Compiler used to improve the leak checker.  That should put us at a release candidate.  I can then push this to my production server where it will get hit for a week.  Assuming nothing comes up we should then be able to do a 0.7.0 release.  But we will need to be prepared for a few minor releases as it gets pushed out to distibutions especially for missing files in the dist which I do not catch.  Do you have enough time availability for that?

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+
+This document describes the guts of jpype.  It is intended 
+lay out the architecture of the jpype code to aid intrepid lurkers 
+to develop and debug the jpype code once I am run over by a bus.
+
+
+History
+~~~~~~~
+
+When I started work on this project it had already existed for over 10 years.
```
marscher: you could mention who you mean by "I", e.g. your github name etc.

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+diagram.  Thus that was reworked to something more in line which could be 
+safely completed without disturbing other layers.  The multilanguage 
+abstraction layer was already perced in multiple ways for speed.  However, 
+as this interwove throughout all the kit it was a terrible lift to remove
+and thus required gutting the Python layer as well to support the operations
+that were being performed by the HostRef.  
+
+The remaining codebase is fairly slim and reasonably streamlined.  This 
+rework cut out about 30% of the existing code and sped up the internal
+operations.  The Java C++ interface matches the Java class hierachy.
```
marscher: i really like your history section! It gives me a great overview of what you have achieved.

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+internal class wrapper.  This class wrapper holds the reflection api used to create
+the class methods and fields.  This field is stored as the ``__javaclass__``
+in the class wrapper.  As the class wrapper is used to create an object instance
+``__javaclass__`` also appears in the objects.  Only objects that have a ``__javaclass__``
+and lack a ``__javavalue__`` are treated as class wrappers for the purposes of 
+duck typing.
+
+Because the Java class is both an object and type, we used the duck typing to 
+allow the class pointer to be converted into a class instance.  This is exposed
+as property called ``class_``.  The this ``class_`` property is the equivalent of the
```
marscher: "The this" replace with "The"

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+duck typing.
+
+Because the Java class is both an object and type, we used the duck typing to 
+allow the class pointer to be converted into a class instance.  This is exposed
+as property called ``class_``.  The this ``class_`` property is the equivalent of the 
+.class member of a Java class name or ``getClass()`` of a class instance.  As it 
+is an object instance of ``java.lang.Class`` it can be used for any reflection 
+needs by the user.  However, when working with the jpype core module, on
+needs to be aware of that the this class instance is not available until 
+the key wrappers are created.  See bootstrapping for further details.
```
marscher: please add a cross-ref to bootstrapping (if there is a section).

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+``jstring``
++++++++++++++++++++++++++++++++
+
+For most practical purposes Java strings are treated as objects.  However, they 
+also need to be able to interact with Python strings.  In the previous version,
+strings were automatically converted to Python strings on return.  This resulted
+in rather strange behavior when interacting with the methods of ``java.lang.String``
+as rather than getting the expected Java object for chaining of commands, the 
+string object would revert to Python.  To avoid this fate, we now require string
+objects to be convert manually with the ``str()`` method in Python.  There are
```
marscher: This string handling is an API breakage isn't it? Is it really mandatory?

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+When dealing with these remember they are called typically three ways.   The
+user calls with the specified arguments to create a resource.  The factory
+calls the ``__new__`` method when creating an instance of the derived object.  And
+the C++ wrapper calls the method with internally construct resource such as
+``PyJPClass`` or ``PyJPValue``.  Deciding which of the three ways it has been called
+is usually simple, but it does constrain the operation of the factories as
+conflicts in the three paths would lead to a failure.  Forcing keyword arguments
+for one of the paths could be used to resolve the dependency.
+
+For example, if ``JClass`` was called with a string argument it from the user.
```
marscher: "it originates from the user"?

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+modules global space.  This functionality is now split into a number 
+of classes.  These classes each have a constructor that is used to create
+an instance which will correspond to a Java resource such as class, array,
+method, or value.
+
+``PyJPModule`` module
+~~~~~~~~~~~~~~~~~~~~~~
+
+This is the front end for all the global functions required to support the 
+Python native portion.  Most of the functions provided in teh module are
```
marscher: typo and two spaces

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+
+
+Python referencing
+~~~~~~~~~~~~~~~~~~
+
+One of the most miserable aspects of programming with CPython is the relative 
+inconsistancy of referencing.  Each method in Python may use a Python object or steal
+it, or it may return a borrowed reference or give a fresh reference.  Similar
+command such as getting an element from a list and getting an element from a tuple 
+can have different rules.  Because the was a constant source of bugs requiring
```
marscher: because the -> because there

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+``call_``
+  This policy both steals the reference and verifies there were no errors
+  prior to continuing.  Errors are promoted to exceptions when this reference 
+  is created.
+
+If we need to pass an object which is held in a smart pointer to Python
+which requires a reference, we call keep on the reference which transfers
+control to a ``PyObject*`` and prevents the pointer from removing the reference.
+As the object handle is leaving our control keep should only be called the
+return statement.
```
marscher: This is very cool. Do we have rigorous memory leak checking? I can only think of one legacy test, which uses java.lang.Strings, but as you correctly mention, there are multiple sources of errors, when using the wrong mode in wrapping the Python objects.

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+Invocation and Fields
+++++++++++++++++++++++
+
+To convert a return type produced from a Java call, each type needs to be
+able to invoke a method with that return type.  This corresponses the underlying
+JNI design.  The methods invoke and invokeStatic are used for this purpose.  
+Similarly accessing fields requires type conversion using the methods
+``getField`` and ``setField``.
+
+Instance verses Type wrappers
```
marscher: versus? :)

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+~~~~~~~~~~~~~~~
+
+In addition to the type wrappers, there are several support classes. These are:
+
+``JPTypeManager``
+  The typemanager serves as a dict for all type wrappers created during the 
+  operation.  
+
+``JPReferenceQueue``
+  Lifetime manager for Java and Python objects.
```
marscher: Is the reference queue still required as you wrote that every object is wrapped in some kind of smart pointer?

marscher: forget about this comment... I should have read further.

doc/develguide.rst
Thrameos: Sure will do.

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+The native implement of the proxy call is the only place in with the pattern
+for reflecting Python exceptions back into Java appears.
+
+As all proxies are ties to Python references, this code is strongly tied to
+the reference queue.
+
+``JPClassLoader``
+++++++++++++++++++
+
+This code is responsible for loading the Java class thunks.  AS it is difficult
```
marscher: AS -> As

doc/develguide.rst
Thrameos: Thanks.

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+with converting this unusual encoding into something that Python can understand.
+
+The key method in this module is transcribe with signature ::
+
+  std::string transcribe(const char* in, size_t len,
+      const JPEncoding& sourceEncoding,
+      const JPEncoding& targetEncoding)
+
+There are two encodings provided, ``JPEncodingUTF8`` and ``JPEncodingJavaUTF8``.  
+By selecting the source and traget encoding transcribe can covert to or
```
marscher: covert -> convert

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+with.  As the point of jpype is to communicate with Java, it is possible to 
+directly communicate with the jpype Java internals.  These can be imported
+from the package ``org.jpype``.  The code for the Java layer is located in
+``native/java``.  It is compiled into a jar in the build directory and then converted
+to a C++ header to be compiled into the ``_jpype`` module.
+
+The Java layer currently houses the reference queue, a classloader which can
+load a Java class from a bytestream source, the proxy code for implementing
+Java interfaces, and a memory compiler module which allows Python to directly 
+create a class from a string.
```
marscher: That is also a very nice feature! However I do not understand why we would users to fiddle around with the ref queue.
To have the possibility of a custom class loader has been a long ongoing request.

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+Tracing
+---------
+
+Because the relations between the layers can be daunting especially when things
+go wrong.  The CPython and C++ layer have a built in logger.  This logger 
+must be enabled with a compiler switch to activate.  To active the logger, touch 
+one of the cpp files in the native directory to mark the build as dirty, then
+compile the ``jpype`` module with: ::
+
+     python setup.py --enable-tracing devel
```
marscher: I like this approach better than setting this particular environment variable

doc/develguide.rst
```diff
@@ -0,0 +1,656 @@
...
+method to operate in JNI.  It is better to have specialized code in Java which
+preforms large tasks such as collecting all of the fields needed for a type 
+wrapper and passing it back in a single call, rather than call twenty different
+general purpose methods.  This would also vastly reduce the number of ``jmethods``
+that need to be bound in the C++ layer.
+
+The world of JVMs is currently in flux.  Jpype needs to be able to support 
+other JVMs.  In theory, so long a JVM provides a working JNI layer, there 
+is no reason the jpype can't support it.  But we need loading routines for 
+these JVMs to be developed.
```
marscher: :+1: :+1: :+1: :+1: :+1: :+1:

marscher:
Can you please remove the double space at each sentence start?
The following regular expression should do the job:

```
s/\.\ {2}/\ /g
```

marscher:
what about the imports.rst file? It is not included in the current index.

doc/develguide.rst
Thrameos: In the previous system is was impossible to keep a string on the java side for any length of time.  Even if you you were just passing around a large string from java to java, it was forced to come back to python.  There was a global switch that was supposed to keep java strings on the java side, but it was filled with bugs.  The revised version is exactly as the switch state was supposed to do.  You call a method and get back a java.lang.String and if you want to pay to convert it call str() on it.

The previous string behavior was very much regrettable.   It was impossible prototype simple string codes because methods things were the user clearly wanted to get back a java string like substring() instead gave pystring instead.  

Mode switching like we had was impossible to maintain for any realistic use  (so that one was out) and our previous default and getting pystring every time a string prohibited a number of patterns.  

Can it be undone, yes (with effort).  In the code returning jvalues back from a method call and in the field string we can add a special case that forces the string to decode back to python object.  As the path string path is no longer internally different than  any other handler (part of the code simplification), we would need to check each of the paths.   

So the question is what will this break.  
- If the user got a string  and expected it to have python methods, it fails with method not found.
- Jstrings convert themselves automatically when going to string concatenation or formatting so no breakage here.
- left equivalence and right equivalence  are not equal.  Jstring knows how to compare to a pystring properly, but pystring does not know how to compare to jstring. 
- Use as dictionary keys, does not work as expected especially if mixed.
    
So as to what you were asking."is it mandatory."  I think so, the previous design left a lot of bad behavior in which ever mode was selected by the global switch.   We can special case it to push this down the road and make old code but it still leaves a lot of holes in the API.  I tested all of my code base with the new behavior and found only two places that needed a str() call added.

doc/develguide.rst
Thrameos: Access to the reference counter is needed to help audit the garbage collecting.  You can ask the reference counter how may python objects are being held currently.

marscher:
Please do not undo anything, I just wanted to understand the new 
behavior. It makes total sense not to enforce the conversion, if it is 
not needed.

In the final changelog we should mention this.

marscher:
I see. That can be very useful.

doc/develguide.rst
Thrameos: The leak checker code does a pretty good job of creating lots of java objects held by python and then destroyed.   If the code path hits the path of python wrapper layer in that case (tuples, dicts, etc) then it would have been caught.  We are a bit weaker for python leaks in general because I can't test the python wrapper directly.  I applied this smart pointer strategy because more often than not I messed something up.  Unfortunately, it only crashed if I undercounted references and overcounts just leak.

doc/develguide.rst
Thrameos: As a tone deaf, dyslexic with bad spelling, I will just trust you on this.

doc/develguide.rst
Thrameos: The only tricky part is sometimes if tool is already built it will skip the source building entirely and ignore the switch.  Still working on that.

Thrameos:
@marscher Thanks for the very thorough review.  I made the changes as suggested and did another full read through myself.  I think imports.rst should be in the index, but I haven't touched it yet.  Also I need to test that these files actually go into the dist.

Thrameos:
@marscher Any idea how to get two multiline preformated code blocks to exist side by side in a rst table?  I would like to get the quickstart guide to rst, but I couldn't figure out how to formulate that.   Assuming that  it were possible, I can make the document generator produce that output.

marscher:
On 5/15/19 7:04 PM, Karl Nelson wrote:
> @marscher <https://github.com/marscher> Any idea how to get two 
> multiline preformated code blocks to exist side by side in a rst table? 
> I would like to get the quickstart guide to rst, but I couldn't figure 
> out how to formulate that. Assuming that it were possible, I can make 
> the document generator produce that output.

Sorry, I have no idea. RST is very obscure sometimes. Maybe you can use 
Markdown for this purpose and convert it with pandoc to get the desired 
formatting?

Thrameos:
I found one example using rst and a code-block.  I made the change, but I need to get a rst viewer so I can see what it does the output.

Thrameos:
Okay I think it is working.  At least it renders something viewable in restview.   Hopefully it will parse with other RST formatters.

Thrameos:
Merging this pull for now.  I can clean  up other issues in the documentation later.  In particular finishing the rather hairy change log.

PR #418, opened 2019-05-13
Beans

Thrameos:
This pull updates the old beans pattern customizer.  I moved it to an optional package as it is one of the key sources of problems with users.  It is occasionally useful so I reactivated it.

It was the only customizer remaining using the undocumented JCustomizer protocol.   Currently, customizers are applied to a specific class in the tree and then propagate to the children.  Customizers can now be applied retroactively.  There are still so rough edges if someone applied more than one customizer to the same class.  But I don't expect many people to exercise it and the behavior is documented as last customizer rules so it should be workable.

This likely has a conflict with one of the earlier patches.  Thus I will resolve it once patches catch up to this point.


Thrameos:
@marscher Okay I have caught the head back up with my current development branch.  This PR deals with the old properties customizer.  It was pretty much a broken features that generated bug reports for the ability to do something more than one way.  But it is possible that someone was making heavy use of the misfeatures, so I added an optional module to support it.  As an optional module it has to work even when it is called late so I implemented retroactive customizers.  You don't need to review this is you don't have time, but I wanted to give you the chance to object.  

 My delivery to my customer is coming up in two weeks so I am going to make a hard press to get us to a 0.7.0 release. (8 months sitting without funding, and they want delivery a month after turning money back on.)   I pushed the reset of changes to 0.8 project.  The only outstanding item after this is the changelog that I am working on later this week.  Once we have a candidate ready I will make some trial dist to make sure all the files required for use are in the dist and make a PR for missing stuff.

jpype/_jcustomizer.py
```diff
@@ -136,9 +113,72 @@ def _applyStickyMethods(cls, sticky):
...
+                setter(rename, members[p])
+
+            setter(p, v)
+        elif p == "__new__":
+            setter(p, v)
+
+def _applyAll(cls, method):
+    applied = set()
+    todo = [cls]
+    while len(todo)!=0:
```
marscher: you could just write: while todo

Thrameos: done.

PR #421, opened 2019-05-18
Work of documentation.

Thrameos:
More work in the endless pursuit of something approaching accuracy.


Thrameos:
@marscher I made another giant pass through the documentation.  I got sphinx to pull our doc strings and moved the reference guide stuff from the userguide to the api reference.  Now we don't have the same information in multiple places (as often).  I also added a minor feature to the JPackage class to fix a long standing peeve of mine that it would happily create packages for non compliant package names.

This patch should be ready to go assuming it passes checks.   Assuming it an the beans pull requests go in we are ready to start testing packaging and swap devel to master.

marscher:
Very well. I do not have time to read it all now. Thank you very much for wrapping it all up and even adding proper api docs. That was a long missing feature.

marscher:
Please merge on your on behalf

PR #424, opened 2019-05-19
Conversion check

Thrameos:
This code adds the ability to audit the type conversion code.  This code was then used to print out the conversion table and identify errors.


Thrameos:
A quick test of the conversions

```
Primitives and Boxed
                 Name                  Type boolean char short int long float double j.l.Character j.l.Byte j.l.Short j.l.Integer j.l.Long j.l.Float j.l.Double j.l.Object
0                None         Py 'NoneType'                                                      I        I         I           I        I         I          I          I
1                True             Py 'bool'       X    I     I   I    I     I      I             E        E         E           E        E         E          E          I
2               False             Py 'bool'       X    I     I   I    I     I      I             E        E         E           E        E         E          E          I
3              int(1)              Py 'int'       I    I     I   I    I     I      I             E        E         E           E        E         E          E          I
4              int(2)              Py 'int'       I    I     I   I    I     I      I             E        E         E           E        E         E          E          I
5         float(12.3)            Py 'float'       E          E   E    E     I      X                      E         E           E        E         E          E          I
6      JBoolean(True)        Java 'boolean'       X                                                                                                                      I
7            JChar(1)           Java 'char'            X                                         E                                                                       I
8           JShort(1)          Java 'short'                  X                                                      E                                                    I
9             JInt(1)            Java 'int'                      X                                                              E                                        I
10           JLong(1)           Java 'long'                           X                                                                  E                               I
11        JFloat(2.1)          Java 'float'                                 X                                                                      E                     I
12       JDouble(3.5)         Java 'double'                                        X                                                                          E          I
13  j.l.Boolean(True)    Java 'j.l.Boolean'       I                                                                                                                      I
14   j.l.Character(1)  Java 'j.l.Character'            I                                         X                                                                       I
15       j.l.Short(1)      Java 'j.l.Short'                  I                                                      X                                                    I
16     j.l.Integer(1)    Java 'j.l.Integer'                      I                                                              X                                        I
17        j.l.Long(1)       Java 'j.l.Long'                           I                                                                  X                               I
18     j.l.Float(2.1)      Java 'j.l.Float'                                 I                                                                      X                     I
19    j.l.Double(3.5)     Java 'j.l.Double'                                        I                                                                          X          I
20     j.l.Integer(1)    Java 'j.l.Integer'                      I                                                              X                                        I
21         JObject(1)       Java 'j.l.Long'                           I                                                                  X                               I
22       JObject(1.1)     Java 'j.l.Double'                                        I                                                                          X          I
23   JObject(1, JInt)    Java 'j.l.Integer'                      I                                                              X                                        I
```
The part regarding boxed types is incorrect.  In Java we can have both a conversion to int and Integer without conflict.  I don't think our internal 4 level match can fit this case.  

Checking the other objects
```
Objects and casts
                               Name                  Type char j.l.String j.l.Object j.l.Number j.l.Class j.l.Exception [I [D
0                        bytes('c')            Py 'bytes'    I          X          I
1                          str('c')              Py 'str'    I          X          I
2                        str('abc')              Py 'str'               X          I
3                      JString('b')     Java 'j.l.String'               X          I
4                      JObject('a')     Java 'j.l.String'               X          I
5                        j.l.Object         Java 'JClass'                          I                    X
6                 j.l.Object.class_      Java 'j.l.Class'                          I                    X
7                      j.l.Object()     Java 'j.l.Object'                          X
8                         JObject()     Java 'j.l.Object'                          X
9               JObject(1, JObject)     Java 'j.l.Object'                          X
10           JObject(None, JObject)     Java 'j.l.Object'                          X
11                    j.l.Throwable  Java 'j.l.Throwable'                          I
12                    j.l.Exception  Java 'j.l.Exception'                          I                                  X
13                      j.l.Byte(1)       Java 'j.l.Byte'                          I          I
14  JObject(j.l.Byte(1),j.l.Number)     Java 'j.l.Number'                          I          X
15                           dict()             Py 'dict'
16                           list()             Py 'list'                                                                I  I
17                            [1,2]             Py 'list'                                                                I  I
18                          [.1,.2]             Py 'list'                                                                E  I
19            JArray(JInt)([1,2,3])          Java 'int[]'                          I                                     X
```

Seems fine to me.

Thrameos:
This pull request was originally just to help me to revise the documentation by printing out the conversion table automatically.  But it lead to discovery of a number of bugs that needed to be resolved.

- Cast operations were hitting a bad path during ``__init__`` leading to an error.
- Cast operations were stealing a global reference leading to objects disappearing out from under themselves.
- Char was incorrectly being matched to float on the conversion table.
- Arbitrary length strings were matching to char.
- Object casting was using the wrapper type rather than the object type when casting leading to the issue it was not possible to upcast properly.

@marscher As this is vital code and it is deep in the guts I am going to commit this code without review as every other pul should depend on this one.  I performed all the usual requirements such as autopep8 and testing so I think it is pretty safe.  There are a few conversions in the test that are still worrisome, but they are issues discovered rather that introduced.  Char from float and Char from long strings should both be `explicit` rather than `none` to help support proxies.  And the block regarding boxed appears to be all `explicit` rather than `implicit`.   Our current matching system is not completely matched as we need a `preferred` vs `implicit` to break up ties types.  Further, we are really harsh about java primitive types, we can afford to have some level of implicit conversions in the table.  As it stands we can have casting for an overload fail from minor changes in the signature.  That said, nothing here appears to introduce a problematic behavior that we can't live with.  At worst we force a unnecessary cast, as opposed to accepting  something we would rather not happen.   Please run the `testConversion.py` script and report anything you see as a problem to issues and I will take care of it.

PR #426, opened 2019-05-20
[travis] add osx and python 3.7

marscher:


marscher:
I think python 3.7 and python 2.7, with/out numpy should be sufficient.

Thrameos:
Closing this request as #436 covered it

PR #427, opened 2019-05-20
Master

Thrameos:
The big pull request.


Thrameos:
I am going to recommend disabling the tests that fail in cygwin.  The problem in cygwins interaction with the subprocess and not a problem in jpype.

Thrameos:
Damn 3 tests away from passing on PyPy 3.   One is a false trigger due to PyPy memory model.   The other two are a difference in the api regarding char.  I don't need it, but I still want it.  The greed!

marscher:
Thanks again! You really changed the game for this library!

PR #428, opened 2019-05-21
Uses a cache to remove ant and jdk dependency.

Thrameos:
As requested this pull request adds a cache for the jars.  To build the jar when it changes use

```
python setup.py --enable-jar develop
```

The jar and class file must be copied back manually to the cache.  Not sure if the native source directory is a good place for the jar.   If you have a better spot feel free to move it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ matrix:
       env: NUMPY="--global-option=--disable-numpy"
   
 install:
-  - python setup.py sdist
+  - python setup.py --enable-build-jar sdist
   - pip install dist/* $NUMPY
   - pip install -r test-requirements.txt
   - ant -f test/build.xml
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-recursive-include native *.h *.xml *.java
+recursive-include native *.h *.xml *.java *.jar *.class
 recursive-include examples *
 include *.rst doc/* LICENSE
 # include everything under test/
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -63,13 +63,14 @@ environment:
       CONDA_PY: "3-latest"
       ARCH: x86_64
 
-    - PYTHON: "C:\\Miniconda34-x64"
-      CONDA_PY: "3.4.3"
-      ARCH: x86_64
-
-    - PYTHON: "C:\\Miniconda34"
-      CONDA_PY: "3.4.3"
-      ARCH: x86
+      ### Dropping 3.4 build due to problem with setuptools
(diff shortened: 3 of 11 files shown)


marscher:
Thank you! I would not put the binaries to git, but just the code generating them.

Thrameos:
Okay.  I guess I could check if the cache is present and use.   Where should the hook be to copy the build/lib contents to the cache location?

marscher:
I would assume it to be in the spot where setup.py expects these files for shipment, or am I missing something?

Thrameos:
I mean I don't know which setup command to put the command to copy from build/lib to native/jars.   I assume we have to add something to make it happen.  Do you know which stage I need to customize?

marscher:
On 21.05.19 06:50, Karl Nelson wrote:
> I mean I don't know which setup command to put the command to copy from build/lib to native/jars.   I assume we have to add something to make it happen.  Do you know which stage I need to customize?
> 

I think we need it both at build and sdist stages. If we want to ship, 
we need to make sure it is up to date as well as for building.

Thrameos:
I made it such that if the cache is missing it will tell you to run with `--enable-build-jar`.  I just need to get it through the CI and then have sdist copy the jars into place so the distribution has the jars with it.

Sounds like a plan?

Thrameos:
Placing it on the build stage seems bad to me.  If it caches the jars on build when doing development then you can end up with stale jars leading to bugs.  Better to build the jars on the fly up until we are ready to make a shipable copy, then push the jars into the cache.

marscher:
That sounds like a good solution. So just check it during sdist

Thrameos:
It may take me a little while to get the pattern right to get distutils to add a hook to copy from build during sdist.  Distutils is a dark, dark place, and the documentation never seems to shine light.

Thrameos:
Okay I think I finally hit the right notes and got it to pull the files into a cache on sdist and use those files on a copy resulting from the sdist.  The ``enable-build-jar`` is not necessary as it is smart enough to run it on automatically.  Note sure if targets that indirectly use sdist will have bad behavior but that is the mystery of distutils.

Thrameos:
@marscher I made an attempt at updating the installation instructions.  They were really choppy has we are describing multiple methods of installation (binary, source from pip, source from sdist, source from github) all of which have differences in requirements and procedure.   I gave a stab at it.  Assuming you don't see anything wrong I will merge the request tonight.

Thrameos:
I think it would be a good idea to cache the Java portion of the test and add instructions for how to run the test into the instructions.  However, I am out of time for now so I will just submit a ticket and work on that in another pull request.

PR #430, opened 2019-05-21
Fixes #374

### LINKED ISSUE #374: Array classes lack methods
Because of the organization of the tree java array classes do not have methods that would be expected of them in java.  Specifically the following methods are present in java that do not appear in the wrapper.

```
getClass()
clone()
equals(Object obj)
hashCode()
toString()
notify()
notifyAll()
wait(...)
```

`length` property is also missing but is covered by `len(a)` syntax.  Most of these could be replicated at the JArray level.  I am not sure how difficult it would be to make arrays Objects so that the get the proper methods.

Thrameos:
Most of this was already fixed by the rewrite, but we can add a property for length to complete the job.

### DIFF
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -112,6 +112,8 @@ class _JArray(object):
         if values is not None:
             self.__javaarray__.setArraySlice(0, sz, values)
 
+    length = property(lambda self: self.__len__(), None)
+
     def __str__(self):
         return str(tuple(self))
 
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -112,10 +112,12 @@ class _JArray(object):
         if values is not None:
             self.__javaarray__.setArraySlice(0, sz, values)
 
-    length = property(lambda self: self.__len__(), None)
-
     def __str__(self):
         return str(tuple(self))
+
+    @property
+    def length(self):
+        return self.__len__()
 
     def __len__(self):
         return self.__javaarray__.getArrayLength()
--- a/jpype/__init__.py
+++ b/jpype/__init__.py
@@ -51,5 +51,5 @@ def JIterator(it):
 # FIXME these should be deprecated. The old JPackage system is only for
 #  python2 series and generates lots of deceptive classes.  At some point
 #  these two are going to have to go away.
-java = JPackage("java", strict = True)
-javax = JPackage("javax", strict = True)
+java = JPackage("java", strict=True)
+javax = JPackage("javax", strict=True)
--- a/jpype/_classpath.py
+++ b/jpype/_classpath.py
@@ -56,7 +56,7 @@ def addClassPath(path1):
 
     Arguments:
       path(str): 
-    
+
     """
     global _CLASSPATHS
 
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -86,11 +86,13 @@ def _initialize():
 def isJVMStarted():
     return _jpype.isStarted()
 
+
 def _hasClassPath(args):
     for i in args:
         if i.startswith('-Djava.class.path'):
             return True
     return False
+
 
 def startJVM(jvm=None, *args, **kwargs):
     """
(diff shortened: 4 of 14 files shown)


jpype/_jarray.py
```diff
@@ -112,6 +112,8 @@ def __init__(self, *args, **kwargs):
         if values is not None:
             self.__javaarray__.setArraySlice(0, sz, values)
 
+    length = property(lambda self: self.__len__(), None)
```
marscher: Aint that a bit obscure? Why not write it as a usual method returning self.`__len__`? I would agree on the property, if we wanted to define a setter.

marscher:
I totally understand why we want this. My concern was only about readability. I would write

def length(self): return self.‘__Len__‘

Am 21. Mai 2019 17:45:12 MESZ schrieb Karl Nelson <notifications@github.com>:
>Thrameos commented on this pull request.
>
>
>
>> @@ -112,6 +112,8 @@ def __init__(self, *args, **kwargs):
>         if values is not None:
>             self.__javaarray__.setArraySlice(0, sz, values)
> 
>+    length = property(lambda self: self.__len__(), None)
>
>Java programmers would expect to find .length on an Java array.  It is
>a getter only in java.  Unfortunately a setter makes not sense as it
>would mean duplicating the array.  
>This is mainly to help with the typical cut and paste from Java code to
>python and back.  It is included for completeness

Thrameos:
Let see if that pattern makes it work...

Original code in Java.  Like numpy shape, length is a property in Java.  (One of the only ones, no clue why.)
```
public class Test
{
  static public void main(String[] args)
  {
    int[] ja = new int[]{1,2,3};
    for (int i =0; i<ja.length; ++i)
      System.out.println(ja[i]);
  } 
}
```
Translated code, cut and paste with required changes for loop structure.
```
import jpype
from jpype.types import *

jpype.startJVM()

ja = JArray(JInt)([1,2,3])

for i in range(0,ja.length):
    print(ja[i])
```

I get 
```
$ python3 testArrayLength.py
  File "testArrayLength.py", line 9, in <module>
    for i in range(0,ja.length):
TypeError: 'method' object cannot be interpreted as an integer
```

Using the property gives me correct behavior, just like `class_`.  Unfortunately, it is really ugly.  

Maybe better would be 
``` 
  @property
  def length(self):
     return self.__len__()
```

This give proper behavior.   So lets make sure it doesn't break when we try to set it.

```
  File "testArrayLength.py", line 12, in <module>
    ja.length = 10
  File "...\devel\open\jpype\jpype\_jobject.py", line 105, in __setattr__
    attr.__set__(self, value)
AttributeError: can't set attribute
```

Okay good.  Lets see if we can make `class_` work the same way.  We have two copies of this.  One for JClass and the other for JInterface.  (I suppose it would be better on JObject, but that is another experiment)

```
print(JClass('java.lang.String').class_)
print(JArray(JInt).class_)
```

Gives
```
class java.lang.String
class [I
```

Try to set it ...
```
  File "testArrayLength.py", line 12, in <module>
    JArray(JInt).class_ = "why would you do that"
  File "...\devel\open\jpype\jpype\_jclass.py", line 160, in __setattr__
    (self.__name__, attr))
AttributeError: int[] does not have field None
```

Okay it works, but the error message is really horrible.  Looking at the code.... hmm some moron forgot to put a "what-if-not-found-in-search" clause.  Blame log says Karl Nelson... damn it is me again.   

Change to a better message.
``AttributeError: Field 'class_' is not settable on Java 'int[]' object``

Hmm, that was a deeper rabbit hole than I expected.

Thrameos:
Sorry about this.  It was supposed to be a nice simple 1 line pull request.  Instead it when down the rabbit hole of properties, descriptors, closed objects, error reporting, and autopep.  Mind reviewing again to make sure I didn't do damage.

With this patch it now reported field not settable and not found separately.  It also differentiates between operating on a Java Object or a Java Class.

```
Field 'length' is not settable on Java 'int[]' object
Field 'foo' not found on Java 'int[]' object
Field 'class_' is not settable on Java 'int[]' class
Field 'foo' not found on Java 'int[]' class
```

Not sure if it should be `Java object 'int[]'` or `Java 'int[]' object`

marscher:
Very well! Your contractor will be very pleased I'd guess. I can not 
even count the numerous bugs you have fixed since working on this project!

PR #435, opened 2019-05-21
JString as hash key for dict Fixes #434

### LINKED ISSUE #434: JString vs Python string as dict keys
Current Java strings and python strings both act as valid keys, but they have different hashes thus they get stored separately.  It is not clear if giving them the same key will make them play nice together as they have different types.

```
import jpype
from jpype.types import *
jpype.startJVM()

try:
    d=dict()
    d['foo']='a'
    print(d[JString('foo')])
except Exception as ex:
    print(ex)

try:
    d=dict()
    d[JString('foo')]='a'
    print(d['foo'])
except Exception as ex:
    print(ex)
```

Thrameos:
This is an attempt to fix our issues with dict and JString.  This will bring us closer to compatibility with 0.6 series.  (assuming it works)

### DIFF
--- a/jpype/_jstring.py
+++ b/jpype/_jstring.py
@@ -73,7 +73,7 @@ class _JString(object):
         return self.contains(other)
 
     def __hash__(self):
-        return self.hashCode()
+        return self.__str__().__hash__()
 
 
 JString = _jobject.defineJObjectFactory(
--- a/test/jpypetest/values.py
+++ b/test/jpypetest/values.py
@@ -133,3 +133,20 @@ class ValuesTestCase(common.JPypeTestCase):
         self.assertEquals(self.Fields.objectField, 2.6125)
         self.assertIsInstance(self.Fields.objectField,
                               jpype.JClass('java.lang.Double'))
+
+    def testStringDictKey1(self):
+        d = dict()
+        d['foo']='a'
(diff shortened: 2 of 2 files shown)


marscher:
Thanks for addressing this. Do you think caching the hash can be worth the effort?

Am 22. Mai 2019 00:25:08 MESZ schrieb Karl Nelson <notifications@github.com>:
>This is an attempt to fix our issues with dict and JString.  This will
>bring us closer to compatibility with 0.6 series.  (assuming it works)
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/jpype-project/jpype/pull/435
>
>-- Commit Summary --
>
>  * Fixes #434
>
>-- File Changes --
>
>    M jpype/_jstring.py (2)
>    M test/jpypetest/values.py (17)
>
>-- Patch Links --
>
>https://github.com/jpype-project/jpype/pull/435.patch
>https://github.com/jpype-project/jpype/pull/435.diff

Thrameos:
It is not a bad idea to cache it.  Though i suspect most JStrings dont live very long.

Thrameos:
Actually a cache really should not be necessary.  JString caches the string it produces when the conversion is made and gives the same string every time.  If caching were needed to use in a map, then string would already have it.  And as we return the same string each time, any caching python does will already be accomplished.  Thus no need for an extra layer.

PR #436, opened 2019-05-22
Upgrading test bench

Thrameos:
This has been sitting on the list for a long time.  I will try to get everything to switch to pytest and get ``python setup.py test`` to actually work.  Lots of changes are required for this,  dead files need to disappear, and I will need to reduce the size of the test matrix to get work done quickly,


Thrameos:
Okay a run down of what was done here

- Drop dead packages jpype.awt
- Drop old stub test files
- Convert to pytest
- Drop nose
- Drop Python 2.6 and Python 3.4 due to out of date packages
- Moved to current test suite packages
- Add 3.7 to matrix
- Add osx to matrix
- Reduced number of NUMPy builds
- Fixed problems on pytest switch

What failed on.... getting ``python setup.py test`` to run. :cry:

Thrameos:
@marscher This one should replace #426.  I am not sure yet why osx with openJDK 12 is failing test.  The test matrix got reduced a bit as well. 

 Appveyor is being a royal pain in terms of stability right now.  I was going to attempt to get travis to build a windows version in another PR.  Right now this one should be good to go.  

I would like the without numpy to die in the future.  Series 0.7 will be the last I can support with that option as further core work needs to integrated numpy into arrays for speed.  I want the option to lock a piece of Java memory in place and use numpy as the front end to it.   But that will tie the code to numpy forever.

Thrameos:
Okay need to freshen this PR.

Thrameos:
@marscher I won't merge this one without a review as it deals with the CI matrix which is your area.  I know that it is a big pull as there are lots of changes to the test system.  When you get the time to review it, it would help.  Otherwise I will need to keep refreshing it as every change to the test system using the old nose has to be converted again to get to pytest.

Thrameos:
I didn't much like the renaming myself.  But I wanted out of the box pytest to actually work.  For the longest time I could not figure out why it wouldn't work no matter what I did.  Then I mistakenly called it without arguments and it started running something out of the old test stubs.  Once it was clear that we would be going against the flow renaming seemed like the least worst option.

Issue #442, opened 2019-05-24
Environment variables not passed to Java on Cygwin


Thrameos:
While testing for another bug, I discovered that Java does not appear to be getting environment variables.  It appears to affect windows/cygwin with openJDK 1.8.  I am not sure if other platforms are affected. As some are vital to operations, this leads to unacceptable behavior on windows as 
``USERPROFILE`` must be set to get a valid temp directory.

Thrameos:
Problem appears to be confined to Cygwin versions of python.

marscher:
So only Cygwin is broken in that sense?

Thrameos:
Appears so.  I can't see any reason for the problem.  Nothing we are doing would affect the environment.  It must be an interaction with cygwin and Java dll.  I am just going to document it and move on.

Thrameos:
I moved this to the limitations. Way to much trouble to figure out how env is moving behind the scenes.

Issue #441, opened 2019-05-24
Collections customizers not tested


Thrameos:
While hunting for another bug, I discovered that ``__iter__`` was not working in as intended.  It appears that none of the customizers for collections are currently being tested.  Thus bugs are likely.

Thrameos:
One bug was caught and fixed in #443.

PR #443, opened 2019-05-24
Documentation and limitations

### LINKED ISSUE #303: Windows/Cygwin issues
A number of issues have cropped up regarding windows and cygwin with regards to the test suite.  I am documenting them so that we have a clear record of the problems until they can be resolved.

1) Cygwin on 32 bit with python2.7.  Test is disabled because of a well known issue with gcc segfaulting on this platform.
2) Cygwin on 32 bit with python 3.6.  Test fails due to a segfault in the first test case.  It is not clear if this is an issue with the appveyor environment or a bug in the code.  Needs to be replicated outside of that platform.
3) Cygwin on 64 bit with python 2.7. Test fails due to a problem in the multithreaded test.  It is not clear if there issue is with the python build on the platform or the jpype code.  The code for thread locking has issues that differ from the standard idiom, but this is not affecting any other platforms.   The error is reproducible outside of appveyor and may be related to other reported failures. #279
4) Windows on 32 bit with python 2.7 & 3.6.  The appveyor tests successfully but appears to be using the  64 bit jvm.  This is likely a platform issue in which the jvm is being picked up elsewhere as it seems unlikely to succeed with a mixed environment.  Further debugging of this may be required, but as the test is successful it is a low priority.

For now I recommend disabling the faulty platforms and leave this thread open as documentation of known issues until they can be addressed.

### LINKED ISSUE #438: is it possible to unload jvm?
When i start JVM and then shut it down and start a new JVM the previous one still remains in memory . Is it possible to delete the previous one in some way ? Version 0.7 fix this issue ? i'm using a python package that communicates with a java application by using Jpype . This package has a method to call jpype.shutdownJVM() and shutdown the JVM which has been created for the bridge communication . My application opens and closes this bridge multiple times ( by doing that multiple JVM are started and shutdown) and when it does that makes the ram usage increase a lot . I think that the problem are the JVMs which aren't unloaded .

Thrameos:
This pull makes it an OSError to try to restart the JVM.  It fixes an issue with java.util.Map wrapper.  And it documents current issues with starting the JVM, PyPy, and Cygwin in the documentation.
 
Closes #438
Closes #303
Reference #442


doc/userguide.rst
```diff
@@ -597,6 +597,39 @@ Known limitations
 This section lists those limitations that are unlikely to change, as they come
 from external sources.
 
+Restarting the JVM
+~~~~~~~~~~~~~~~~~~
+
+JPype caches many data structures within the JVM. Those resource are still
```
marscher: resources

doc/userguide.rst
```diff
@@ -597,6 +597,39 @@ Known limitations
...
+allocated after the JVM is shutdown as there are still Python objects 
+that point to those resources.  If the JVM is restarted, those stale Python
+objects will be in a broken state and the fresh startiJVM will reallocated the
```
marscher: and the new JVM instance will obtain the references to these resulting in a memory leak.

jpype/_jcollection.py
```diff
@@ -169,7 +169,7 @@ def __len__(self):
         return self.size()
 
     def __iter__(self):
-        return _WrappedIterator(self.keySet().iterator())
+        return self.keySet().iterator()
```
marscher: do we have a test for this one?

Thrameos: See issue #441

PR #444, opened 2019-05-24
Final work on startJVM

Thrameos:
Our startJVM has made a lot of progress, but still was not very friendly.  It didn't report errors and the exception behavior was not well defined.  I have corrected this as much as I could.


jpype/_core.py
```diff
@@ -95,48 +95,84 @@ def _hasClassPath(args):
...
-
-        # Check to see that the user has not set the classpath
-        # Otherwise use the default if not specified
-        if not _hasClassPath(args) and 'classpath' not in kwargs:
-            kwargs['classpath'] = _classpath.getClassPath()
+    args = list(args)
 
-    if 'ignoreUnrecognized' not in kwargs:
-        kwargs['ignoreUnrecognized'] = False
+    # JVM path
```
marscher: this voodoo is not necessary with the old signature. If jvm is already passed via the named keyword, passing it again via kwargs will raise a TypeError.

jpype/_core.py
```diff
@@ -95,48 +95,84 @@ def _hasClassPath(args):
...
+    classpath = None
+    if 'classpath' in kwargs:
+        if _hasClassPath(args):
+            raise TypeError('classpath specified twice')
+        classpath = kwargs['classpath']
+        del kwargs['classpath']
+    elif not _hasClassPath(args):
+        classpath = _classpath.getClassPath()
+
+    if classpath:
```
marscher: handling of unicode is something we should keep.

jpype/_core.py
```diff
@@ -95,48 +95,84 @@ def _hasClassPath(args):
...
+    Parameters:
+     *args (Optional, str[]): Arguments to give to the JVM.
+        The first argument may be the path the JVM.
+
+    Keyword Arguments:
+      jvm (str):  Path to the jvm library file,
         Typically one of (``libjvm.so``, ``jvm.dll``, ...) 
-      *args (Optional, str[]): Arguments to give to the JVM
-      classpath (Optional[string]): Set the classpath for the jvm.
+        Default of None will use ``jpype.getDefaultJVMPath()``
```
marscher: much better docstring!

jpype/_core.py
```diff
@@ -95,48 +95,84 @@ def _hasClassPath(args):
 
 _JVM_started = False
 
-def startJVM(jvm=None, *args, **kwargs):
+def startJVM(*args, **kwargs):
```
marscher: wildcard arguments are a pain! Please keep jvm (also for backward compatibility reasons).

jpype/_core.py
```diff
@@ -95,48 +95,84 @@ def _hasClassPath(args):
...
-    args = list(args)
-    if 'classpath' in kwargs and kwargs['classpath'] != None:
-        args.append('-Djava.class.path=%s' % (kwargs['classpath']))
-
-    _jpype.startup(jvm, tuple(args), kwargs['ignoreUnrecognized'])
+    classpath = None
+    if 'classpath' in kwargs:
+        if _hasClassPath(args):
+            raise TypeError('classpath specified twice')
+        classpath = kwargs['classpath']
```
marscher: you could use pop() to write this in one line.

native/common/include/jp_platform_linux.h
```diff
@@ -41,9 +42,7 @@ class LinuxPlatformAdapter : public JPPlatformAdapter
 
 		if (jvmLibrary == NULL)
 		{
-			std::stringstream msg;
-			msg << "Unable to load DLL [" << path << "], error = " << dlerror();
-			JP_RAISE_RUNTIME_ERROR( msg.str().c_str());
+			JP_RAISE_OS_ERROR_UNIX( errno, path);
```
marscher: on windows you have a human readable string, but it seems like on linux people just see an error code?

jpype/_core.py
Thrameos: I believe it does the same thing here.

I wanted to be able to support 
```
  # old
   jpype.startJVM(getDefaultJVMPath(), "-ea", '-Djava.class.path=%s'%getClassPath())
  # Better
   jpype.startJVM(getDefaultJVMPath(), '-ea', classpath = getClassPath())
   jpype.startJVM(getDefaultJVMPath(), '-ea')
  #  Simple
   jpype.startJVM('-ea')
 ```

File: native/common/include/jp_platform_linux.h (hunk 6 above)
Comment (Thrameos) [reply native/common/include/jp_platform_linux.h]: I was sort of forced to do something non-standard on Windows because the Python version call for windows reporting is optional.   I can do the same  thing for linux if you would like.

File: jpype/_core.py (hunk 2 above)
Comment (Thrameos) [reply jpype/_core.py]: Not sure what I am missing on unicode in this comment.  Perhaps the review was labeled on another line.

**Comment by Thrameos** [comment]
Sorry the jars slipped in due to a missing .gitignore statement.  I will try to resolve it.

File: jpype/_core.py (hunk 4 above)
Comment (Thrameos) [reply jpype/_core.py]: Just to be clear ``jvm`` as the first argument before the ``*args` meant you had to specify the path even if you just wanted to start a jvm with default.  Passing in None there is the alternative, but looked really ugly to me.  I can change it back.  

Also minor issue.... i think the argument should be ``jvmpath`` rather than ``jvm``.  I want to reserve ``jvm`` for when we actually can start more than one machine.

File: jpype/_core.py (hunk 5 above)
Comment (Thrameos) [reply jpype/_core.py]: done.

File: native/common/include/jp_platform_linux.h (hunk 6 above)
Comment (Thrameos) [reply native/common/include/jp_platform_linux.h]: It looks like they should get symmetric errors

```
      case JPError::_os_error_unix:
      {
        PyObject* val = Py_BuildValue("(iz)", m_Error, (string("JVM DLL not found: ") + mesg).c_str());
        if (val != NULL)
        {
          PyObject* exc = PyObject_Call(PyExc_OSError, val, NULL);
          Py_DECREF(val);
          if (exc != NULL)
          {
            PyErr_SetObject(PyExc_OSError, exc);
            Py_DECREF(exc);
          }
        }
        return;
      }

      case JPError::_os_error_windows:
      {
        PyObject* val = Py_BuildValue("(izzi)", 2, (string("JVM DLL not found: ") + mesg).c_str(), NULL, m_Error);
        if (val != NULL)
        {
          PyObject* exc = PyObject_Call(PyExc_OSError, val, NULL);
          Py_DECREF(val);
          if (exc != NULL)
          {
            PyErr_SetObject(PyExc_OSError, exc);
            Py_DECREF(exc);
          }
        }
        return;
      }
```

Thrameos:
I am making some progress on the 0.8 core changes right now, but it is a long path.  I want to gut the rest of the JNI layer out of C++ so that all the brains and object clean up lives in Java.  This is the easy part.  The hard part is to make it work, I need to gut every global variable out of the JPype core.  Unfortunately, the core _jpype module is full of globals.  However, if I succeed in theory we can restart the JVM or even run more than one JVM at a time.

Thrameos:
@marscher Assuming this and the exception pull are good then we are ready for release of 0.7.0.

PR #448, opened 2019-05-25
Pull changes from tcalmant

Thrameos:
These are a set of minor changes from the other major fork of JPype.  Although the master is not mergeable, I scanned the entire source to determine what if anything we have not yet overrun.  

Thus far there were very minor differences in the API.

 -  JProxy uses ``dictionary`` rather than ``dict`` as its kwarg
 -  cygwin contains an extra function for reprocessing the JVM arguments
 -  minor differences in the C++ layer from the branch point
 -  renames and such

I pulled the test bench from that fork into our own and ran it with only the one API difference being hit.
As the code is not mergeable I replicated the relevant changes in the C++ layer.  (removed direct access to Python structure rather than using macro, and error reporting for arrays.)

See Thrameos/jpype#41 and tcalmant/jpype-py3#28

I think that tcalmant fork can safely be repointed back to the org copy.  Now I just need to understand the  [ha_JPype1 ](https://libraries.io/pypi/ha_JPype1) fork.  I personally find it rather antisocial to fork an active project, change the maintainer pointer, add one to the version number and then push it out to the distributions.

### DIFF
--- a/native/common/jp_array.cpp
+++ b/native/common/jp_array.cpp
@@ -71,14 +71,19 @@ JPPyObject JPArray::getRange(jsize start, jsize stop)
 void JPArray::setRange(jsize start, jsize stop, PyObject* val)
 {
 	JP_TRACE_IN("JPArray::setRange");
+
+	// Make sure it is an iterable before we start
+	if (!JPPySequence::check(val))
+		JP_RAISE_TYPE_ERROR("can only assign a sequence");
+
 	JPJavaFrame frame;
 	JPClass* compType = m_Class->getComponentType();
 	unsigned int len = stop - start;
 	JPPySequence seq(JPPyRef::_use, val);
-	size_t plength = seq.size();
+	long plength = seq.size();
 
 	JP_TRACE("Verify lengths", len, plength);
-	if (len != plength)
+	if ((long) len != plength)
 	{
 		// Python would allow mismatching size by growing or shrinking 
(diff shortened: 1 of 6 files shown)


native/common/jp_array.cpp
```diff
@@ -71,14 +71,19 @@ JPPyObject JPArray::getRange(jsize start, jsize stop)
...
+	// Make sure it is an iterable before we start
+	if (!JPPySequence::check(val))
+		JP_RAISE_TYPE_ERROR("can only assign a sequence");
+
 	JPJavaFrame frame;
 	JPClass* compType = m_Class->getComponentType();
 	unsigned int len = stop - start;
 	JPPySequence seq(JPPyRef::_use, val);
-	size_t plength = seq.size();
+	long plength = seq.size();
```
marscher: can size be negative or why change type here? If this was just added to avoid a compiler warning, I'd suggest to change JPPySequence size return type also to a size type (size_t).

marscher:
Thanks for taking a look. It would be best to bundle the effort for this project.

native/common/jp_array.cpp
Thrameos: The problem here was that the error message reported a very large number instead of -1.  I will look at JPPySequence.

Thrameos: Okay I confirmed it was already a jlong so it was supposed to be signed.  The ``size_t`` was incorrect.

PR #453, opened 2019-05-28
Expose old symbol (Fixes #452)

### LINKED ISSUE #452: JPype get_default_jvm_path
I think that the way the trunk looks now we can no longer run:

```
import jpype

path = jpype.get_default_jvm_path()
```

As it now results in this:
```
AttributeError: module 'jpype' has no attribute 'get_default_jvm_path'
```

In 0.6.3 this was runnable and documented, but since #334 the jpype/_core.py __all__ has been refactored so that `get_default_jvm_path` is not exported by default. 

Was this deliberate to move away from get_default_jvm_path -- should we add it back to __all__ to preserve backwards compatibility?

Current:
```
__all__ = [
    'isJVMStarted', 'startJVM', 'attachToJVM', 'shutdownJVM',
    'getDefaultJVMPath', 'getJVMVersion', 'isThreadAttachedToJVM', 'attachThreadToJVM',
    'detachThreadFromJVM'
]
```

Modified???
```
__all__ = [
    'isJVMStarted', 'startJVM', 'attachToJVM', 'shutdownJVM',
    'getDefaultJVMPath', 'getJVMVersion', 'isThreadAttachedToJVM', 'attachThreadToJVM',
    'detachThreadFromJVM','get_default_jvm_path'
]
```

Thrameos:
Expose only symbol for compatibility.

### DIFF
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -37,7 +37,7 @@ from . import _jinit
 __all__ = [
     'isJVMStarted', 'startJVM', 'attachToJVM', 'shutdownJVM',
     'getDefaultJVMPath', 'getJVMVersion', 'isThreadAttachedToJVM', 'attachThreadToJVM',
-    'detachThreadFromJVM', 'synchronized'
+    'detachThreadFromJVM', 'synchronized', 'get_default_jvm_path'
 ]
 
 # See http://scottlobdell.me/2015/04/decorators-arguments-python/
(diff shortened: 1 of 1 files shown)


Thrameos:
@marscher 

We can also consider putting a line in ``jpype/__init__.py`` that reads

```
# Remove old symbol from default module export
__all__.remove("get_default_jvm_path")
```

This would keep the old symbol from being imported on ``from jpype import *``

PR #454, opened 2019-05-28
Added missing clone method for arrays.

Thrameos:
Adds clone() to java arrays.  Clone does not appear in array methods, but is a language construct on the Java side.  I used ``Arrays.copyOf()`` to achieve the same result.  Like Java clone, the copy is shallow.

Example:
```
import jpype
from jpype.types import *

jpype.startJVM()

# Verify clone dups the array
array =JArray(JInt)([1,2,3,4])
print(array.__javavalue__)
print(array.clone().__javavalue__)

# Verify clone is shallow
array =JArray(JInt, 2)([[1,2],[3,4]])
print(array[0].__javavalue__)
print(array.clone()[0].__javavalue__)
```

Gives:
```
$ python3 testClone.py
<java value class [I  value = 0x5555564ad3e0 [I@7440e464>
<java value class [I  value = 0x5555564cbfd8 [I@5acf9800>
<java value class [I  value = 0x5555564d2958 [I@36baf30c>
<java value class [I  value = 0x5555564d2978 [I@36baf30c>
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ language: python
 
 python:
   - '2.7'
-  - '3.3'
   - '3.4'
   - '3.5'
   - '3.6'
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -173,6 +173,23 @@ class _JArray(object):
             return True
 
     def clone(self):
+        """ Clone the Java array.
+
+        Copy the aray and return a new copy. The copy is "shallow" and
+        thus the new array points to the same contents as before. For
+        single dimensional array of primitives, the cloned array is
+        completely independent of the original. For objects or 
+        multidimensional arrays, the new and old array point to the
+        same objects.
+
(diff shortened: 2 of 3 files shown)


Thrameos:
Sure thing.  I will add a test in arrays.

Thrameos:
@marscher  Shall we disable tests on Python 3.3?

marscher:
I think that should be safe. It has been deprecated a long time ago.

Thrameos:
Reviewing the code I see two items that are missing.  We currently don't have a defined way to convert to a python sequence or list.  The current idiom is to call ``x[:]`` which as a side effect produces a sequence or numpy array.  But there is no guarantee that this will always be the case.  Second there isn't a good way to get the component type.  It is possible to access it through reflection but that gives the java.lang.Class rather than the Python type wrapper.  Thus when creating multidimensional arrays one can can't just get the array type and use it to make a new array.

Neither of these are critical to the functioning, but should likely be addressed at some future point.

PR #458, opened 2019-06-01
jvm found raise not support  exception

Hexmagic:
jvm_found raise ```No JVM shared library file (jvm.dll) "
         "found. Try setting up the JAVA_HOME "
                                       "environment variable properly.```, this make me confused,I check my java  	environment variable setting  again and again, finally  after I  found maybe I install wrong version Python3.6(x32),which not compatible with jdk1.8(x64)


Thrameos:
Thanks for the contribution.  I merged your request and will update the documentation.

Issue #463, opened 2019-06-03
Debugging tricks


Thrameos:
Two tricks that I have been using for joint python/java development are attaching debugging data to Java exceptions and use of serialization to produce core dumps.  As these are very strong tools for working with our code they should appear in our code base somewhere.  However, they are completely unlike other tools in our module so I am not sure how to frame them.  Rather than being so part of the python  module, they are sections of pure Java code that can be used as part of any Java project.  Here are the relevant snippets from my projects internal utility library.

The following code is hereby given the public domain.  If it breaks, it you get to keep all the pieces.

The first is a utility intended to cause a failure in an algorithm to dump some context that lead to the fault into a corefile.  This file can be unloaded in JPype and then examined to see what went wrong.  This is useful when doing large batch runs of a process or capturing a failure at a remote site where the user can send the fault file back to the developer.
```
package example;

import java.io.IOException;
import java.io.ObjectOutputStream;
import java.io.OutputStream;
import java.io.Serializable;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.time.Instant;
import java.util.HashMap;
import java.util.zip.GZIPOutputStream;

/**
 * Utility to simplify the pattern for dumping a core file.
 *
 * To enable core dumps, specify the system property with ".dumpCore" appended
 * java.lang.System.setProperty('org.myproject.dumpCore', 'true');
 *
 * <p>
 * Example:
 * <pre>
 * {@code
 *   package bar;
 *
 *   // Make sure the class and all members are serialiable
 *   class Foo implements Serializable
 *   {
 *      // Be sure to set the version UID so that the methods can be
 *      // altered without breaking the core file
 *      public static final long serialVersionUID=-123727878335L;
 *
 *      public void evaluate(Input input)
 *      {
 *        // Place a try block around the entire body of the function
 *        try {
 *          // something may go wrong here
 *          ...
 *        }
 *
 *        // Catch all exceptions that warrent a core or simply Exception
 *        catch (Exception ex)
 *        {
 *          // Create a core dump file condition on a System property.
 *          CoreDump dump = new CoreDump("org.myproject");
 *
 *          // Add each of the required items to produce the problem
 *          dump.add("Foo", this);
 *          dump.add("input", this.input);
 *          dump.add("exception", ex);
 *
 *          // Each of the items will be saved in the core file
 *          dump.write("foo");
 *
 *          // Rethrow the exception to keep the control flow
 *          throw ex;
[...]

PR #465, opened 2019-06-05
Added debugging hints.

Thrameos:
This is a request related to discussions with JPype users at my site in response to #463.  Joe Guensche and Caleb Mattoon contributed a section on attaching the debugger to JPype along with two tricks I posted earlier on the attachment of data to exceptions and use of serialization. The docs are pretty minimal but should be enough to point our users in the right direction.


Thrameos:
Documentation change only.  Self approved as it doesn't change the API.  

Thy spelleng essuez well bi adressed lader.

PR #466, opened 2019-06-06
Add better error messages to jpype.imports

Thrameos:
This has burned me for the last time.  Die! Die! Die!

This pull request improves the error reporting for jpype.imports to try to report wrong version and missing dependencies.   The JClass method we were using gave back the same exception every time thus making it hard to distinguish the source of the problem.  To debug this, the user would then need to call forName to find the actual source of the problem.  This pull request cuts out the middle man and just directly calls forName to get class in the first place.  The performance is slightly worse, but this is a very low frequency operation.

### DIFF
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -95,6 +95,18 @@ def _hasClassPath(args):
             return True
     return False
 
+def _handleClassPath(clsList):
+    out = []
+    for s in clsList:
+        if not isinstance(s, (str, _jtypes._unicode)):
+            raise TypeError("Classpath elements must be strings")
+        if s.endswith('*'):
+            import glob
+            out.extend(glob.glob(s+'.jar'))
+        else:
+            out.append(s)
+    return _classpath._SEP.join(out)
+
 
 _JVM_started = False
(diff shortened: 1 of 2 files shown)


Thrameos:
Bonus.  Wildcards are acceptable at the command line for Java but not recognized by the jvm library at start time.  This pull request fixes that defect.  Note that wildcards function per Java rules.  A trailing wildcard expands to all files fitting "*.jar". 

```
jpype.startJVM(classpath=[
   '../../src/myproj/dist/myproj.jar',
   '../../src/dependencies/*']
   )
```

Thrameos:
This one was to address a concern on stack overflow (as well as my own personal needs).  It has minor API issues as it introduces a glob pattern to the classpath syntax.  I have been running with it for a week without issues.  

@marscher If you are fine with the API implications of reporting the errors and the glob, I will commit it without further review.   Nothing was large enough in this to trip the testbench as we never checked the messages on errors being returned .

marscher:
this globbing pattern will make peoples life easier. Thank you!

PR #469, opened 2019-06-07
Enable option to support legacy conversion of strings.

Thrameos:
Attempt to support the legacy behavior of string conversion on return.  I am not sure how well this matches with the original.  We had no test bench to check before and after with the switch thrown.  For now everything is converting on return regardless of the source requesting conversion (BFFI approach).  This should match the old behavior for methods and likely fields.  But may not be a good match for other paths.  The most problematic are likely the ArrayItem, ArrayRange, and Proxy args, where conversion does not make as much sense.  I can add source tracking flags on the call if we need finer grain controls.

To use the legacy behavior start the JVM with ``convertStrings=True``.  The behavior is currently set once, because I really don't want to try to deal with the issues of having this change during operation in the future especially if multithreading were used.

A reformatter removed some dangling spaces which resulted in some extra lines in the PR.  Sorry.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,7 @@
 *.cpp text eol=lf
 *.c text eol=lf
 *.h text eol=lf
+*.java text eol=lf
 
 *.xml text eol=lf
 *.md text
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -123,8 +123,14 @@ def startJVM(*args, **kwargs):
         cast to Python strings. This option is to support legacy code
         for which conversion of Python strings was the default. This
         will globally change the behavior of all calls using
-        strings, and is not recommended for newly developed
-        code.
+        strings, and a value of True is NOT recommended for newly 
+        developed code.
+
+        The default value for this option during 0.7 series is 
+        True.  The option will be False starting in 0.8. A
+        warning will be issued if this option is not specified
+        during the transition period.
+
 
     Raises:
       OSError: if the JVM cannot be started or is already running.
@@ -177,7 +183,21 @@ def startJVM(*args, **kwargs):
                         (_classpath._SEP.join(classpath)))
 
     ignoreUnrecognized = kwargs.pop('ignoreUnrecognized', False)
-    convertStrings = kwargs.pop('convertStrings', False)
+
+    if not "convertStrings" in kwargs:
+        import warnings
+        warnings.warn("""
+-------------------------------------------------------------------------------
+Deprecated: convertStrings was not specified when starting the JVM. The default
+behavior in JPype will be False starting in JPype 0.8. The recommended setting
+for new code is convertStrings=False.  The legacy value of True was assumed for
+this session. If you are a user of an application that reported this warning,
+please file a ticket with the developer.
+-------------------------------------------------------------------------------
+""")
+
+    convertStrings = kwargs.pop('convertStrings', True)
+
 
     if kwargs:
         raise TypeError("startJVM() got an unexpected keyword argument '%s'"
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -101,7 +101,7 @@ _JVM_started = False
 def startJVM(*args, **kwargs):
     """
     Starts a Java Virtual Machine.  Without options it will start
-    the JVM with the default classpath and jvmpath.  
+    the JVM with the default classpath and jvmpath.
 
     The default classpath is determined by ``jpype.getClassPath()``.
     The default jvmpath is determined by ``jpype.getDefaultJVMPath()``.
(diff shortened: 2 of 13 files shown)


kyle-downey:
This looks good to me.

Thrameos:
@marscher  I am going to commit this one as being required for user operation.  The only issue I see is that the convertStrings flag is not being noted in the change log.  I will clean that up post merge.

marscher:
What is the default value of string conversion now? Are you setting legacy mode to prevent people need to change something for using 0.7 or not?

Thrameos:
The default mode is to leave Java strings on the JVM side until conversion is requested.  Having strings auto convert by default is a performance issue and handicaps that use of JPype.  However, some existing projects are too heavily dependent on the old behavior.  For those projects they would need to add the ``convertStrings=True`` to the startJVM.   Is this an acceptable solution?

marscher:
Prior breaking these projects I'd suggest to make legacy default for now 
and show a deprecation warning, so people can transition to it smoothly. 
I absolutely agree about the performance considerations.

Thrameos:
Also assuming there this is some project out there that was throwing the switch back and forth because some calls needed to access Java string methods while other parts were using the conversion, will now be forced to chose one policy.  The implications of changing while in operation are too difficult to deal with.

marscher:
I'm fine with the once per jvm boot setting. It is great to have this 
switch in any case!

Thrameos:
I will differ to you on this issue.   I will need to modify my scripts to force it to False for now.  

One question, what setting will the test bench be run under?  I haven't been running anything with the autoconvert mode on and it does a lot more than just hitting methods and fields.  I am a little hesitant to force the testbench to run twice and I am not sure if the autoconvert implementation is sketchy.  

My recommendation if you want to keep the default as autoconvert.  Please give the patch a full review as we need to get this right.  Then we should switch the behavior and run the test bench.  Then we need to add a few test cases that actually hit those other paths.  Last we figure out how to give appropriate warning this option is flipping in JPype 0.8 series.  If everything is working after that I will approve shipping it with the default as autoconvert.

Thrameos:
4 tests currently fail with legacy mode.  Most were incidental where the conversion was not the point of the test.  One test specifically was using JString, but that can be switched to another class.

```
___________________________________________________________ ArrayTestCase.testGetSubclass ___________________________________________________________

self = <test_array.ArrayTestCase testMethod=testGetSubclass>

    def testGetSubclass(self):
        t = JClass("jpype.array.TestArray")()
        v = t.getSubClassArray()
>       self.assertTrue(isinstance(v[0], jpype.JString))
E       AssertionError: False is not true

test/jpypetest/test_array.py:94: AssertionError
____________________________________________________ ExceptionTestCase.testExceptionByJavaClass _____________________________________________________

self = <test_exc.ExceptionTestCase testMethod=testExceptionByJavaClass>

    def testExceptionByJavaClass(self):
        try:
>           self.jpype.exc.ExceptionTest.throwRuntime()
E           jpype._jclass.java.lang.RuntimeException: Foo

test/jpypetest/test_exc.py:48: java.lang.RuntimeException

During handling of the above exception, another exception occurred:

self = <test_exc.ExceptionTestCase testMethod=testExceptionByJavaClass>

    def testExceptionByJavaClass(self):
        try:
            self.jpype.exc.ExceptionTest.throwRuntime()
            self.fail()
        except JException(java.lang.RuntimeException) as ex:
            self.assertIs(type(ex), java.lang.RuntimeException)
            self.assertEqual('Foo', ex.message())
            trace = ex.stacktrace()
>           self.assertTrue(trace.startsWith(
                'java.lang.RuntimeException: Foo'))
E           AttributeError: 'str' object has no attribute 'startsWith'

test/jpypetest/test_exc.py:54: AttributeError
_______________________________________________________ ExceptionTestCase.testExceptionThrown _______________________________________________________

self = <test_exc.ExceptionTestCase testMethod=testExceptionThrown>

    def testExceptionThrown(self):
        try:
>           self.jpype.exc.ExceptionTest.throwRuntime()
E           jpype._jclass.java.lang.RuntimeException: Foo

test/jpypetest/test_exc.py:37: java.lang.RuntimeException

During handling of the above exception, another exception occurred:

self = <test_exc.ExceptionTestCase testMethod=testExceptionThrown>

    def testExceptionThrown(self):
        try:
            self.jpype.exc.ExceptionTest.throwRuntime()
            self.fail()
        except JException as ex:
            self.assertIs(type(ex), java.lang.RuntimeException)
            self.assertEqual('Foo', ex.message())
            trace = ex.stacktrace()
>           self.assertTrue(trace.startsWith(
                'java.lang.RuntimeException: Foo'))
E           AttributeError: 'str' object has no attribute 'startsWith'

test/jpypetest/test_exc.py:43: AttributeError
```

Thrameos:
Next make sure at least something in the testbench has executed each of these paths at least once.

- [ ] Objects getField
- [ ] Objects invokeStatic
- [ ] Objects invoke
- [ ] Objects getArrayRange
- [ ] Objects getArrayItem
- [ ] Proxy calling arguments

Thrameos:
@marscher I added the message as requested, added tests of each path, and fixed the testbench.  That leaves one item on the check list for you.

Unfortunately, I was not able to get FLASHING to work properly on the warning due to incompatibility between MSDOS and Apple 2e character encoding differences.  

This just leaves the elephant of why I switched to do not auto convert on the first place  (other than the performance differences and ideology).  4 tests in Python 2.7 cannot pass with convertStrings true.  Those are all the ones involving unicode.  Attempting to convert to string results in unicode in string which broke Python 2.7 where str and unicode are different concepts.  We can just disable these tests for series 0.7.  But the behavior will be busted.

Thrameos:
The only other option I have to try to fix the tests is to make JPPyString scan the string looking for any UTF extends codes.  If it finds them then it would produce a unicode rather than a string.  But that would mean that we would have a polymorphic return.

PR #470, opened 2019-06-07
Improvements to JProxy

Thrameos:
Fixes the problem reported by the user and give the test bench some TLC as this and a bunch of other things were being missed.  There were a lot of cases in this area where one form or another was not tested.  Fixing it took much longer than expected because all of the corner cases (must be list[str or JClass], or str, or JClass, but all must resolve to JInterface).  

Note.  This does make the change that both bare lists and defined lists are accepted in the old JProxy interface.
```
JProxy([intf1, intf2], dict={...})
JProxy(intf1, intf2, dict={...})
```

I wanted the new interface to be simple (though specifying a list explicitly is also accepted)
```
   Java:  
        public class Foo implements MyInterface1, MyInterface2
   Python (preferred): 
        @JImplements(MyInterface1, MyInterface2)  class Foo
   Python (acceptable): 
        @JImplements([MyInterface1, MyInterface2])  class Foo
```
I can kill the second form if it is too much.

This change may break the total nonsense form  ``JProxy(intf1, foo)`` where dict is being implied as based on position, which should never have worked.

A secondary change is it now reports an error on the old JProxy if either no interfaces are specified or the dict or inst is missing.  Letting it quietly pass by this sort of error just to blow up later offends me.

The error message also became more terse.  Giving the correct usage is the point of the documentation
and not the exception string.  I added checks to the test bench to make sure the error was reported for the right reason rather than any reason.  There needs to be an action item to do this universally, as people do get burned by deceptive messages. 

@marscher general question, do you prefer many small tests that test each aspect of a function individually and thus report the exact failure by the test that failed, or large tests that test multiple aspects of a function?  My personal preference is many small because otherwise I get burned when I fix something in a test and it just falls through the next part of the test.  But I will defer to you as we are already over 400 tests.

### DIFF
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -14,8 +14,12 @@
 #   limitations under the License.
 #
 # *****************************************************************************
-import collections
 import sys as _sys
+try:
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
+
 
 import _jpype
 from . import _jclass
@@ -245,7 +249,7 @@ def _JArrayNewClass(cls, ndims=1):
 # Cannot be Mutable because java arrays are fixed in length
 
 def _isIterable(obj):
-    if isinstance(obj, collections.Sequence):
+    if isinstance(obj, Sequence):
         return True
     if hasattr(obj, '__len__') and hasattr(obj, '__iter__'):
         return True
--- a/jpype/_jcollection.py
+++ b/jpype/_jcollection.py
@@ -14,7 +14,10 @@
 #   limitations under the License.
 #
 # *****************************************************************************
-import collections
+try:
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
 
 from . import _jclass
 from . import _jcustomizer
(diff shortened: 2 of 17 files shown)


marscher:
Regarding the test organization question, I'd also prefer small tests (actually as small as possible) to keep things tidied up.

Thrameos:
Due to compatibility issues with Python 2.7 JProxy only takes the form ``JProxy([intf1, intf2])`` when multiple interfaces are specified.  JImplements takes only the form ``@JImplements(intf1, intf2)``.  Most is bug fix with the exception of the error message change.

@marscher Post this pull request a memory leak was discovered proxy implementation, but it will require a significant pull request to correct it.  None of the JProxy code checks for a cyclic dependency in the instance.  As there is no GC methods on that class there is no way that it can ever clean up if the Proxy is referred to either in the inst, dict, or Implements forms.  I can try to clean up the problem as I worked through the issue in JPype 0.8 core.  But it will take me at least a day or two to finish it.

Thrameos:
@marscher Okay I fixed the memory issue and backported all the relevant bug fixes from the 0.8 branch.   This should fix #482 .  Assuming this compiles, this and the string fix are all that are needed to release 0.7.0.  You need to give me some direction on the other pull.

Thrameos:
Merging this so that the string pull can be made fresh.

PR #476, opened 2019-06-11
Core rewrite for 0.8 series

Thrameos:
I know that 0.7 is still waiting for formal release, but I am trying to make head way on the very long task list I have laid out for the 0.8.  

This pull request removes all global variables from within ``native/common`` and transfers the majority of the major functions to Java.  The change list thus far is included in the pull request.  There is no major functionality improvements.  The Java transfer is still in flux and there are major performance bottlenecks on startup of the JVM.


Thrameos:
Hmm.  That is painful.  Python 2.7 is failing for lack of support for mutex.  The reason this was added it that it appears at least superficially, that the tracer would causing some of the crashes when trying to debug.   I can remove it but then running the tracer will likely result in the same problem.  This is one of the reasons I abandoned C++ so long ago.  Anything involving cross platform eventually devolves to break out the stone knives and bear skins, and then working to build your own iron forge rather than concentrating on what you want to do.

marscher:
So the Py27 Mutex class is a dummy? I can remember that I used it.

Am 12. Juni 2019 16:56:13 MESZ schrieb Karl Nelson <notifications@github.com>:
>Hmm.  That is painful.  Python 2.7 is failing for lack of support for
>mutex.  The reason this was added it that it appears at least
>superficially, that the tracer would causing some of the crashes when
>trying to debug.   I can remove it but then running the tracer will
>likely result in the same problem.  This is one of the reasons I
>abandoned C++ so long ago.  Anything involving cross platform
>eventually devolves to break out the stone knives and bear skins, and
>then working to build your own iron forge rather than concentrating on
>what you want to do.

Thrameos:
No, it is just Windows 2.7 that is an issue.  Python2 series on Windows still uses Visual Studio 2008 which is back when Microsoft pretty much ignored every standard.  Yes they have mutex and threads, but not through any standard mechanism.  Thus if I want to write code that compiles in Python 2.7 under windows I have to figure out what worked in 2008 (and just finding the online docs from that period can be a problem).   I suppose I could make everything go through Python rather than C++ mechanisms, but that would mean placing stuff that potential could interrupt control flow on tracer and exceptions paths.  I really try to avoid doing that as those make for really challenging and pointless bugs in which I am burning a productive night or two tracking down something on optional code that most users will never even compile.

Thrameos:
Okay so the steps that need to make this workable.  In no particular order

- [ ]  Merge master into this so that we don't end up with miles of difference between the master and the devel branches.
- [ ]  Fix the performance issues that loading all classes referred to on start up
- [ ]  Try to load a second JVM at the same time and see what crashes and burns.
- [ ]  Start on adding the type conversion interface.

Thrameos:
Closing this for now.  Backporting to main branch.

PR #478, opened 2019-06-13
Adds __doc__ to Java methods

Thrameos:
Adds support for dynamically creating ``__doc__`` for Java methods on request.  We could do this for fields and classes as well.  Currently pretty minimal but we can get as fancy as we want.  The ``__doc__`` attribute just calls a method defined by setResource which must be a Python method accepting the method dispatch, the java.lang.Class holding the dispatch, and the list of overloaded methods (in the order that they are matched).

I haven't developed any tests for this one yet, so it is possible that the stability on it is less than stellar.  We may want to wait to apply this in 0.7.1.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -6,6 +6,7 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 - **Next version - unreleased**
 
 - **0.7.0 - 2019**
+  - Doc strings are generated for classes and methods.
 
   - Complete rewrite of the core module code to deal unattached threads,
     improved hardening, and member management.  Massive number of internal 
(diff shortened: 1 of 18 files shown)


Thrameos:
Turns out there is no way to put docstrings on attributes because you can't prevent dereferencing.  Thus I just put the docs in the class docstring.

Thrameos:
Example method docstring
```
Java method dispatch 'substring' for 'java.lang.String'

Overloads:
```

Example class docstring
```
Java class 'java.lang.String'

Derived from ``java.lang.Object``
Interfaces:
  ``java.io.Serializable``
  ``java.lang.Comparable``
  ``java.lang.CharSequence``

Constructors:
  String(byte[])
  String(byte[], int, int)
  String(byte[], java.nio.charset.Charset)
  String(byte[], java.lang.String)
  String(byte[], int, int, java.nio.charset.Charset)
  String(java.lang.StringBuilder)
  String(java.lang.StringBuffer)
  String(char[], int, int)
  String(char[])
  String(java.lang.String)
  String()
  String(byte[], int, int, java.lang.String)
  String(byte[], int)
  String(byte[], int, int, int)
  String(int[], int, int)

Raises:
  ``java.io.UnsupportedEncodingException``

Attributes:
  CASE_INSENSITIVE_ORDER (java.util.Comparator): final static field
```

Thrameos:
I was able to get sphinx and autodoc to actually import a Java class and document it.  Though I don't think we will use currently, it was good to verify that the docstrings were valid enough to be usable.  I left most of the hooks for doing so in place.  

If we need to exercise this option

- sphinx config needs to start the JVM
- the mock section of sphinx config needs to be removed as it interferes with operation.
- the classes need to place themselves in an accessible module path. 

The last one was a bit of a puzzle.  If I don't do that all I get is "alias to"  in the docs.  I poked around and others have similar problems, but I don't see a solution.

It did remind me a bit about the challenge of how to get pickle to play with java.  There it had the same requirement that all classes have a path to fetch the static methods from.  I am thinking it may be necessary long term to create module like class called "jpype.classes" which will pretend to house all of dynamic methods.  However, we are going to end up with some issues if I do get multiple JVMs with the `JPackage`  class.  If we want things to actually clean up between execution the caching mechanism in the JPackage jpype.java is very likely to interfere.  I know it will never be possible to get the ``jpype.imports`` to play nice with a second JVM, but JPackage may need to become a front end that houses resources in the JVM context.  I am also thinking that there may need to be a general set of shutdown handlers are placed in the JVM to finalize the shutdown.  This would allow the shutdown routine to got through and render the ``jpype.imports`` and JPackage objects into a Java free state (cleaning up resources and preping them for reuse if needed)  But this is all going to have to wait for the  0.8 series, as I can't have any globals in ``_jpype`` and ``jpype`` for this to actually work.

zhouzhen1:
@Thrameos there is a typo here at https://github.com/jpype-project/jpype/pull/478/files#diff-1665d860263e43d71785bf8555bace0fR499 "fieldnfo" should be "fieldInfo".

marscher:
very good catch! It seems enums are not tested for documentation

Thrameos:
@marscher Seems like this one needs some additional work regarding Python 2.7 that I will need to get to later.  Guess it will miss the cut for JPype 0.7.0 unless I get to it tonight before you have time to ship.

Thrameos:
@marscher Okay I took a few minutes between meetings to polish off the final issues in this PR.  It should be ready for review and inclusion if you haven't already started the release process.

Thrameos:
Thrameos pokes @marscher for signs of life.

marscher:
@Thrameos sorry for the delay. Which JDK should I use to build the jar?

Thrameos:
@marscher that should wrap up everything that was anticipated for the 0.7.0 release.  Pickle, annotations, and some other items are on track for completion in a month or two once I have a bit more time applying testing.   When will you have time to tag and release this version?  I am not sure how much stuff I have busted in the release system.

If the build pattern fix when in it will build the jar binary compatible to 1.7 which is the oldest we support.  Let me check...

```
roject default="all" name="JPype - Native">

  <property name="JAVA_VERSION" value="${ant.java.version}"/>
  <!-- src can't be "java" as it breaks nose tests -->
  <property name="src" location="java"/>
  <property name="build" location="build"/>

  <target name="test" depends="compile">
```

Nope still has it set to latest which is wrong.

Thrameos:
One minute... doing a quick pull request to fix that.

marscher:
Thank you! I will tag this immediately after your PR.

PR #481, opened 2019-06-16
Support for Pickle

Thrameos:
This completes the work started long ago to support saving and loading of Java objects as part of a pickle file.  

The implementation on this was actually simpler that I thought thus I was able to get this to work in the 0.7 series where I had just enough thunk support for the custom classes.  I needed to hack the dispatch table so that all Java classes could be routed to a serializer.  A custom Java class for encoding and decoding is then stored in our private jar which does the work.  Then loading required creating an extra resource.  Attempting to load a stored Java pickle file with the regular unpickler, gives an error.  I added documentation and some trivial test cases. 

I feel positively dill about this one.  But lets hold it until the next release.


Thrameos:
This one needs two changes before it will be usable.
1) We need an entry in changelog.
2) We need a method to convert a byte[] to bytes that does not involve going through numpy.

Thrameos:
I completed the required tasks to get this to work without numpy and on Python 2.7.

Python 2.7 and Python 3.x are very different in the Pickler.  I had to hack into an entirely different table to do the work because there is no concept of a private extension dispatch_table in Python2.  Unfortunately, Python3 removed the dispatch table entirely as it is all internal to the cPickler implementation.  And the two tables require implementing different methods.

I think this one is ready for review and inclusion in 0.7.1

marscher:
very well!

PR #483, opened 2019-06-17
Enable option to support legacy conversion of strings with polymorphic return.

Thrameos:
This is a variant of #469 that automatically selects between str and unicode on the return.  I am not sure how much of a change this is from 0.6.3.  I don't recall if it autoselected before.  Please select either this or #469 with test disabled for the release.


Thrameos:
@marscher  Just one wafer thin mint more to go and we are done with 0.7.0.  Just .... one more.

marscher:
Since this is only an intermediate solution, I'd prefer it over #469, because it does minimal harm to our user base. I have added two more jobs on Travis to test the conversion=False branch.
Please merge on your behalf and thanks for implementing it.

Thrameos:
And thank you for all your help and putting up with my pushy demands.   I know I have been pushing hard for the last few weeks to get this to a "production" state.

marscher:
Never mind. I really appreciate your all your effort for this project!

Thrameos:
Unfortunately, I believe all remaining actions for producing 0.7.0 are now on you.   Hopefully I haven't broken anything in the release system during my wandering through the china shop.

marscher:
Don't you want to include the doc PR as well?

Thrameos:
Up to you on that PR.   I consider it a nice to have rather than urgent which all the rest of the 0.7 stuff was.  I marked it as 0.7.1, but if you feel that it is sufficiently ready we can add a note to the ChangeLog.

PR #484, opened 2019-06-18
Annotations

Thrameos:
This is a test pull for annotations.  Thus far I haven't seen any change in behavior.  Please do not merge this one until we confirm function.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,7 +5,17 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 - **Next version - unreleased**
 
+  - PyJPMethod now supports the FunctionType API.
+
+  - Tab completion with Jedi is supported.  Jedi is the engine behind
+    tab completion in many popular editors and shells such as IPython.
+    Jedi version 0.14.1 is required for tab completion as earlier versions 
+    did not support annotations on compiled classes.  Tab completion
+    with older versions requires use of the IPython greedy method.
+
+
 - **0.7.0 - 2019**
+
   - Doc strings are generated for classes and methods.
 
   - Complete rewrite of the core module code to deal unattached threads,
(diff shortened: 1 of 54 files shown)


Thrameos:
This one is getting closer to ready.  Although it is not possible to get tab completion to work as was request, I did manage to get 95% of the Python function API working.  I also found a difficult to hit segmentation fault.  So despite being a failure I was able to get something good out of it.

Thrameos:
The kind author of Jedi did an update that should correct our issue with signatures for Python3.   Hopefully in combination with this pull request we wall actually have working tab completion.  This latest commit converts our PyJPMethod to be a FunctionType for the purposes of Python2 and Python3.  Though this is good because we now don't require special treatment in places that are probing the object, it also may have some unexpected side effects.  I am debating if I should rename the object type for PyJPField and PyJPMethod to JField and JMethod, and create an alias in ``jpype``.  Unlike the rest of ``_jpype`` types these are directly exposed to the users.

Thrameos:
Some jedi patterns are completing properly now with unreleased version 0.14.1 but others are still failing.  Thus I put another ticket in with jedi.

Thrameos:
Well it was a very long road but we now support docstrings and annotations.  This requires jedi version 0.14.1 which is unreleased but on my modified copy I get successful completions of classes, methods and fields. 

@marscher This is now ready for review for version 0.7.1.

marscher:
Shall we include this for 0.7 or really wait for a patch release?

Thrameos:
Lets keep this one for now.  I think that it is all good, but I have not given it a good beating.  The FunctionType has a flag for VectorOperations that I do not yet understand the implications of.  Thus I suspect that there is a path that will cause problems.  I just don't know how to hit it.

Thrameos:
@marscher Any chance for this to get reviewed or should I self review to get us up to a release candidate?

PR #490, opened 2019-06-24
Fix docs for RTD with mock

Thrameos:
Okay I am a complete and utter failure with getting mock to do anything for me.  So I cheated.  Call me a failure if you will.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -53,34 +53,34 @@ copyright = u'2014-18, Steve Menard, Luis Nell and others'
 # built documents.
 #
 # The short X.Y version.
+import mock
+class TypeMock(type):
+    def __new__(cls, name, bases=None, methods={}):
+        if not bases:
+            bases = tuple()
+        methods['__init__'] = TypeMock._init
+        methods['__getattr__'] = TypeMock.__getattr__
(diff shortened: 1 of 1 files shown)


Thrameos:
@marscher This was ten times harder than it should have been.  Mock does not have the concept of actually mocking a type, thus one has to bend over backwards to wrap it and still be a type.  Forum postings offer solutions from "don't do that, use duck typing", to "just override the subclass api to fool it".   It wouldn't be that hard for mock to use an actual type wrapper that is a proper type.  So I just meta programmed it and made a simple TypeMock class that does what is desired which was a bit overkill.  We just need `_jpype.isStarted()` to return false, and the basic `_jpype` classes to be types and all is good.

Thrameos:
Unfortunately, my hack didn't work completely as Python2 system doesn't like the ``__init__`` hack.   
Just need a bit more reformulation to before it is completely safe.

marscher:
On 25.06.19 00:24, Karl Nelson wrote:
> Unfortunately, my hack didn't work completely as Python2 system doesn't like the ``__init__`` hack.
> Just need a bit more reformulation to before it is completely safe.
> 
I think we can force RTD to use Python3 to build, so we can postpone 
refining the hack.

PR #493, opened 2019-06-25
Define behavior for casting JProxy with JObject

### LINKED ISSUE #492: Proxies wont wrap with JObject
A defect in the search pattern of JObject does not recognize that Proxies are already able to become Java Objects.   A relatively small change should fix this defect.

Thrameos:
Fixes #492


Thrameos:
It would be better if JObject cast to the correct interface type when there is only one implemented interface, but for now this will have to do as we don't have easy access to the interface list.

Issue #486, opened 2019-06-24
[Question] Define custom Java class in Python


iTakeshi:
(I'm sorry if any existing docs/issues have solved this question)

I'd like to define a Java class in python to make it a type argument of a generic constructor.
Consider we have a generic java class like
```java
public class GenericClass<T> { ... }
```
and I want my custom class defined in python be the type argument `T`:
```python
class MyClass(object):
    pass

ConcreteClass = jpype.JClass("GenericClass<MyClass>")
```
The python code above is terminated with a segmentation fault because the python class `MyClass` cannot be seen in JVM.

Is there any way to get around this problem?

Thrameos:
Due to type erasure in Java there is no difference between `GenericClass<MyClass>`  and `GenericClass`.  Thus the line `ConcreteClass = jpype.JClass("GenericClass")` should do.   There are other issues with your example.  Importing of classes without a package is a problem in Java .  Thus you may have issues with getting Generic class.   

If you want the make a Python class visible in Java it must implement some interface.  Any interface including trivial ones like `java.io.Serializable` will do.  Thus if your container takes some Java interface type and you then create a proxy class for it in Python it will store those items.  

However, looking at JPype code, I am not sure exactly what happens when a Proxy class is returned from Java.  It doesn't appear that the returned object is the original proxy class.  Thus I am not sure that anyone has ever requested that proxy classes are able to return back to the original Python object on return.  Seems like an oversight.

Here is an example in which a custom Python class was placed in a generic container.  Unfortunately, the returned item is a Java object (which holds reference to Python object) and not the original Python instance.

```
import jpype
from jpype import JImplements, java, JClass, JObject, JOverride

jpype.startJVM()

@JImplements(java.io.Serializable)
class MyClass(object):
    def __init__(self, mesg):
        self.mesg=mesg
    def getMessage(self):
        return self.mesg
    @JOverride
    def toString(self):
        return self.mesg

jl=JClass('java.util.ArrayList')()
jl.add(MyClass("foo"))

print(jl.get(0))
print(type(jl.get(0)))
```

I can toss this onto the requested feature list.   I am not sure yet how difficult it would be to recognize a Python proxy being returned from Java and retrieving the original.  It is pretty trivial then it can appear in the next release 0.7.0.  It if requires a structural change then it would have to go toward the future version 0.8.

[...]

PR #494, opened 2019-06-25
Pass a python object into Java and return unharmed.

Thrameos:
This is a follow up to #486.  The user was looking to store a Python object in a Java container.  This is the first half of a patch to allow that to happen.  In this PR, we delegate the lookup of  the class for an object to Java where it splits out common types from special cases such as lambda, anonymous classes, and proxies.  For proxies that belong to Python we cam the return the item. 

This is currently happening automatically.  The only potential breakage is a proxy that receives another proxy on the parameters will get the Python object rather than the Java mostly useless version. 

A  follow up should be able allow manual casting of any arbitrary Python class to be cast to Java and live  there, without the need to force implementation of an interface.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,12 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 - **Next version - unreleased**
 
+  - JProxy objects now are returned from Java as the Python objects 
+    that originate from. Older style proxy classes return the 
+    inst or dict. New style return the proxy class instance.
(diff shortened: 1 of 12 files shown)


marscher:
thats pretty cool!

peschue:
Thanks a lot for making these fixes! I am building a Java API for a Python tool and need to recognize the Python objects that I feed into Java and use their non-Java properties.

So I just moved from jpype1 in conda to self-built jpype and noticed that this PR indeed fixes all the trouble I ran into today. :-)

PR #502, opened 2019-07-01
Classhints

Thrameos:
This is work for the prototype version of classhints.   It is still a long way from working.

Several other PRs will need to go in before this one is ready for review.


Thrameos:
This pull request is currently up to date with the changes to master.  Caller sensitive, annotations and the like are completely ported.  The type resolution system is reworked, but no new functionality is complete yet.  Preliminary changes to support sharing primitives across multiple JVMs are now complete.   Widening conversions were added to the basic types and type resolution between primitives is based on a type code, so that we can support multiple copies of classes (differing by JVM) at the same time. 

Next will be working on the performance issues with startup.  Currently about 9000 entities are created due to method requested classes.  That is to load a class, it pulls in all the methods, which request more classes, which have more methods which request more classes and so on.  Truncating this process by not loading the classes used by methods reduces this burden to about 1000 entities on startup.  This will be needed before the startup penalty is similar to the 0.7 series.  Once this is completed I can resume work on type hints.  The current starting penalty is too large to complete debugging activities.

Thrameos:
Fixes #515

Thrameos:
@marscher I am still struggling with this one.  I am trying to correct the fundamental restriction that the JVM is a monolith that can only be started once.  I have introduced the concept of a JContext which represents a specific instance of the JVM.  With this, it is possible to start and stop the JVM or create more than one JVM at a time.  

The key issue is that the current API requires all the types live at the top level.  I have created a shadow copy of the default context at the top level so that we do not break the current API, but there are still lots of weird edge cases that happen if someone uses both the default and the client jvms.  For example, each JVM has its own copy of each primitive type (JInt, JFloat).  This breaks stuff like isinstance.  Weirdness happens if you do something like "jvm1.JArray(jvm2.JInt)"  (which JVM does the array type live in?)  Also what happens if the default jvm is used with a client type "jpype.JArray(jvm2.JInt)"?  I am going to try to resolve the issues enough to have a workable version (resolved in some way, not necessarily the best) and then put together some examples.

I can probably get primitives and Strings to work cross JVM, but the rest will need to be pushed from one JVM to another explicitly.   The oddity would happen if the default JVM is not running because the client JVM would not be able to create the primitives except for explicitly.

Some portions of the API just will never work outside the default JVM such as jpype.imports.  But we will always be able to do ``jvm.JClass("MyClass")``.  

Still a lot of issues.

Thrameos:
@marscher You may want to look this one over a bit now.  This is the mostly completed core for the 0.8 series.  I finally completed the speed optimization, but unfortunately that requires modifications to the class structure which is pretty much not portable to older versions.  We may be able to get one or two of the older ones going, but Python series 2 is pretty much out of the question.  275% of the original speed is certainly better.

I will next get back to working on the classhints which should allow for user installed type conversions and thus mostly eliminate the need for the collection customizers.

One version in this patch series was compatible with starting more than one JVM at a time, but the JVM support was very poor.  Out of the box you can't start more than one JVM nor get it to restart.  I was able to do some tricks on Linux to load multiple copies of the JVM dll, but it was very specific to the OS call hooks and I was not able to resolve performance and memory issues, so I dropped it for now.

Thrameos:
This one appears to have two problems that need to be resolved.  

There is a leak in resources associated with strings.  Likely the dictionary is not being traversed.  Should be correctable.

Second older pythons do not support paths or other recent peps.  Thus the classhints are not working as expected.  As classhints are for new pythons anyway I will just disable those tests for everything before 3.6.

Thrameos:
Think I am getting closer.  It appears for whatever reason that the dictionary in PyJPValue is not getting visited thus leaving dangling objects.  But they don't appear the live object references so they are currently invisible.   Hopefully now that I can see them them I can fix it.

Thrameos:
So there appear to be three separate and distinct leak paths.

1) creating a JString produces a leak and the leak size is proportional to the size of the string.
2) But it I stop UTF8 translation then it still leaks but it is no longer proportional.  
3) The java.lang.Integer is never calling the destructor.  

Seems like I will have to isolate each of these and add some more tests to the test bench to keep it from coming back.  A few of these may be in 0.7 as well and just are escaping the leak detector.

Thrameos:
Hmm.  I have tried and can't replicate the errors in 3.4 nor 3.5.  Something is different on those installs of Python than is in a clean Trusty install.   Perhaps we should drop 3.4 from the matrix as it is currently out of service life.

https://devguide.python.org/#status-of-python-branches

Thrameos:
Okay one leak remains on this one, but it is a puzzler.  Boxed types need special wrappers, but those wrappers leak somewhere in the python type system. I will take another shot at it next week.

Thrameos:
Closing this for now.  Backporting to be closer to main branch.

PR #523, opened 2019-08-06
Improved exception handling

KOLANICH:
Current approach to stack traces from Java has following drawbacks:

* Messagges produced by Java had limited depths of stack traces. I mean that stack traces are truncated and removed parts are replaced with messages about that.
* Format of stack traces are not pythonic. I mean not compatible to the software in python ecosystem.

I have created the code creating a stacktrace obj native for python from a JVM exception. I guess customized exceptions formatters like the ones used in IPython should understand it too, but IDK how to pass it to them because they construct them from the exceptions themselves using funcs in python stdlib. So a better solution may be to
1 raise not the exception produced the error, but a generic exception pointing to the piece of python code, but having the type inherited from the real java exception object and the props redirecting to it
2 set its `__cause__` to a real exception object. Make its stack a JVM stack.
3 make for real JVM exceptions `__cause__`  a property using `getCause`.

### DIFF
--- a/jpype/_jexception.py
+++ b/jpype/_jexception.py
@@ -26,6 +26,40 @@ else:
     _unicode = unicode
 
 
+from traceback import FrameSummary, StackSummary, TracebackException, _some_str
+
+def stack_trace_to_StackSummary(stack_trace):
+    res = []
+    for el in stack_trace:
+        ln = int(el.getLineNumber())
+        fn = el.getFileName()
+        if fn:
+            fn = str(fn)
+        else:
+            fn = None
+        res.append(FrameSummary(fn, ln, ".".join((str(el.getClassName()), str(el.getMethodName()))), lookup_line=False, locals=None, line=None))
+    res = StackSummary.from_list(res)
+    return res
+
+class JPypeTracebackException(TracebackException):
+    def __init__(self, exc_value):
+        exc_type = type(exc_value)
+        self.stack = stack_trace_to_StackSummary(exc_value.getStackTrace())
+        self.exc_type = exc_type
+        self._str = _some_str(exc_value)
+        cause = exc_value.getCause()
+        if cause:
+            self.__cause__ = self.__class__(cause)
+        else:
+            self.__cause__ = None
+        self.__context__ = None
+        self.exc_traceback = True
+    
+    @classmethod
+    def from_exception(cls, exc, *args, **kwargs):
+        return cls(exc, *args, **kwargs)
+
+
 class _JException(object):
     """ Base class for all ``java.lang.Throwable`` objects.
 
@@ -70,6 +104,14 @@ class _JException(object):
 
     # Included for compatibility with JPype 0.6.3
     def stacktrace(self):
+        """ Get a string listing the stack frame.
+
+        Returns:
+          A string from lines of ``traceback.TracebackException.format()`` result.
+        """
+        return "".join(l for l in JPypeTracebackException(self).format())
+
+    def java_stacktrace(self):
         """ Get a string listing the stack frame.
 
         Returns:


Thrameos:
I am not sure what this is trying to solve.  I am also not sure why you want stacktrace to be different than it is now.  Can you give me an example of the behavior before and after being fixed?

Thrameos:
I looked over this further and I think that it has some problems

- It completely changes the meaning of stacktrace by trying to make in have Python like output.
- It doesn't follow the naming conventions elsewhere in the package.
- It is going in the wrong direction because it is encouraging the use of the method I have been working to deprecate which is stacktrace()

My goal is to try to make things more pythonic by removing functions that are redundant between the languages.  Thus a better direction would be to make Java exceptions have a ``__cause__`` which is an object which converts the Java stacktrace into a python on.  I have tried this a few times with only moderate success.  The problem being that Python traceback objects are completely opaque and don't use macros to access their methods.  Thus

KOLANICH:
>The problem being that Python traceback objects are completely opaque and don't use macros to access their methods.

Or we probably should try to mess with that directly bypassing the macros if needed.

Thrameos:
Yeah, I think that added a support call to the C interface which takes the current Java stacktrace unpacks it into objects and then directly creates the Python stack trace would be the best course of action.   I was able to generate the correct traceback frames using Python code like

```
import types
import sys

glob = {'TypeError':TypeError}
code = compile("raise TypeError('no')", "foob.py", "exec")
code = type(code)(code.co_argcount, code.co_kwonlyargcount, code.co_nlocals, code.co_stacksize, code.co_flags, code.co_code, code.co_consts, code.co_names,
      code.co_names, code.co_filename, "java.lang.String.substr", 1000, code.co_lnotab)
method=types.FunctionType(code, glob, "bar")
glob['s1'] = method

code = compile("s1()", "foob.py", "exec")
code = type(code)(code.co_argcount, code.co_kwonlyargcount, code.co_nlocals, code.co_stacksize, code.co_flags, code.co_code, code.co_consts, code.co_names,
      code.co_names, code.co_filename, "com.Broken.method", 500, code.co_lnotab)
method=types.FunctionType(code, glob, "bar")
glob['s2'] = method

try:
    method()
except Exception as ex:
    ex1= ex

try:
    1/0
except Exception as ex:
    ex.__cause__= ex1
    raise
```

which gave output like
```
  File "testtb2.py", line 18, in <module>
    method()
  File "foob.py", line 500, in com.Broken.method
  File "foob.py", line 1000, in java.lang.String.substr
TypeError: no

  File "testtb2.py", line 23, in <module>
    1/0
ZeroDivisionError: division by zero
```

The key problem with this native method is the Python native overhead is just way too high for what is just setting up some objects in the python with specified names.  Because there are no internal macros which look at Python implementation before directly accessing, we will have to create the fake frame even if the frame is never printed which is unfortunate.  I am guessing they made a performance choice in the implementation that is preventing direct intervention.  I think we can just assemble the object in C and return the backtrace as a cause attached to the Java exception.  This will be much more efficient than the method in https://unterwaditzer.net/2018/python-custom-tracebacks.html.

This should also make output objects that Python frontends can deal natively with so we won't need to give any special instructions on how to work with the exception classes.

Hopefully everything we need is exposed.  It appears that because of legacy concerns in Python when they did not support circular referencing, the traceback frames and exception code pieces do not derive from ordinary objects so we likely just have to use them rather than extend them.  But if we keep the overhead low I don't see why we can't support this feature.  We will need to make sure we don't create leaks because we will need to mess with objects with very specific referencing requirements.  

I will also see if I can fool the internal native code to take a derived exception class for the cause which I can lazy evaluate stackframe from.  I know that I can't do it from within Python but perhaps the extension module can override the lookup.  For some reason they have it set as a property by overriding the property in the derived class does no produce the correct output.  Thus something is grabbing access to the member without hitting the virtual table.  If we work around this then we would not have to pay any performance cost beyond creating a cause object.  If this proves impossible, we may need to try to propose a patch to CPython to help with this. 

Does this sound like a plan?

Thrameos:
Okay I am going to take a shot at this one.

Thrameos:
I made some progress on this one.  The key issue is that Python doesn't allow any type in the traceback system which isn't an actual traceback which must point to an actual code block and so forth.  I can fake it well enough to fool the traceback module, but not the actual print routine that handles errors.  It appears there isn't even dictionary lookups or anything that I can hook on.  Thus I would need to convert the entire C++ and Java stack objects into actual traceback objects.

The API issues on this is pretty nasty.  The code and traceback objects are variable and there in no external API to create them so I would have to replicate Python code unless I can find a way to hack it.  The entire thing would have to be coded in CPython extension because there is no programmatic way to do it from within Python.

KOLANICH:
Shouldn't cpython be modified to have the needed signatures in its headers and the needed symbols exported?

Thrameos:
Likely.  Though the simplest mod is likely to allow a frame like object rather than requiring a frame.  For example the offending routine (best guess) is print_exception in pythonrun.c.  It is calling PyFrame_Check indirectly which requires exact type match which is fine because there is no basetype flag on frames.  There is also no zeros checks on many fields so even to add a frame means filling out the whole structure. For now I am just going to try to see if I can satisfy it, then if I can identify a simple mod can submit it.  Work first the simplify later. 

The big fear here is because this API is without a PEP anything I cook up will get changed on me in the next version.  But at least without having to duel version this to 2.7 it may be possible.  I have constructed two versions mocking this up that work geeat for Python traceback and other utilities.  But the implementation  that is hit for general purpose is all in C.


I should note that Java has a very similar problem.  Thus far I haven't found any way to insert fake frames.  The complexity there is the exception frames are tied to the frame lookup which is part of the security system.  That is why it was so hard to make caller sensitive methods work properly.

My long term plan for Java is create a PythonException that extends RuntimeException.  You cant really do much with it in Java but I can use the round trip logic to pass back the original exception to Python. That way we dont lose the original Python exception.

Thrameos:
So the code that we are trying to access is 

```
static int
tb_printinternal(PyTracebackObject *tb, PyObject *f, long limit)
{
    int err = 0;
    Py_ssize_t depth = 0;
    PyObject *last_file = NULL;
    int last_line = -1;
    PyObject *last_name = NULL;
    long cnt = 0;
    PyTracebackObject *tb1 = tb;
    while (tb1 != NULL) {
        depth++;
        tb1 = tb1->tb_next;
    }
    while (tb != NULL && depth > limit) {
        depth--;
        tb = tb->tb_next;
    }
    while (tb != NULL && err == 0) {
        if (last_file == NULL ||
            tb->tb_frame->f_code->co_filename != last_file ||
            last_line == -1 || tb->tb_lineno != last_line ||
            last_name == NULL || tb->tb_frame->f_code->co_name != last_name) {
            if (cnt > TB_RECURSIVE_CUTOFF) {
                err = tb_print_line_repeated(f, cnt);
            }
            last_file = tb->tb_frame->f_code->co_filename;
            last_line = tb->tb_lineno;
            last_name = tb->tb_frame->f_code->co_name;
            cnt = 0;
        }
        cnt++;
        if (err == 0 && cnt <= TB_RECURSIVE_CUTOFF) {
            err = tb_displayline(f,
                                 tb->tb_frame->f_code->co_filename,
                                 tb->tb_lineno,
                                 tb->tb_frame->f_code->co_name);
            if (err == 0) {
                err = PyErr_CheckSignals();
            }
        }
        tb = tb->tb_next;
    }
    if (err == 0 && cnt > TB_RECURSIVE_CUTOFF) {
        err = tb_print_line_repeated(f, cnt);
    }
    return err;
}
```

Thus each stackframe requires setting up 3 objects.  We need the traceback, the frame, and the code.  The traceback can be created and has an `__init__` method so it is possible to create.
The frame  is a private object that can't be created within Python.  So that is what I will need to try to replicate manually.  Code appears to have an `__init__` and perhaps can be created.

I love the help string on the Code object
```
Create a code object.  Not for the faint at heart.
```

Thrameos:
Okay got it... now I just need to get all the plumbing hooked up.

```
PyTracebackObject *tb_create(
		PyTracebackObject *last_traceback,
		PyObject *dict,
		const char* filename,
		const char* funcname,
		int linenum)
{
	// Create a code for this frame.
	PyCodeObject *code = PyCode_NewEmpty(filename, funcname, linenum);
	// Create a frame for the traceback.
	PyFrameObject *frame = (PyFrameObject*) PyFrame_Type.tp_alloc(&PyFrame_Type, 0);
	frame->f_back = NULL;
	if (last_traceback != NULL)
		frame->f_back = last_traceback->tb_frame;
	frame->f_builtins = dict;
	Py_INCREF(frame->f_builtins);
	frame->f_code = (PyCodeObject*) code;
	frame->f_executing = 0;
	frame->f_gen = NULL;
	frame->f_globals = dict;
	Py_INCREF(frame->f_globals);
	frame->f_iblock = 0;
	frame->f_lasti = 0;
	frame->f_lineno = 0;
	frame->f_locals = NULL;
	frame->f_localsplus[0] = 0;
	frame->f_stacktop = NULL;
	frame->f_trace = NULL;
	frame->f_trace_lines = 0;
	frame->f_trace_opcodes = 0;
	frame->f_valuestack = 0;

	// Create a traceback
	PyTracebackObject *traceback = (PyTracebackObject*)
			PyTraceBack_Type.tp_alloc(&PyTraceBack_Type, 0);
	traceback->tb_frame = frame;
	traceback->tb_lasti = frame->f_lasti;
	traceback->tb_lineno = linenum;
	traceback->tb_next = last_traceback;
	return traceback;
}

PyObject *PyJPModule_createTrace(PyObject* module, PyObject* sequence)
{
	if (!PySequence_Check(sequence))
	{
		PyErr_SetString(PyExc_TypeError, "createTrace requires a sequence");
		return NULL;
	}

	// Check all of the types first.
	Py_ssize_t sz = PySequence_Size(sequence);
	for (Py_ssize_t i = 0; i < sz; ++i)
	{
		PyObject *item = PySequence_GetItem(sequence, i);
		if (!PyTuple_Check(item))
		{
			PyErr_SetString(PyExc_TypeError, "createTrace entries must be tuples");
			return NULL;
		}
		if (PyTuple_Size(item) != 3)
		{
			PyErr_SetString(PyExc_ValueError, "tuples must have 3 items");
			return NULL;
		}
	}

	// Construct the traceback stack
	PyObject *dict = PyModule_GetDict(module);
	PyFrameObject *last_frame = NULL;
	PyTracebackObject *last_traceback = NULL;
	char* filename = NULL;
	char* funcname = NULL;
	int linenum = 0;
	for (Py_ssize_t i = 0; i < sz; ++i)
	{
		PyObject *item = PySequence_GetItem(sequence, i);
		if (PyArg_ParseTuple(item, "ssi", &filename, &funcname, &linenum) == -1)
			break;
		last_traceback = tb_create(last_traceback, dict,  filename, funcname, linenum);
	}
	if (last_traceback == NULL)
		Py_RETURN_NONE;
	return (PyObject*) last_traceback;
}
```

Thrameos:
Here is what the final product looks like.   It will generate similar for C++ errors.

```
  File "ExceptionTest.java", line 57, in jpype.exc.ExceptionTest.throwChain
  File "ExceptionTest.java", line 62, in jpype.exc.ExceptionTest.method1
  File "ExceptionTest.java", line 67, in jpype.exc.ExceptionTest.method2
Exception: Java Exception

  File "test.py", line 16, in <module>
    ExceptionTest.throwChain()
jpype._jclass.RuntimeException: java.lang.RuntimeException: Inner
```

Does that look any better?

KOLANICH:
Nice. Thank you. I have one more idea how to improve it, but it is a separate standalone feature (use a java decompiler written in java).

Thrameos:
I got the C++ one working now.

```
  File "native\python\pyjp_module.cpp", line 193, in PyJPModule_startup
    PY_STANDARD_CATCH;
  File "native\common\jp_env.cpp", line 179, in JPEnv::loadJVM
    JP_TRACE_OUT;
  File "C:\cygwin64\home\nelson85\devel\open\jpype\native\common\include\jp_platform_win32.h", line 57, in Win32PlatformAdapter::loadLibrary
    JP_RAISE_OS_ERROR_WINDOWS( GetLastError(), path);
Exception: C++ Exception

  File "C:\cygwin64\home\nelson85\devel\open\jpype\jpype\_core.py", line 227, in startJVM
    _jpype.startup(jvmpath, tuple(args), ignoreUnrecognized, convertStrings)
OSError: [WinError 193] JVM DLL not found: C:\Program Files\Java\jdk-12.0.1\bin\server\jvm.dll
```

If you have another idea please post it.  If it is something that that is optional and requires a hook I will see what I can do to add it.  I made some advancements with regard to adding hooks in the classhints branch.

PR #530, opened 2019-10-15
[setup] added option for coverage measuring

### LINKED ISSUE #511: Setup code coverage
Looking at other repos I see a number of them run code coverage tools.  It may be worth doing the same for that project.  But we are a very mixed language package, so it is less than clear how well these tools will work.  I poked a bit at Codecov, seems like it is simple and free to add at Github market place.

@marscher any opinion on whether Codecov is worth trying here?  Is there an alternative we should consider?

marscher:
* use it on travis ci and upload results to codecov.io
* fixes #511


marscher:
@Thrameos To get the coverage it seems mandatory to have the build inside the source directory. How do I convince the setup to place the jar in the correct spot?
The result would look like this: https://codecov.io/gh/jpype-project/jpype/pull/530/tree?path=native%2Fcommon

Thrameos:
By build do you mean everything include the CPython module or just the jar portion?  I have just been using hooks inside setupext directory to move files around after it was created.   If I have a tree layout where things need to be and which command they need to appear there I think I can make it happen.

For example the bdist command copies the built pieces back into the source directory (native) at a specific step.

Thrameos:
I am going to take a shot at this.  Do you know how the structure should look or doc that have that info?

setupext/dist.py
```diff
@@ -9,12 +9,15 @@ class Distribution(_Distribution):
         ('enable-build-jar', None, 'Build the java jar portion'),
         ('enable-tracing', None, 'Set for tracing for debugging'),
         ('ant=', None, 'Set the ant executable (default ant)', 1),
-        ('disable-numpy', None, 'Do not compile with numpy extenstions')
+        ('disable-numpy', None, 'Do not compile with numpy extenstions'),
+        ('enable-coverage', None, 'Instrument c++ code for code coverage measuring'),
```
Thrameos: So when I get enable-coverage I need to copy the jar to a new location?

Thrameos:
I added the copy code command over but we likely need to also add some code to the pytest command so that it will use the jar rather than the thunk code.  You can move it to a different location if you would like.  The pattern is pretty easy.  In the run command of each section use self.distribution to get to the command line options of setup.  So you can push files around for testing or coverage tools as needed.

With regard to thunks, right now it always pulls the thunk code first rather than asking for the external jar.  To use the external jar we must add the external jar to the class path explicitly so that it is found before the bootstrap loader.  I am planning on making an attempt to get the jar file to install with the shared library so that we don't need the thunk code at all.  I am vacation for the next two weeks so I may have a few hours to try that.

Thrameos:
I managed to get support for jacoco to work locally, but we need to figure out how to get jacocoagent and jacococli installed to use it in practice.  Then we would have to upload coverage.xml, coverage_py.xml and coverage_java.xml (unless we can merge them).

It doesn't appear that the java source has to be in a particular location, we just need to run the jar with the javaagent and then post process it to xml.

Thrameos:
Not sure if this will do it.  I managed to get coverage from all three languages, but I can't really see how to get them all into the coverage report so I am trying to post each of the pieces and see if the web site can handle it.  Also for the record I grossly dislike maven but as far as I can tell it is the only way I can get the jacoco jars.

Thrameos:
@marscher Okay I think everything is good to go on this one.  It didn't produce a coverage report (yet), but I am not sure if that is a problem with tokens or other permissions.  Thanks for getting this started.  Once we have working coverage reports I can start addressing it with improved testing.

Thrameos:
Do you mind if I merge this one?

marscher:
On 1/3/20 4:27 PM, Karl Nelson wrote:
> Do you mind if I merge this one?
> 
please go ahead

PR #534, opened 2019-11-12
Charsequence

### LINKED ISSUE #527: CharSequence from string
Certain classes use CharSequence rather than string such as the java.time library.   JPype currently does not convert a Python string to this type automatically, thus a forced conversion is required.  This should likely be corrected.

Thrameos:
This patch adds CharSequence to the automatic conversion list. 

Fixes #527

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -4,6 +4,8 @@ Changelog
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
 - **Next version - unreleased**
+  - Automatic conversion of CharSequence from Python strings.
+
   - java.lang.AutoCloseable supports Python "with" statement.
 
   - Updated getJVMVersion to work with JDK 9+.
--- a/native/common/include/jp_jniutil.h
+++ b/native/common/include/jp_jniutil.h
@@ -22,6 +22,7 @@ namespace JPJni
 	extern jclass s_ObjectClass;
 	extern jclass s_ClassClass;
 	extern jclass s_StringClass;
+	extern jclass s_CharSequenceClass;
 	extern jclass s_NoSuchMethodErrorClass;
 	extern jclass s_RuntimeExceptionClass;
 	extern jclass s_ProxyClass;
--- a/native/common/include/jp_stringclass.h
+++ b/native/common/include/jp_stringclass.h
@@ -21,7 +21,7 @@ class JPStringClass : public JPClass
 {
 public:
 
-	JPStringClass();
+	JPStringClass(jclass cls);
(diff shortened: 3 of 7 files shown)


marscher:
Thanks for the provided fix. I have a couple of question regarding the untyped usage of a string conversion class here. Is it now possible to literally set any jclass as a string conversion class? What would be the consequence, if this can be set somewhere, where we did not intend it and potentially harm security then?
If my concern is justified, wouldn't it be better to define a list of allowed string conversion classes?

Thrameos:
Well the C++ conversion wrappers can be applied to any class,  but Java will not accept an object which is the wrong type.  In this case CharSequence accepts String as String is derived from it.  If this was applied to another type it would take the Python string convert it to a Java string and then produce an exception when Java checks the types.  Of course there are some cases this may not work as JNI is sometimes prone to failure to check which usually results in a segmentation fault.  For example, Java always checks the type of an object but it has no way to know if the item is actually an object because everything is passed as a union of types.  Thus passing an integer to an object will cause Java to try to access the memory that the integer points to even if that is outside of the memory space of the JVM.  We do a pretty good job of shielding the user from this with the exception of the memory backed array types where I have already noted issues.

In general, I have been working towards removing all of the specialized wrappers for Java object types and instead give a typehints structure with a list of accepted conversions supplied either by the system or the user.  Thus we would be able to hands Python Path to Java Path, Python Date to Java Date, Python SQL to Java SQL, etc transparently by simply adding a conversion to the typehints for the Java class.  It will consult either if the types is specific, derived from, or has an attribute (duck).  The hard part is getting reasonable performance out of it as I do not want to call arbitrary python code during the resolution phase of the process.  Once the method has been resolved, we can perform the conversion either in C++ or by calling back to Python to get the appropriate Java object type.  Thus rather than doing this in C++ with fixed tables, I would have just added Python String to Java String conversion to the CharSequence accepted conversion list.  You could apply this to other classes, but the result would just be Java receiving a Java String as one of the arguments something you can achieve already with casting.   The wrappers for primitives have to stay as they are as they require fixed rules to prevent passing arbitrary pointers to Java. Unfortunately, I am not there yet so the best I could do in this limited case is give the JPString conversion class to the CharSequence.

This may have some security implications.  As we provide generalized monkey patching to the Java classes there is always a possibility to insert arbitrary python code into our Java conversion process.  But I do not believe that Python has anything approaching a secure class model thus we aren’t adding anything that wasn’t there already, and as Java will force the type checking on its side, it shouldn’t be possible to call something which isn’t already available.

marscher:
Thanks for the detailed explanation. It makes much more sense to me now.

PR #541, opened 2019-12-07
Updated keywords

### LINKED ISSUE #540: "and" and other keywords should be added to _pykeywords.py
the keyword "and" is missing in the `KEYWORDS` in _pykeywords.py as well as 
{'False', 'True', 'async', 'await', 'nonlocal', 'with'}

perhaps 
```
import keyword;
KEYWORDS=set(keyword.kwlist)
```
could be used

Thrameos:
Fixes #540 

@marscher if you have a chance we should do a maintenance release.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -4,6 +4,8 @@ Changelog
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
 - **Next version - unreleased**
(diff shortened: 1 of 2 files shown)


marscher:
Thanks! Is the changelog up to date, we can do this maintenance release.

Thrameos:
I believe that the change log is current.  I have been posting everything as work is performed.

PR #548, opened 2019-12-16
updated version and changelog (release date)

marscher:


Thrameos:
LGTM

marscher:
The pypi source package is in place. Do you want to built the wheels?

Thrameos:
Yes, I should be able to do so tonight.

PR #551, opened 2019-12-24
Improved exception stacktraces

Thrameos:
This is an attempt to integrate Java stacktrace into Python traceback system to create seamless error reporting thus negating the need for stacktrace() to be called to extract the error information.  The pull request is to test functionality in the test matrix.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -4,6 +4,10 @@ Changelog
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
 - **Next version - unreleased**
+  - C++ and Java exceptions hold the traceback as a Python exception
+    cause.  It is no longer necessary to call stacktrace() to retrieve
+    the traceback information.
+
   - Speed for call return path has been improved by a factor of 3.
 
   - All exposed internals have been replaced with CPython implementations
--- a/native/common/include/jp_exception.h
+++ b/native/common/include/jp_exception.h
@@ -110,7 +110,7 @@ public:
 		return line_;
 	}
 } ;
-typedef list<JPStackInfo> JPStackTrace;
+typedef vector<JPStackInfo> JPStackTrace;
 
 typedef union
 {
@@ -200,5 +200,7 @@ public:
 	int line;
 	string message;
 } ;
+
+void JPException_init();
 
 #endif
--- a/native/common/jp_env.cpp
+++ b/native/common/jp_env.cpp
@@ -170,6 +170,7 @@ void JPEnv::loadJVM(const string& vmPath, const StringVector& args, bool ignoreU
 	JP_TRACE("Initialize");
 	try
 	{
+		JPException_init();
 		JPJni::init();
 		JPClassLoader::init();
 		JPTypeManager::init();
--- a/native/common/jp_exception.cpp
+++ b/native/common/jp_exception.cpp
@@ -15,6 +15,9 @@
 
  *****************************************************************************/
 #include "jpype.h"
+#include <frameobject.h>
+PyObject* PyTrace_FromJPStackTrace(JPStackTrace& trace);
+PyObject *PyTrace_FromJavaException(JPJavaFrame& frame, jthrowable th);
 
 JPypeException::JPypeException(jthrowable th, const char* msn, const JPStackInfo& stackInfo)
 : m_Throwable(th)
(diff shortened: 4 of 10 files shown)


Thrameos:
The remaining work on this is just to add #if over sections so that symbols that aren't in each of the versions.  I will update it when I get back from vacation.

Thrameos:
I think this is ready for review.  This pull request integrates the Java and Python exception systems so that Java exceptions appear as native Python exceptions.  

This required a lot more effort that I hoped.  Many elements of the Python exception system require access to private or unextendable resources.  It may be a good idea to try to get better support in Python for this, but for now this will have to do.

Thrameos:
This one has gone stale and will require a fair amount of work to pull forward.

Thrameos:
@marscher This one is freshened and ready to go.  Ignore the missing "ant" on macos, it appears to be a CI bletch.  Once we have these resolved I can move on to merging the head back into development so we can move forward to classhints and its big speed improvement.

Thrameos:
This one is back to the drawing board until I can remerge it with master.

Thrameos:
@marscher back alive and ready for review.  This pull request hacks the Python error reporting system to attach Java exception information as the "cause" of an exception thrown from Java or C++ so that we don't need a special stacktrace() command or produce long exception messages with "at... at...".

```
  File "ExceptionTest.java", line 57, in jpype.exc.ExceptionTest.throwChain
  File "ExceptionTest.java", line 62, in jpype.exc.ExceptionTest.method1
  File "ExceptionTest.java", line 67, in jpype.exc.ExceptionTest.method2
Exception: Java Exception

  File "test.py", line 5, in <module>
    cls.throwChain()
jpype._jclass.RuntimeException: java.lang.RuntimeException: Inner
``

There was a bit of wizardary here as Python doesn't let one directly mock a stacktrace that is usuable with native methods because they directly access internal fields so the method tb_create in jp_exception.cpp replicates creating the structures used internally in Python.  It requires the Java code be compiled with "lines" as the debugging flag if accurate lines are to be reported on the trace (otherwise we just get -1).   

I am slightly worried about the long term stability on this because of the level of structure hacking involved.  I couldn't attract any attention from the Python devs on the forums I posted on (for this or other internals questions) so this likely the best I can do.  Until they add a "make exception from a list of stack info" in the API we will have to hack it.

**Comment by marscher** [comment]
I share your concern about long term stability. The CPython looks tricky. Maybe there is a good reason for the separated stacktrace generation functionality in the first place. If you think the benefits of this outrun the additional future overhead, I'd be to merge this.

**Comment by Thrameos** [comment]
I believe @KOLANICH can speak best to this (see #523).  The problem is specifically that a default unhandled exception goes to the default Python printer which is not aware of the Java stack trace so it prints just the Python portion.  Thus ever piece of code ends up looking like.

```
try:
    ... bunch of code...
catch JException as ex:
   print(ex.stacktrace())
   throw ex
```

Secondarily the C++ portion of the exception is being attached to the message so rather than getting "Exception here",  we get "Exception here
   at  fred
   at  george
   at bob"

Thus we have two very different things both of which are not very Pythonic.   The problem being we really need some input from a CPython core developer to give use some direction on how to implement this Pythonically.  I am fine with this hack solution for now if they plan give a "PyErr_AddCauseWithStackTrace(PyObject* Exception, PyObject *StackFrameList)" at some point in the future that I can then remove this code.   At least when I worked in Gnome had an IRC channel and they monitored the SVN commit log on email I could get input if needed.

marscher:
I can see the improvements of better error messages. One question remains, what happens, when people still call ex.stacktrace() in their codes?

Thrameos:
Stacktrace() still contains the Java stacktrace just like before.  There is no change to existing code.   Instead if you do not handle a JException different than a Python exception you will get the stacktrace information as the “cause”.   This of course only works because we have dropped Python 2 where cause was not supported.  Thus if the code caught JException first and calls print(ex.stacktrace())  they get the stacktrace (in the expected Java order inner to outer).  If they do not handle the exception or call the Python methods for printing the stacktrace, they get the Python portion of the stacktrace proceeded by the C++ or Java in a fake exception with the exception information order Pythonically from outer to inner.

Once I get Java tied back into the Python system (work in progress), then if a Proxy throws in Python, you get a set of exceptions( Python caused… Java caused… Python caused).  If gets a bit tricky if a Java exception is thrown in a Proxy as Java exception will have a cause the skips over the Python portion, but otherwise it should be pretty smooth.

The only thing this pull request drops is the “…at foo.cpp:1\n...at bar.cpp:2”  from being attached to the message when an internal error happens in JPype.  But that information was primarily for reporting bugs to the project and not for the user.  Thus attaching it as the cause instead should be better.

PR #553, opened 2020-01-04
Fix for JInterface implementation.

### LINKED ISSUE #547: JInterface implementation
While working on JPype 0.8 code, I found that the implementation of JInterface was incorrect.  The idea was to create a way to determine if something was an interface primarily for the purposes of setting up JProxy.  But as I did not know about ``__instancecheck__`` is was implemented as part of the object tree.  Because it was only supposed on non concrete objects, the mro is hacked so that it disappears from the type tree, which is a really bad hack.  

The correct implementation should have had JInterface being a meta class like JClass.  The primary difference being what method you use to perform the check "isinstance" vs "issubclass".  As this is just a meta interface it should properly be "isinstance" but is it currently implemented as "issubclass".  The documentation is a bit mixed as some sections also give it as ``isinstance`` but that currently does not work.

I believe the best course of action is to reimplement it with a metaclass that overrides both ``__instancecheck__`` and ``__subclasscheck__`` so that either methods will produce desired result.  Then modify the documentation so that we always recommend the proper ``isinstance`` method.  The likely impact on users either way is very low.  Normal users would almost never need to condition the code on whether or not a particular class is an interface.

Thrameos:
Fixes #547


Thrameos:
This one appears to depend on removing Python 2.7 unless I want to reformulate.

Thrameos:
@marscher This is ready for review.  It purpose is to remove a evil type hack that I was using to put JInterface into the type tree.  It also updates the documentation.  

There was one minor change in the test having to do with the exception that was being thrown.  It was defaulting to a Java exception and is now going to a more correct Python TypeError.  This pull request also contains a backport of pyjp_class from ClassHints pull request.   The reason of the backport is that when we add the correct extern "C" declaration it revealed there was a path that could pass C++ exceptions to C causing a segfault.  I don't think it was possible to exercise it under any normal situation.  I am only backporting when I am working on a section for other things in a code section as backporting is time consuming.

Thrameos:
Removal of numpy will depend on this one as backported pyjp_class is required.

PR #554, opened 2020-01-04
Fix local referencing issues.

Thrameos:
Fixes referencing issues enabling Xcheck:jni without segfault.


marscher:
Good catch! I also like the permanent reference checking in the test suite. Good job.

Thrameos:
No problem.  Next I will try to increase our coverage score and a cleanup pass to remove Python 2.

marscher:
On 1/4/20 5:42 PM, Karl Nelson wrote:
> No problem.  Next I will try to increase our coverage score and a cleanup pass to remove Python 2.
> 
sounds good.

PR #559, opened 2020-01-08
Support for building manylinux1 wheels

### LINKED ISSUE #557: Publish manylinux wheels for v0.7.1
`pip install JPype1==0.7.1` requires a compiler toolchain.  If manylinux wheels are published, this requirement is eliminated.

Thrameos:
Fixes #557 

This pull request adds scripts to use docker to build manylinux wheels.


marscher:
Hi, this will greatly simplify things. Is there also a Dockerfile to specify the containers details?

Thrameos:
Yes the dockers are from github manylinux repo.   We can likely build our own if you would like.  

https://github.com/pypa/manylinux

marscher:
On 1/8/20 2:23 PM, Karl Nelson wrote:
> Yes the dockers are from github manylinux repo.   We can likely build our own if you would like.
> 
> https://github.com/pypa/manylinux
> 
We don't need the jdk anymore, do we?

Thrameos:
No building the sdist locally removes that dependency.  Thus could yse manylinux out of box

Thrameos:
We cant actually test then in the docker without java, but we can build them.

marscher:
On 1/8/20 2:28 PM, Karl Nelson wrote:
> No building the sdist locally removes that dependency.  Thus could yse manylinux out of box
> 
pretty cool :) Thanks

marscher:
On 1/8/20 2:30 PM, Karl Nelson wrote:
> We cant actually test then in the docker without java, but we can build them.
> 
testing has been done before releasing, so the built version should be fine.

Thrameos:
Do you know how to push the results to pypi?  I pushed them to github release for now.

marscher:
I can upload them.

marscher:
I guess we also want to have builds for Python 3.8

Thrameos:
Okay i can add them to the build but it wont be until evening PST.

marscher:
400 Binary wheel 'JPype1-0.7.1-cp35-cp35m-linux_x86_64.whl' has an unsupported platform tag 'linux_x86_64'

I think it should be manylinux. Somethings wrong. I guess these need to be converted somehow.

Thrameos:
Perhaps we need to add auditwheel repair to the build pattern.

Thrameos:
There is a manylinux demo on github that i was working from.  Perhaps i missed a step.

Thrameos:
Also possible the no index flag was a requirement.  Will have to investigate further.

PR #560, opened 2020-01-08
Fixes tags for manylinux

Thrameos:
@marscher Okay second attempt on this one.  Please try to upload the whls and see if it is working now.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -2,5 +2,8 @@
 set -e -x
 echo ====== $PY $VER
 PYBIN=/opt/python/$PY/bin
-"${PYBIN}/pip" install -r /io/dev-requirements.txt
-"${PYBIN}/pip" wheel /io/JPype1-$VER.tar.gz -w /io/wheelhouse
+if [ ! -e /io/wheelhouse/JPype1-$VER-$PY-linux_$PLAT.whl ]; then
+	"${PYBIN}/pip" install -r /io/dev-requirements.txt
+	"${PYBIN}/pip" wheel /io/JPype1-$VER.tar.gz --no-index -w /io/wheelhouse
+fi
+auditwheel repair /io/wheelhouse/JPype1-$VER-$PY-linux_$PLAT.whl -w /io/wheelhouse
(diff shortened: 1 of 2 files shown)


marscher:
LGTM. Now PyPI complains about the language tag we use in setup.py:

```
HTTPError: 400 Client Error: Invalid value for classifiers. Error: 'Programming Language :: Java :: 8' is not a valid choice for this field for url: https://upload.pypi.org/legacy/
```

marscher:
https://pypi.org/classifiers/ It seems there is only Java allowed, but now specific version. I think it sufficient to have the supported versions only in the long_description of setup.py. That way this information is also bound to a jpype release.

Thrameos:
Okay, so I need to edit it and resubmit?

Thrameos:
Okay I think I am very confused.  I looked at the setup and the unpacked the wheels and I don't see "Java :: 8" anywhere.

I thought we deleted the Java classifier a long time ago. (Jun 22, 2019, patch "Changes for pypi push") 

```
    classifiers=[
        'Programming Language :: Python :: 2',
        'Programming Language :: Python :: 2.7',
        'Programming Language :: Python :: 3',
        'Programming Language :: Python :: 3.5',
        'Programming Language :: Python :: 3.6',
        'Programming Language :: Python :: 3.7',
    ],
```

Has it reappeared somewhere?

marscher:
Today it worked. I guess it was some kind of pypi caching hickup. Could you please also create a Windows build for Python3.8? Then we should be done.

Thrameos:
Sure.   I will also look to automate the building of windows at some point.  I see some dockers that may work .  That just leaves macosx as needed before our binary would be complete.

marscher:
On 1/10/20 4:37 PM, Karl Nelson wrote:
> 
> Sure.   I will also look to automate the building of windows at some point.  I see some dockers that may work .  That just leaves macosx as needed before our binary would be complete.
> 
> 

I think Azure pipelines would be well suited for this task, as it 
provides all three major platforms.

PR #566, opened 2020-01-15
Change requirements for wheels

Thrameos:
Yet another attempt to improve the manylinux wheel build.  In this one we make the bdist_wheel command update the requirements for numpy to match the binary it was built against.  Assuming it is good I will push another set of wheels to the distribution.


Thrameos:
@marscher Hopefully this works now.  In the docker script you can pin the version for each binary build to a specified target.  Obviously selecting one on the list of pypi available manylinux numpy distributions results in the fastest, most likely to be compatible build.  Pinning it older forces docker to rebuild numpy which then requires an additional step of calling the fix on the build numpy so that it doesn't create binaries that depend on the version that the docker built.   The wheel script then makes sure that bdist requirements are kept up to date so that pip pulls a compatible binary version.  This is only reflected in the manifest for pip.  The source version requirements remain as current.

If you would like a different set of versions for the pinning just update this pull request and I will rerun the docker.  I am not sure what other python modules do regarding binary versions so I can't really tell if this is an evil hack or what the version system was intended for.

Thrameos:
Depending on success of numpy removal pull request perhaps we should just wait two weeks and release 0.7.2 without need for numpy?

PR #568, opened 2020-01-17
Customizer for byte[] to bytes

### LINKED ISSUE #567: Bytes conversion
This issue has come up twice in the past month where someone wanted to convert a Java byte[] to a corresponding native Python bytes type.  The current method is to call `[:]` which is not very intuitive and a side effect of the slice implementation rather than good practice.  Worse others have tried to use `toString` which is completely different functionality wise.     

I recommend that we add the appropriate conversion customizers to JPype to make this a bit more intuitive.  We should add the customizer to the library, add a test for the customizer, and last add it to the documentation.

See #565 and 
https://stackoverflow.com/questions/59608488/python-jpype-get-value-from-instance-of-java-class/59773940#59773940

Thrameos:
Fixes #567


Thrameos:
Guess this one is on hold until we scrub out Python 2.7 as well.  Also need special handling when numpy is not installed.

Thrameos:
@marscher This one is ready to go.

I don't like having to iterate everything through a list comprehension as it has very poor performance, but it is required until I resolve the numpy/byte buffer issues.  There is a minor issue that Python defines bytes as unsigned and Java defines them as signed,  but as the point is raw byte representation I am not sure it makes any real difference.

Thrameos:
Numoy removal will likely consume this pull request as bytes support buffer mech.

marscher:
Please reopen, if there is still need to be merged.

Thrameos:
I think the only thing left in this is a documentation update and one test code to make sure that bytes() actually works. We do a lot of testing on memoryview but we do not specifically test "bytes" and "bytearray".   Perhaps I should just add that as a suite to the test battery.

Issue #570, opened 2020-01-22
array.fromstring was removed in favor of tobytes in Python 3.9


tirkarthi:
https://bugzilla.redhat.com/show_bug.cgi?id=1791951

https://github.com/jpype-project/jpype/blob/49faa31c26a10377ebdeb4fa6a89d5ba9e8f97f5/setupext/build_thunk.py#L36

Thanks for the library!

Thrameos:
Thanks for the bug report.  We will get it fixed shortly.

hugovk:
Please see PR https://github.com/jpype-project/jpype/pull/572.

PR #572, opened 2020-01-23
Fix array.array.frombytes for Python 3.9

hugovk:
Fixes https://github.com/jpype-project/jpype/issues/570.

* array.fromstring was removed in favour of frombytes in Python 3.9
* Python 2 has been dropped, remove PyPy2 from CI
* Sequence is in collections.abc in Python 3.3+. This is commented code, but updated just in case it's uncommented in future

nightly on Travis CI now points to Python 3.9, and it's now green :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ python:
   - '3.5'
   - '3.6'
   - 'nightly'
-  - 'pypy'
   - 'pypy3'
 
 env:
@@ -16,7 +15,7 @@ env:
     - PIP=pip
     - INST=""
     # Moved numpy testing to Xenial
-    #   Testing with and without numpy on more than 2.7/3.7 for each arch
+    #   Testing with and without numpy on more than 3.7 for each arch
     #   is likely a waste.  The paths to a CPython module do not really
     #   change, but byte order and major version may make a difference.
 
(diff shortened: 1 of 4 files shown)


Thrameos:
Fail is a belch in the CI system.

hugovk:
Thanks for merge! Do you have a rough schedule for the next release? Cheers!

Thrameos:
I will have to consult with @marscher regarding the next release.  JPype is currently struggling because of incompatibility between with and without numpy on the distribution.  My general opinion is that binary modules can’t use other binary modules as extras as it basically forks the distribution.   The portion that we are using from numpy is vanishing small so I am making a pull request that replaces numpy use with a class that replicates the same features.   I am guessing it will be about two weeks of calendar time clean up all the edge cases that are likely to be generated on that effort.   After that we should be ready for a release.

Thrameos:
Also we have two mystery segfaults on our issues list.  I haven’t analyzed them so I can’t put a time on correction, but I would hope to have time to deal with them before another release.  If they look too nasty I guess I can leave them for a future release.

Thrameos:
@hugovk we have completed a new release.  Sorry for the delays.

hugovk:
Thank you!

PR #576, opened 2020-01-27
Proxy convert

### LINKED ISSUE #574: Issue when passing Python implementation of Java Interface back to Python for comparison in HashSet
Thank you for this great software!

I ran into a problem where I implement a Java interface in Python and then add these objects into a HashSet and try to remove such objects from that HashSet: the remove operation calls equal() / __eq__ and this obtains as first argument nicely the Python object, but as second argument an instance of org.jpype.proxy.JPypeInvocationHandler and this object does not permit to access members of the Python object.

Now I am not sure this should work, but it would be very nice for my use case. I use v0.7.1 built from github.

The following is a minimal example that demonstrates the issue, the real code does more useful stuff.

The idea is that the Test1Class constructor gets two sets of objects and subtracts the second from the first. The Test1Interface serves as the interface for the objects and at the same time as provider of the two sets (to keep the example small).

Test1Interface.java
```
package at.ac.tuwien.kr.hexlite.test;

import java.util.ArrayList;
import java.util.HashSet;

public interface Test1Interface {
   public int hashCode();
   public boolean equals(Object o);
   public String toString();
   public HashSet<ArrayList<Test1Interface> > provide1();
   public HashSet<ArrayList<Test1Interface> > provide2();
};
```

Test1Class.java
```
package at.ac.tuwien.kr.hexlite.test;

import java.util.ArrayList;
import java.util.HashSet;

public class Test1Class {
   public Test1Class(Test1Interface s) {
      HashSet<ArrayList<Test1Interface> > result = new HashSet<ArrayList<Test1Interface> >();
      System.err.println("initial: "+result);
      for(ArrayList<Test1Interface> x : s.provide1()) result.add(x);
      System.err.println("after add: "+result);
      for(ArrayList<Test1Interface> y : s.provide2()) result.remove(y);
      System.err.println("after remove: "+result);
   }
}
```

My Python script:
```
import jpype
jpype.startJVM(convertStrings=False)
from jpype import java
from jpype.types import *

print("Version:"+jpype.__version__)
Test1Interface = JClass("at.ac.tuwien.kr.hexlite.test.Test1Interface")
Test1Class = JClass("at.ac.tuwien.kr.hexlite.test.Test1Class")

@jpype.JImplements(Test1Interface)
class Test1Implementation:
	def __init__(self, something):
		self.something = something

	@jpype.JOverride
	def hashCode(self):
		return int(hash(self.something) & 0x7FFFFFFF)

	@jpype.JOverride
	def provide1(self):
		s = java.util.HashSet()
		
		tup1 = java.util.ArrayList()
		tup1.add(Test1Implementation("foo"))
		s.add(tup1)

		tup2 = java.util.ArrayList()
		tup2.add(Test1Implementation("bar"))
		s.add(tup2)
		return s

	@jpype.JOverride
	def provide2(self):
		s = java.util.HashSet()
		
		tup1 = java.util.ArrayList()
		tup1.add(Test1Implementation("bar"))
		s.add(tup1)

		tup2 = java.util.ArrayList()
		tup2.add(Test1Implementation("baz"))
		s.add(tup2)
		return s

	def __eq__(self, other):
		print("__eq__ repr(self)=%s vs repr(other)=%s", repr(self), repr(other))
		print("__eq__ str(self)=%s vs str(other)=%s", str(self), str(other))

		# does not work, other seems to be a Java-only-usable object - can I cast other to Test1Implementation?
		return self.something == other.something

	@jpype.JOverride
	def equals(self, other):
		return self.__eq__(other)

	def __str__(self):
		return str(self.something)

	@jpype.JOverride
	def toString(self):
		return str(self.something)

i = Test1Implementation("nothing")
c = Test1Class(i)
jpype.shutdownJVM()
```

This fails with:

```
Version:0.7.1
initial: []
after add: [[bar], [foo]]
__eq__ repr(self)=%s vs repr(other)=%s <__main__.Test1Implementation object at 0x7f2825bda8d0> <jpype._jclass.org.jpype.proxy.JPypeInvocationHandler object at 0x7f2825be3d90>
__eq__ str(self)=%s vs str(other)=%s bar bar
...
  File ".../test1minimal.py", line 64, in <module>
    c = Test1Class(i)
  File ".../python3.7/site-packages/jpype/_jobject.py", line 95, in __init__
    jv = self.__class__.__javaclass__.newInstance(*args)
jpype._jclass.RuntimeException: java.lang.RuntimeException: Python exception thrown: AttributeError: 'org.jpype.proxy.JPypeInvocationHandler' object has no attribute 'something'
```

I would appreciate any help what I could do differently or whether this is actually should work as I try to use it.

Thrameos:
Fixes #574 

This patch fixes one memory leak in which the JProxy handler was being reregistered every time a proxy was called and makes it such that proxies always get the most derived class when passing arguments rather than the class declared in the interface.   This bug was preventing the appropriate type converter from being found with a proxy was passed through.  It would also have prevented methods defined in derived type from being accessed as the wrapper was being forced to use the base class.

Pull request contains documentation, fix, and a test for the previous behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,5 @@ doc/html
 native/jars
 docker/*.tar.gz
 docker/wheelhouse
+/.eggs
+/tmp*
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,11 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 - **Next version - unreleased**
 
+  - Fixed memory leak in Proxy invocation
+
+  - Fixed bug with Proxy not converting when passed as an argument to Python functions
+    during execution of proxies
+
 
 - **0.7.1 - 12-16-2019**
 
--- a/native/common/jp_proxy.cpp
+++ b/native/common/jp_proxy.cpp
@@ -18,12 +18,12 @@
 
 namespace
 { // impl detail, gets initialized by JPProxy::init()
-	jclass handlerClass;
-	jmethodID invocationHandlerConstructorID;
-	jfieldID hostObjectID;
+jclass handlerClass;
+jmethodID invocationHandlerConstructorID;
+jfieldID hostObjectID;
 
-	jclass proxyClass;
-	jmethodID getInvocationHandlerID;
+jclass proxyClass;
+jmethodID getInvocationHandlerID;
 }
 
 JNIEXPORT jobject JNICALL Java_jpype_JPypeInvocationHandler_hostInvoke(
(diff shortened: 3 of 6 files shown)


peschue:
The bugfix also makes my original project run without exceptions. Thank you very much!

Issue #562, opened 2020-01-12
Add Python 3.8 to test matrix


Thrameos:
Now that we are building for 3.8 we need to add it to the test matrix.

Thrameos:
Fixed by nonumpy

PR #577, opened 2020-02-12
Remove numpy

### LINKED ISSUE #528: Performance of python wrapping?
We are investigating using jpype to provide python bindings to [powsybl](https://github.com/powsybl/powsybl-core) framework, a java framework aimed at electrical power system modelling and computations.

First experience is very promising, congratulations for the great job done so far ! :)

However, the main drawback we have identified so far is the performance of building python wrappers around our java objects.

**An illustrative use case:**

For example, a large electrical network may contain around 10000 "electrical buses", represented as java objects. Retrieving the list of thoses buses in python takes around 0.5s, which is a lot for our use cases :
```python
network = ...
buses = [b for b in network.getBusView().getBuses()]
```

This is easily reproduced without any framework specific jar, for example:
```python
# Create a java array list containing java objects
from java.lang import Object
from java.util import ArrayList

start = time.time()
javaList = ArrayList()
for i in range(0, 10000):
    javaList.add(Object())

print('Fill time = {}s'.format(time.time() - start))

# Gather all objects in a python list
start = time.time()
l = [javaList.get(i) for i in range(0, 10000)]
print('Get time = {}s'.format(time.time() - start))
```

This gives the following result:
```
Fill time = 0.28787922859191895s
Get time = 0.28006649017333984s
```

If we want to improve the performance or our binding, I think we are left with 2 solutions :
 - Provide in our java framework a more vector-oriented API around our core object model, which could leverage numpy support. But this may not completely suit our needs.
 - Try to improve the performance of building python wrappers in jpype ? *Do you foresee future evolutions that could improve this ?*

**A more detailed analysis:**

A C profiling with google perftools shows that for the "get" part above, less than 5% of time is actually spent in JNI calls, whereas around 60% of time is spent in `JPClass::convertToPythonObject` for java calls results, and 20% of time is spent in `JPPythonEnv::getJavaValue`.

A python profiling of the same code seems to show that this time is mainly spent in python code, notably `JObject.__init__`.

I would think that moving some of jpype python code to the C extension, for example the JObject class, could greatly improve this situation (for example, the `__javavalue__` field could be retrieved directly from the corresponding C struct instead of going through the whole "get attribute" machinery) :
do you think this could be an interesting direction to go for, or does it sound too much of a development effort (or maintainability loss...) ?

Thank you!

### LINKED ISSUE #536: Boxed types as primitives
@marscher  I am trying to improve the speed in the JPype 0.8.  Unfortunately, I have run into an issue that would break compatible.  Currently boxed types inherit from primitives so that they will function anywhere that requires a number type without requiring a forced conversion.  This is forcing everything is to go through a type system which pulls in a huge number of python calls for each routine.  I can remove this restriction, but the isinstance will no longer work for boxed types.  I will have to make them work like numbers rather than being int or floats.

Any opinions?

### LINKED ISSUE #542: JChar does not accept single characters for initializer
Because char is listed as an integer rather than its own special type we can't initialize it from a single char string like one would expect.

```
import jpype
jpype.startJVM()
print(jpype.JChar('a'))
```

Gives the unexpected result.

ValueError: invalid literal for int() with base 10: 'a'

I am not sure if Java primitives should really derive from Python primitives through it seems reasonable for them to do so.  But char really needs to be a special case as we are not supposed to be able to do math with it.

### LINKED ISSUE #563: Minimum Numpy versions for latest JPYPE version is no longer numpy>=1.6
By using versions between 1.6 and 1.15.4, importing jpype will results in the following error: 

```
>>> import jpype
ModuleNotFoundError: No module named 'numpy.core._multiarray_umath'
  File "/usr/local/lib/python3.6/site-packages/jpype/__init__.py", line 17, in <module>
    import _jpype
ImportError: numpy.core.multiarray failed to import
```

Updating the minimal requirements to numpy>=1.16.0 will solve the issue.

See https://github.com/jpype-project/jpype/blob/afe8a8bc002959ba0a727795d6be68fbd6aabefc/setup.py#L41

### LINKED ISSUE #564: Consider removing numpy
One of the big hang ups is currently numpy.  Our actual use of numpy is just as an interface that supports memoryview byte transfers and slicing.  There is no reason we cant just provide an array slice object with the appropriate apis implemented and drop numpy entirely.  We can still get high levels of integration by implementing `__array_interface__` without actually using numpy binarary functions. 

Related to #133, #563

### LINKED ISSUE #567: Bytes conversion

### LINKED ISSUE #573: jpype.startJVM failing with java.lang.ClassNotFoundException: org.jpype.Utility
Hi,

I try using Jpype (0.7.1) via JayDeBeApi (1.1.1) on python 3.7, JayDeBeAPI installed via pip in a conda environment on Windows 10 Pro, but keep getting the following error.

Any idea how I can fix that?

Thanks,
E.

```
conn = jaydebeapi.connect(
    jdbc_driver_name,
    connection_string,
    {'user': dsn_uid, 'password': dsn_pwd},
    jars=jdbc_driver_loc)

  File "C:\Program Files\JetBrains\PyCharm 2019.1\helpers\pydev\pydevd.py", line 1758, in <module>
    main()
  File "C:\Program Files\JetBrains\PyCharm 2019.1\helpers\pydev\pydevd.py", line 1752, in main
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files\JetBrains\PyCharm 2019.1\helpers\pydev\pydevd.py", line 1147, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "C:\Program Files\JetBrains\PyCharm 2019.1\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Users/.../postgres_jdbc_new.py", line 50, in <module>
    jars=jdbc_driver_loc,
  File "C:\Users\...\AppData\Local\conda\conda\envs\py37\lib\site-packages\jaydebeapi\__init__.py", line 381, in connect
    jconn = _jdbc_connect(jclassname, url, driver_args, jars, libs)
  File "C:\Users\...\AppData\Local\conda\conda\envs\py37\lib\site-packages\jaydebeapi\__init__.py", line 176, in _jdbc_connect_jpype
    jpype.startJVM(jvm_path, *args)
  File "C:\Users\...\AppData\Local\conda\conda\envs\py37\lib\site-packages\jpype\_core.py", line 227, in startJVM
    _jpype.startup(jvmpath, tuple(args), ignoreUnrecognized, convertStrings)
RuntimeError: java.lang.ClassNotFoundException: org.jpype.Utility
```

### LINKED ISSUE #575: convertToDirectByteBuffer leaks
The implementation of convert to direct byte buffer is incorrect.  It creates a view of the Python memory but never releases it.  It appears to be a misunderstanding of the way that byte buffers work (likely from long ago). 

To fix this problem, Java needs to call release buffer when complete rather than unreferencing the memoryview which it does now.  Additional support on the Java side may be required.

Thrameos:
This pull request removes numpy as an extra package and instead implements the buffer API directly.  To implement this pull request it was necessary to backport the speed patch from 0.8 so that I can directly derive from types.  This does have minor API breakage in that the return types on array slices change and thus a few comparison tests in the test suite changed.  

Implementation notes follow....

Buffers and NumPy removal
=========================

NumPy was used primarily for supplying Python buffers on slicing.  Numpy has
always been problematic for JPype.  As an optional extra is may or may not be
built into the distribution, but if it is compiled in it is required.
Therefore, is itn't really on "extra".  Therefore, removing it would make
distribution for binary versions of JPype much easier.

NumPy returns on slicing is but one part of the three uses of Python buffers in
JPype.  Thus to properly remove it we need to rework to remove it we need to
review all three of the paths.  These paths are

 - Conversion of Python buffers to Java arrays on setArrayRange. 
 - Conversion of Java arrays to slices (and then from slices to buffers so that
   they can be transferred to NumPy.)
 - Connecting bytearray to Java direct byte buffers.

We reviewed and revised each of the paths accordingly.

On conversion of Python buffers, the implementation was dated from Python 2.6
era where there was no formal support for buffers.  Thus the buffer
implementation never consulted the buffer type to see what type of object was
being transferred nor how the memory was oriented.  This entire section had to
be replaced and was given the same conversion rules as NumPy.  Any conversion is
possible (including lossy ones like float to bool).  The rules to trigger the
conversion is by slicing just as with NumPy.  By replicating the rules of NumPy
we hide the fact that NumPy is no longer used and increase typesafety.  There
was a number of cases where in the past it would reinterpret cast the memory
that will now function properly.  The old behavior was a useless side effect of
the implementation and was unstable with machine architecture so not likely used
be a user.

The getArrayRange portion has to be split into two pieces.  Under the previous
implementation the type of the return changes from Python list to NumPy array
depending on the compile option.  Thus the test suite tested different
behaviors for each.  In removing NumPy we replace the Java array slice return
with a Java array.  Thus the type is always consistent.  The Java array that is
returned is still backed by the same array as before and has the start, end, and
step set appropriately.  This does create one change in behavior as the slice
now has left assignment (just like NumPy) and before it was a copy.  It is
difficult exercise this but to do so we have to copy a slice to a new variable
then use a second array dereference to assign an element.  Because we converted 
to either list or NumPy, the second dereference could not affect the original.

There is no way to avoid this behavior change without adding a large transaction 
cost.  Which is why NumPy has the exact same behavior as our replacement
implementation.  We could in principle make the slice read only rather than
allowing for double slicing, but that would also be an API change.  

There is one other consequence of producing a view of the original having to do
with passing back to Java.  As Java does not recognize the concept of an array
view we must force it back to a Java type at the JNI level.  We will force copy
the array slice into a Java array at that time.  Thus replicates the same
functionality.  This induces one special edge case as `java.lang.Object` and
`java.lang.Serializable` which are both implemented by arrays must be aware of
the slice copy.

Before we trigger the conversion to NumPy or list by calling the slice without
limits operator `[:]` on the array.  Under the new implementation this is
effectively a no-op.  Thus we haven't broken or forced any changes in the API. 

The third case of direct byte buffers also revealed problems.  Again the type of
the buffer was not being check resulting in weird reinterpret cast like
behaviors.  Additionally, the memory buffer interface was making a very bad
assumption about the referencing of the buffer by assuming that referencing the
object that holds the buffer is the same as referencing the buffer itself.  It
was working only because the buffer was being leaked entirely and was likely
possible to break under situations as the leaked buffer essentially locked the
object forever.

To implement all of this properly unfortunately requires making the Python
wrapper of Java arrays a direct type.  This is possible in JPype 0.8 series
where we converted all classes to CPython roots, thus our only choice is to
backport the JPype speed patch into JPype 0.7.

API change summary
------------------

 - The type of a slice is no longer polymorphic but is always a Java array now.
 - The unbounded slice is now a no op.
 - Buffer types now trigger conversion routines rather than reinterpret casting 
   the memory.
 - Direct buffers now guard the memory layout such that they work only with
   mutable bytearray like types.
 - Assignment of elements though double slicing of an array now affects the
   original rather than just doing nothing effective like before.

JPype Speed Patch
=================

Speed has always been an issue for JPype.  While the interface of JPype is great
the underlying implementation is wanting.  Part of this was choices made early
in the development that favors ease of implementation over speed.  This forced a
very thin interface using Python capsules and then the majority of the code in
a pure Python module.  

The speed issue stems from two main paths.  The first being the method
resolution phase in which we need to consider each method overload argument by
argument which means many queries applied to each object.  The second is the
object construction penality when we return a Java object back to Python.  The
object bottleneck is both on the cost of the wrappers we produce, but
additionaly all of the objects that are constructed to communicate with a Python
function.  Every int, list, string and tuple we use in the process is another
object to construct.  Thus returning one object triggers dozens of object to be
constructed.

We have addressed these problems in five ways

 - Improve resolution of methods by removing the two phase approach to resolving
   a type match cutting the resolution time in half.
 - Caching the types in C so that they don't have to go back to Python to execute a
   method to check the Python cache and construct if necessary.
 - Converting all of the base types to CPython so that they can directly access
   the C++ guts without going back through entry points.
 - Remove all Python new and init method from the Java class tree so that we
   don't leave C during the construction phase.  Thus avoiding having to
   construct Python objects for each argument used during object construction.
 - Adding a Java slot to so that we can directly access Java resources both
   increasing the speed of the queries and saving us an additional object
   during construction.

All of these are being implemented for JPype 0.8 series.  For now we are
backporting the last four to the JPype 0.7 series.  The first is not possible to
backport as it requires larger structural changes.

Lets briefly discuss each of the items

Method resolution 
---------------------

During method resolution the previous implementation had a two phase approach.
First is tried to match the arguments to the Java types in canConvertToJava.  In
this each Java argument had to walk through each possible type conversion and
decide if the conversion is possible.  Once the we have the method overload
resovled we then perform the conversion by calling convertToJava.  That would
then walk through each possible type conversion to find the right one and apply
it.  Thus we did tha work twice.

To prevent this from happening we need to reserve memory for each argument we
are trying to convert.  When we walk through the list and find a conversion
rather than just returning true, we place a pointer to the conversion routine.
That way when we call convertToJava we don't have to walk the list a second time
but instead go straight to the pointer to get the routine to execute. 

This change has two additional consequences. First the primary source of bugs in
the type conversion was a mismatch between canConvertToJava and convertToJava
thus we are removing that problem entirely.  The second and more important to
the user is that the type system is now open.  By installing a routine we can
now add a user rule.  Therefore if we need `java.sql.TimeStamp` to accept a
Python time object we just need to add this to the type conversion table at the
Python level.  This is implemented in the ClassHints patch.  About half of our
customizer code was to try to achieve this on a per method level.  Thus this
elimiates a lot of our current Python customizer code.  The remaining customizer
code is to rename Java methods to Python methods and that will remain.

Caching of Python wrappers
--------------------------------

In the previous implementation there was a text keyed dictionary that was
consulted to get type wrappers.  To access it C++ called to a Python function
that decided when to return a cached type and when to create a new one.  This
meant dozens of object constructed just to find the wrapper.  To solve this we
simply move the cache and add it to the JClass directly.  We have to back
reference the Python class so it can't go away while the JVM is running.

There is one section of code that also uses the wrapper dict in the customizers
which needs to decided does a wrapper already exist for the customizer.  We have
replaced these calls with methods on the module.

Conversion of the Base classes
-----------------------------------

JPype has a number of base classes (object, primitive, exception, string, array)
which hold the methods for the class.  If they are implemented as pure Python
than every access from C++ to these elements needs to create objects accordingly
when then are passed back through the module entry points to get back to C++.

We can avoid this by implementing each of these in CPython first at the module
layer and then extending them in the exposed module so that they have the same
outward appearance as before.  

We made one refinement during the conversion by implementing all of the CPython
classes using the Heep type API which has the distinct advantage that unlike
static types, it can be changed at runtime.  Thus from Python we can add
behavior to the heap types simply with by using `type.__setattr_`.  This was
a bit of a challenge as the documentation on heap types is much more sparse than
for static types.  However, after going through the process I would recommend
that all new CPython modules should use heap types rather than static as API is
much better and the result much more flexable and stable.  The only downside
being the memory footprint increases from 400 bytes to 900 bytes.  There are a
few rough spots in the heap type API in that certain actions like setting the
Buffer have to be added to the type after creation, but otherwise it is a big
improvement.  Now if all of the documentation would just drop the old static API
in favor of heap types it would be great.

Constructor simplifications
-------------------------------

In order to benefit from moving all of the base classes to C, we have to make
sure that derived classes do not transfer control back to Python.  Currently
this happens due to the factory nature of our classes.  The entry point for
JObject is shared between the construction of objects from Python and a return
from Java.  Thus we have to either separate the factory behavior by pushing
those types out of the type tree or pushing the factory behavior into the C
layer.

We have chosen to split the factories and use overrides of the type system in
the meta class to apply `isinstance` and `issubtype` behavior.  We can further
restrict the type system if we need to by adding verifications that the
`__new__` and `__init__` methods must point the original base class
implementations if need.  Howver, we have not taken this step as of yet.
The split approach effectively removes these heavy elements from type creation.
The concequence of this is that means all of the rest of logic needs to be in
CPython implementation.  These can be rather cumbersome at times.

It is always a slippery slope when pushing code from Python back to CPython.
Some thing are needed as they are on the critical path while others are
called only occasionally and thus represent no cost to leave in Python.  On the
other hand some things are easy to implement in CPython because the have
direct access rather than having to go through a module entry point.  We have
gone with the approach that all critical path and all code the eliminates the
need for an entry point should be pushed back to C.

Java Slots
-------------

In order to get any reasonable speed with Python, the majority of the code
needs to be in C.  But additionally there needs to be the use of slots which are
hard coded locations to search for a particular piece of information.  This
presents a challenge when wrapping Java as we need a slot for the Java value
structure which must appear on Python object, long, float, exception, and type.
These different types each have their own memory layout and thus we can't just
add the slot at the base as once something is added to the base object it can
no longer be used in multiple inheritance.  Thus we require a different
approach.

Looking through the CPython source, we find they have the same quandary with
respect to `__dict__` and `__weakref__` slots.  Those slots do not appear one
the base object but get added along the way.  The method they use is to add
those slots to the back of the object be increasing the basesize of the object
and then referencing them with two offset slots in the type object.  If the type
is variable length the slot offset are negative thus referencing from the end of
the object, or positive if the object is a fixed layout.

Thus we tried a few formulations to see what would work best.

Broken tree approach
------------------------

The problem with just directly adding the slots in the tree is that the Java
inheritance tree forces the order of the Python tree we have to apply.  If we
add a slot to `java.lang.Object` we have to keep the slot on
`java.lang.Throwable` but that is not possible because Throwable requires it to
be derived from Python `Exception`. Thus if we are going to add a slot to the
base we would have to break the tree into pieces on the Python side.  This is
possible due to Python inheritance hacking with some effort.

But this approach had significant down sides. When we go to access the slot
we have to first figure out if the slot is present and if not then fall back to
looking in the dictionary.  But one of the most common cases is one in which the
item has no slot at all.  Thus if we have to both look for the slot and then hit
the dictionary, this is worse than just going to the dictionary in the first
place. Thus this defeats the point of a slot in many cases.

Python dict approach
------------------------

We attempted the same trick by increasing the basesize to account for our extra
slot.  This leaves to difficulties.  First, the slot has no offset so we need to
find it each time by implying its location.  Second, the extra objects have to
be "invisible" during the type construction phase, or Python will conclude the
memory layout of the object is in conflict.  We can fool the type system by
subtracting the extra space from the type during the declaration phase and then
adding it back after the base types are created.  

This approach failed because the "invisible" part is checked each and every time
a new type is added to the system.  Thus every dynamic type we add checks the
base types for consistency and at some point the type system will find the
inconsistency and cause a failure.  Therefore, this system can never be robust.

Dict and weakref appear to be very special cases within the Python system and as
there is no general facility to replicate them working within the system does
not appear to be viable.

Memory hacking approach
--------------------------------

The last system we attempted to mess with the memory layout of the object during
the creation phase to append our memory after Pythons.  To do this we need to
override the memory allocator to allocate the requested memory plus our extra.
We can then access this appended memory by computing the correct size of the
object and thus our slot is on the end.

We can test if the slot is present by looking to see if both `tp_alloc` and
`tp_finalize` point to our Java slot handlers.  This means we are still
effectively a slot as we can test and access with O(1).

The downside of this approach is there are certain cases in which the type of an
object can be changed during the destruction phase which means that our slot can
point to the wrong space if the basesize is changed out from under us.  To guard
against this we need to close our type system by imposing a ClassMeta which
closes off mixin types that do not inherit from one of the special case base
classes we have defined.  

The API implications should be small.  There was never a functional case where 
extending a Java object within Python actually made sense as the Python portion 
is just lost when passed to Java and unlike Proxies there is no way to retrieve
it.  Further the extending a Java object within Python does not bind the
lifespan of the objects so any code that used this is likely already buggy.  We
will properly support this option with `@JExtends` at a latter point.

With this limitiation in mind, this appears to be the best implementation
 - It adds the slot to all of the required types.
 - The slot is immediately accessable using just two fields (basesize, itemsize)
 - The slot can be tested for easily (check tp_alloc, tp_finalize)
 - It closes the type system by forcing a meta class that guards against
   inappropraite class constuction.

We could in principle add the slot to the "front" of the Python object but that
could cause additional issues as we would require also require overriding the
deallocation slot to disappear our memory from the free.  This is dangerous
because any module that mistakenly called the Python generic object base free
rather than the slot free would cause Python to crash.  Of course that should
already be a problem for the module with the bug, but as it would be a correct
assumption in 99% of the cases it may not be exposed.  Thus in our opinion
adding to the tail is safer than adding to the head.

Speed patch implications
------------------------

Other than improving the speed, the speed patch has a lot of below the hood
changes.  So long as the user was not accessing private members there is no API
change, but everything below that is gone.  All private symbols like
`__javaclass__` and `__javavalue__` as well as all exposed private members
vanish from the interface.  There is no longer a distinction between class
wrappers and `java.lang.Class` instances for purposes of the casting system.
The wrapper is an extension of a Python type and has the class methods and 
fields, and the instance is an extension of a Python object without these.
Both hold Java slots to the same object.  Therefore a lot of complexity of the
private portions is effectively removed from the user view.  Every path now has
the same resolution, check the Java slot first and if not assume it is Python.

Two private methods now appear on the wrapper (though I may be able to hide them
from the user view.)  These are the test entry points `_canConvertToJava` and 
`_convertToJava`.  Thus the speed patch should be transparent all user code that
does not access our private members.  That said some code like the Database
wrappers using JPype have roots in some code that did make access to the private
tables.  I have sent corrections when we upgraded to 0.7 series thus making them
conforming enough not to touch the private members.  But that does mean some
modules may be making such accesses out in the wild.

The good new is after the speed patch pretty much everything that is supposed to
be under the hood is now out of the reach of the user.  Thus the chance of
future API breakage is much lower.

Below the hood changes
======================
 - We have tried to prevent backend changes from reaching the API, though this
   is not always entirely the case.  The majority of the cases not already noted
   appear to be in the range of implementation side effects.  The old
   implementation had bugs that create undefined behaviors like reinterpet
   casting buffers and the like.  It is not possible to both fix the bugs in the
   backend and make preserve buggy behavours on the front end.  We have limited
   our changes to only those for which we can see no desirable use existing.
   Calling a list slice assignment on a float from a numpy array of ints and
   getting a pile of gibberish was as far as we can tell not useful to the user.
 - setResource is dropped in favor of caching the module dictionary once at
   start of the JVM.  We have a lot of resources we will need and the
   setResources method was cumbersome.
 - There is a lot of thrashing for the Python module style between C and C++
   style.  The determining blow was that C++ exception warning showed up when
   the proper linkage was given.  Thus the perferred style flipped from C++
   style to C.  Thus the naming style change accordingly.
 - In addition to the style change there is also an attempt to isolate symbols
   between the different classes.  The older style with a formal header that
   declares all the symbols at the top encouraged access to the functions and
   increased the complexity.  Moving to a C style and making everything static
   forces the classes to be much more independent.
 - With the change to C style there is natural split in CPython class files
   between the structure declaration, static methods that implement Python API
   functions, the declaration of the type, and the exposed C++ style API used by
   the rest of the module. 
 - There is some thrashing on how much of the C++ wrapper style Python API to
   keep.  The rewrapping of the API was mainly so support differences between
   Python 2.7 and 3.x.  So we dropped where we could.  Only the JPPyObject which
   acts as memory handling device over pure Python style (because Python style
   is not exception safe) is strongly needed.
 - There is some spacing thrashing between different editors and the continuing
   debate of why C was written to have the pointer stick to to the variable
   rather than the type.  When writing `Object* foo` it implies that the star
   is stuck to type rather than the variable where C reads it as the opposite.
   Hence there is the endless churn between what is correct `Object *foo` and
   what we would say in which `Object*` is actually a type.  As Python favors
   the former and we currently have the latter that means at some point we
   should just have formatter force consistency.
 - We introduced Py_IsInstanceSingle.  Is is a missing Python API function first
   for fast type check method of a single inherited type using the mro.  Then
   something is singlely inherited we can bypass the whole pile of redirects and
   list searchs and just compare if the mro list matches at the same point
   counted from the end.  As all of our base types are single inherited this
   saves time and dodges the more expensive calls that would trigger due to the
   PyJPClassMeta overrides.
 - We introduced Py_GetAttrDescriptor.  This was previously implemented in
   Python and was very slow.  Python has very good implementations for GetAttr
   but unfortunately it has the behavior that it always dereferences
   descriptors.  Thus it is useless in the case that we need to get the
   descriptor from the type tree.  We have reimplemented the Python type search
   without the descriptor dereferencing behavior.  It is not nearly as complete
   as the Python method as we only need to consult the dictionary and not handle
   every edge case.  That of course means that we are much faster.
 - As with every rewrite there is the need to cleanup.  Anything that wasn't
   reached by the test bench or was identified as being called only from one
   location was removed or reencorperated back in into the function that call
   it.

Fixes #575 
Fixes #573 
Fixes #567 
Fixes #564 
Fixes #563 
Fixes #542 
Fixes #536
Fixes #528


Thrameos:
Fixes #562

marscher:
Wow, this is very detailed. I need time to review this. Thanks for all the efforts.

jpype/nio.py
```diff
@@ -20,11 +20,6 @@
 __all__ = ['convertToDirectBuffer']
 
 
-
-
-
-
-
 # FIXME this is not documented.
 # FIXME there are no tests of this method
```
marscher: are these two FIXMEs still true?

native/common/include/jp_array.h
```diff
@@ -25,13 +25,34 @@ struct _object;
 typedef _object PyObject;
 #endif
 
+class JPArray;
+
+class JPArrayView
+{
+public:
+	JPArrayView(JPArray* array);
```
marscher: Wouldn't the usage of a smart pointer protect us from releasing the JPArray too early? I guess you have dealt with this problem elsewhere.

native/common/include/jpype.h
```diff
@@ -12,16 +12,15 @@
...
 #ifndef _JPYPE_H_
-#define _JPYPE_H_ 
-
-// Define this to generate the trace calls
+#define _JPYPE_H_
 
-// Define this to make the trace calls do their output. 
-//    use "setup.py --enable-tracing build" to enable 
-//#define JP_TRACING_ENABLE
+#ifdef __GNUC__ // Then this must mean a variant of GCC on win32 ...
```
marscher: This comment seems wrong here, since it is outside of the WIN32 definition check.

native/common/include/jpype.h
```diff
@@ -101,11 +88,16 @@ using std::list;
 /** Definition of commonly used template types */
 typedef vector<string> StringVector;
 
+typedef jvalue (*jconverter)(char*) ;
+
+extern jconverter getConverter(const char* from, int itemsize, const char* to);
```
marscher: So the type conversion for an array happens within a Java class? I think this deserves some comment to be easily understandable.

native/common/include/jpype.h
```diff
@@ -166,4 +169,42 @@ class JPField;
 #include "jp_proxy.h"
 #include "jp_monitor.h"
 
+template <typename array_t,  typename ptr_t>
```
marscher: Since you love whitespace changes, please delete this double space :)

marscher: :+1: for the template pattern. Shouldn't this go to a separate header, which is specialized to arrays? If you object, please do not justify it - I trust that it you put it there for a good reason.

native/common/jp_array.cpp
```diff
@@ -128,8 +123,46 @@ JPPyObject JPArray::getItem(jsize ndx)
...
+
+void JPArrayView::reference()
+{
+	refcount++;
+}
+
+bool JPArrayView::unreference()
+{
+	refcount--;
+	array->getClass()->getComponentType()->releaseView(*this, refcount == 0);
```
marscher: shouldn't the refcount check happen here, rathen than demangling the type and doing a null-op, if refcount is still positive?

native/common/jp_arrayclass.cpp
```diff
@@ -207,10 +203,38 @@ jvalue JPArrayClass::convertToJavaVector(JPPyObjectVector& refs, jsize start, js
...
+JPValue JPArrayClass::newInstance(JPJavaFrame& frame, JPPyObjectVector& args)
+{
+	JP_TRACE_IN("JPArrayClass::newInstance");
+	if (args.size() != 1)
+		JP_RAISE(PyExc_TypeError, "Arrays require one argument");
+
+	if (PySequence_Check(args[0]) == 1)
+	{
+		JP_TRACE("Sequence");
+		Py_ssize_t sz = PySequence_Size(args[0]);
```
marscher: Is there a type check for args[0] happening in PySequence_Size?

native/common/jp_booleantype.cpp
```diff
@@ -218,3 +243,21 @@ void JPBooleanType::setArrayItem(JPJavaFrame& frame, jarray a, jsize ndx, PyObje
...
+	JPJavaFrame frame;
+	view.memory = (void*) frame.GetBooleanArrayElements(
+			(jbooleanArray) view.array->getJava(), &view.isCopy);
+	view.buffer.format = "?";
+	view.buffer.itemsize = sizeof (jboolean);
+}
+
+void JPBooleanType::releaseView(JPArrayView& view, bool complete)
+{
+	JPJavaFrame frame;
```
marscher: you need the guard only if complete holds, right?

native/common/jp_class.cpp
```diff
@@ -461,7 +467,7 @@ void JPClass::postLoad()
 
 void JPClass::loadFields()
 {
-	JPJavaFrame frame(32);
+	JPJavaFrame frame(512);
```
marscher: 512 fields/methods in one class? Is that a reasonable number or was it popping out of some tuning experiment?

native/common/jp_convert.cpp
```diff
@@ -0,0 +1,644 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <jpype.h>
+
+static jvalue bool2bool(char* c)
```
marscher: You could also use an anonymous namespace to avoid declare every function statically. This would look more like c++ than good old c.

native/common/jp_convert.cpp
```diff
@@ -0,0 +1,644 @@
...
+{
+	jvalue v;
+	v.z = (*(uint8_t*) c) == 0 ? false : true;
```
marscher: Why do you need to cast to a pointer first, and then dereference it? Shouldn't you use c++ like casts so it can be compile time expressions?

native/common/jp_encoding.cpp
```diff
@@ -55,7 +55,7 @@ struct membuf : std::streambuf
 	{
 		this->setg(begin, begin, end);
 	}
-};
+} ;
```
marscher: :+1:

project/jpype_cpython/Makefile.linux
```diff
@@ -0,0 +1,68 @@
+PYVER = 3.8
```
marscher: When did we introduce Cython bindings in JPype? I can not use this file on my machine. What is this makefile used for?

marscher:
The following code seems to create a copy, which it shouldn't in my opinion.

``` python
import jpype
import numpy as np 
jpype.startJVM()
n = int(1e3)
x = jpype.JArray(jpype.JDouble, 1)(n)
view = np.array(x[:])
assert view[0] == x[0] == 0.0
view[0] = 1
assert x[0] == 1
```

The last assertion fails. So there is a copy made instead of viewing the memory inside the JVM. The old (and hacky) Numpy related code allowed to avoid copying memory around. Is there a way to steer this behaviour?

jpype/nio.py
Thrameos: Those fixmes have been addressed.  I will remove them.

native/common/jp_array.cpp
Thrameos: The code was supposed to call a commit with each dereference.  The refcounts represent the number of active buffers.  There is a difficulty that java does not allow write through, thus no changes actually commit until the view is deleted.  For most transfers this did not matter.  But memoyview which holds it open made a huge deal.   Thus the best i could achieve was to try to commit with each close hear.  

I personally think this is too awkward to be usable as the user can only control if the only have one open.  Someone may try to tell numpy to use an array on the memorybuf which would not write through until the last copy dies.  

The alternative is to force read only.  But as most buffer transactions request read and write even if the only need read this ended up not working well.

native/common/include/jp_array.h
Thrameos: Arrayviews must be reference counted or independently created for each request.   I tried the shared which lead to this implementation.   There is a lot of awkwardness here as the reference counters for views and python interact.  Originally I made views embedded in the JPArray itself which other than space is more correct.  I was going to go with that solution, but the future pull request to support the numpy `__array__` is going to require creating a view object with a lot of substance as it has to open all of the views at once.  

I am thus far trying to keep as little C++ in the Python layer as possible as I am already running into serious issues with the fact that CPython is dealing with them pretty close to directly.  If you would like me to try to use smart pointers I can, but the pointer is going to get held by a C struct in PyJPArray.

native/common/include/jpype.h
Thrameos: I will look this over.  I was mainly deleting old stuff and after the deletion the remaining was left like this.  Perhaps the editor ate something.

native/common/include/jpype.h
Thrameos: When making a buffer transfer from one buffered type to another (numpy to Java array) where the format code does not match I have to get a function which does the conversion.  I originally had this split out by the type requested and such but that was really awkward.  Thus I just created one giant autogenerated file that returns every combination that we could need.  Then the getConverter looks up the function and passes it back so we can use it over and over for each item in the list.

This is a big enhancement over the reference code in CPython that I was looking at.  For CPython implementation they do a giant switch statement with a comment (paraphrasing) "Don't touch this or it will have huge speed changes as gcc does a lookup table".  As they go through the table for each item to convert I am not sure why they don't just allocate a function for each conversion and then go through the switch once like I did.

native/common/include/jpype.h
Thrameos: I am planning a mass white space pull request, but for now I am in the tenuous position that I have backported 1/3 of JPype 0.8 into 0.7.  I am going to need to forward port the stuff that I did here back into 0.8 so I can finish up that development.  Thus I have been trying (without much success) to convince the editor to stop doing stray whitespace changes.

The template pattern can be separate if you would like.    This was the lone survivor of primitive header massacre that started that started the numpy removal process.

native/common/jp_booleantype.cpp
Thrameos: The correct behavior is to do a release with commit (1) when write buffer holds and either an abort (2) or close (0) when the last view is written, but JNI_CHECK issues a segfault on the second commit (1).   I tried a number of experiments and honestly I don't get what JNI_CHECK.  The documentation in JNI specifically says COMMIT does not free the buffer but apparently that doesn't mean you can write to it some more and then release again as JNI has already washed its hands.     Perhaps I should just bite the bullet and refuse to share the buffers, but that created a really wacky side effect that if you get two memory views and wrote to the first and closed the first and the second the changes the first did get erased even if the second makes no changes.  

 CPython does not seem to request read only access when it should so the release is copying bytes when it shouldn't.  Hence a decision point, should we just always say we are read only and then file bug reports for users that request write access when they are only doing read ops.  (Or I am completely misreading the CPython API.)

native/common/jp_class.cpp
Thrameos: It was to avoid the JNI_CHECK warning that I was getting in testing.  I just doubled until it went away.   There is a minor structural problem here.  You can't get the size of the frame (which is the total number of members to be converted) until you have constructed a frame, but once you construct the frame the items are selected.  It does autogrow the frame if you exceed.

In this case each member, field, and array access is generating one reference.  For some of my test cases I was hitting 256 which produced a warning, thus I took it to the next power of 2.    This section is all going to be removed when I convert to having Java control the type manager as Java will send over the data with the correctly sized frame.

native/common/jp_encoding.cpp
Thrameos: Yep, this is the wonders of autoformatting using netbeans as the editor on C++.  It has tied spacing for one common feature to spaces for something annoying like this.  If I tell it no spaces after a close brace, then it does something like remove all spaces in operator sequences.   One checkbox does both behaviors. Thus I had to accept bad things like this to make common case acceptable.

project/jpype_cpython/Makefile.linux
Thrameos: The makefile is so that my sorry laptop can do incremental builds during development.  It takes 10-15 minutes to recompile because setup.py does not support incremental builds anymore.  It is fine for compiling the whole thing, but when I am doing incremental work it is beyond annoying to have to wait 10 minutes to discover a typo in pyjp_value.cpp.  Thus like everything in project dir this is just a tool to speed up development.

native/common/jp_convert.cpp
Thrameos: This was selected because I was arguing with the editor.  The editor has one setting to indent namespaces which it interprets as indenting namepaces and all `extern "C" {` sections.  Thus I either got indents on every sections of C linked code (annoying) and namespaces (which we were using) or not which was creating pointless whitespace diffs.  

Thus in this case I just selected "static" rather than anonymous namespace to avoid the fight.   Personally I would like to see this sort of utility function in CPython rather than re-implementing it here.

native/common/jp_convert.cpp
Thrameos: Not sure if there is a performance difference.  I can recode if you feel strongly.    This is really just about doing reinterpret cast over and over.  It doesn't matter at all for bool because I going to force true or false, but for others  I am forcing signed to unsigned conversions.

I was planning a big test bench that exercised each of the type conversions, but decided to request a review as we have a bunch of people stuck over the numpy extra problem, so  decided to rush this out and then search for edge cases once we had agreement on direction..

project/jpype_cpython/Makefile.linux
Thrameos: If you can figure out how to get the makefile to consult python setup.py and get the correct flags it would be great so that this could be used universally.   It is annoying that I have to run setup.py then copy down the flags for the build and put it in the makefile each time I switch Python versions / architectures during testing, but that is the best solution I have currently.

Thrameos:
The behavior on your test is unchanged from the previous version.   It shows the same assertion error in both versions.  (But for different reasons)

In either case JNI will make copies (and likely more than you are thinking.)

Analyzing the old code.
```
import jpype
import numpy as np 
jpype.startJVM()
n = int(1e3)
x = jpype.JArray(jpype.JDouble, 1)(n)
view = np.array(x[:])
```
Here we generate 2 copies of the data typically on the `[:]`.
1) We call GetPrimitiveArrayCritical, JNI copies the memory from the  Java space to the transfer space.
2) We then call memcpy to copy the transfer buffer into numpy.
3) We then release critical which may copy back again if commit was set.

The first one is supposed to be bypassable, if the JVM supports pinning of arrays, but thus far I haven't seen any cases where it did. (Perhaps it is size triggered).  

We should have used GetArrayRegion instead which would have saved a copy as we could have gone straight from Java to numpy array backing.
```
assert view[0] == x[0] == 0.0
view[0] = 1
```
So  at this point the numpy copy is set to 1 but the Java array is still 0.
```
assert x[0] == 1
```

Analyzing the new code.
```
import jpype
import numpy as np 
jpype.startJVM()
n = int(1e3)
x = jpype.JArray(jpype.JDouble, 1)(n)
view = np.array(x[:])
```
The slice operator does nothing here as we just get a view of the same array.

np.array(v) then calls for a memory buffer view of the array which produces a pair of Get/Release calls.  This is slightly worse than before as we can't use Critical because we do not know how long we will hold the array.  Same result on average which is two copies get triggered.

The flag to the buffer makes a difference here.  If they specify "WRITE" then we will trigger a 3rd copy to push the transfer area back to the Java memory.  I can do some auditing and see what triggers the copy.  The four relevant cases are `bytes`, `bytearray`, `memoryview, and `np.array` which all call for a Py_buffer to be created.

```
assert view[0] == x[0] == 0.0
view[0] = 1
```
So  at this point the numpy copy is set to 1 but the Java array is still 0.
```
assert x[0] == 1 <== Fails because np.array requested a copy.
```

In general as far as I can tell in Python when you call a class ctor you expect a copy to be produced.

For example consider the following code
```
import numpy as np
n = int(1e3)
x = np.zeros(n)
view = np.array(x[:])  <== User explicitly is requesting a new array that is a copy of the old
assert view[0] == x[0] == 0.0
view[0] = 1
assert x[0] == 1
```

And we get the same assertion failed.



There is one huge difference though between the implementations...

Try this code on the old version and the new
```
import jpype
jpype.startJVM()
n = int(1e3)
x = jpype.JArray(jpype.JDouble)(n)
view = x[:]  <== This should do nothing because we are just saying v is a slice of x
assert view[0] == x[0] == 0.0
view[0] = 1
assert x[0] == 1  <== old fails assertion as we made a copy we should not have.  New code passes.
```

CPython lacks a call to get a buffer to request direct transfers of bytes but instead favors request access.  Thus I can't make use of the JNI short cuts for getting regions.  Thus I am forced to use calls that provide for full access which are less efficient.

Thrameos:
By the way thanks for the quick efforts to get the review started.  This pull request is much larger than I was originally hoping as I needed to make use of the direct access to CPython types that I implemented in JPype 0.8.  I expect that you will need some level of comments (and possible revision) on the level of memory hacking that I required on the speed patch.  I plan to take then notes posted on the pull request when I update the devel guide.   At times it felt like I was being forced to copy whole sections of CPython onto a piece of paper as psuedo code and then recomposing them back because CPython lacks simple things like `PyType_FromSpecWithBasesAndMeta`.  

Unfortunately there will be implementation implications that leak from this.  Java JNI is really sloppy about supporting pinning of arrays and I can't guarantee that even with the current efforts that I have covered all the cases.  Sometimes it gives a copy and you can do whatever you want until a commit is forced.  Sometimes it gives direct access which means anything goes.  If you open more than one view to the same array (by casting to an object and back so the views don't share) you can get arguments where the last one to commit rules.   This results in frustrating and sometimes really hard to explain behavior.   I had to explicitly create views with copies and then use del to remove references to actually get it to pretend to write through.    Worse the copy can fail and I am not sure I have accounted for this.

If you want me to try to force the views to all be read only (and thus save us from explaining how JNI functions) I would completely agree to it as it was beyond counter intuitive to me and I know this stuff pretty well.

native/common/include/jpype.h
Thrameos: Actually it appears that it was just me cutting and pasting to try to disable a warning.  I will clean this section up.  Near as I can tell most of it is cruft.

Thrameos:
rebased and squash to allow for proper coverage check.

native/common/jp_convert.cpp
Thrameos: I updated the pattern to more C++ like

native/common/jp_convert.cpp
Thrameos: Fixed.

native/common/jp_class.cpp
Thrameos: Fixed.

native/common/include/jpype.h
Thrameos: addressed.

native/common/include/jpype.h
Thrameos: Done.

Thrameos:
I disabled the ability to write back to Java memory with memory buffers for now.  It was so counter intuitive that no one is likely to every use it.

Thrameos:
@marscher I hereby swear that this pull request is actually ready to go despite my two days of vane attempts to increase the coverage score.  I did turn up 3 segfaults on edge cases so it wasn't wasted effort. Unfortunately 50% of the C++ code is aggressive checking for failures that i can't seem to exercise without creating testing hooks in the code.

native/common/include/jp_array.h
marscher: Thanks for the explanation. I can imagine that this is really tough. I think you have the right strategy here.

native/common/jp_array.cpp
marscher: The old impl commited after each write request (which implies releasing the actual jni view, which may cause the jvm array to be freed). So your pattern protect against this. It also means that we can leak JVM memory, if the Python view is somehow never released. But seems much more stable. I would not enforce read-only, it is too restrictive.

native/common/jp_booleantype.cpp
marscher: Does it segfault on OpenJVM? We could try to ask over there about implementation details. I honestly don't know. But I guess we have to commit every change once it is finished writing to actual buffer. 

The hassle with CPython has been dealt by the usage of NumPy, which wraps around some hassles of it. I'm again sorry, but I do not know the details here. Eventually you can look up some things in the C++ pybind11 project, since they have a well established NumPy layer (also dealing with memoryviews, I'd guess).

native/common/jp_class.cpp
marscher: The final solution paths sounds awesome!

native/common/jp_convert.cpp
marscher: I don't see why these util functions would be desired in CPython. I like the templated version of jp_converter much better!

native/common/jp_encoding.cpp
marscher: just wondering which IDE would do such things :)

project/jpype_cpython/Makefile.linux
marscher: My solution to the long build time was using ccache. I do not know if there is such a thing as a distutils/setuptools makefile wrapper, but it just sounds like something which would be inherently broken.

native/common/jp_booleantype.cpp
Thrameos: Yes.  I am testing primarily on open JDK 11 using ubuntu on WLS and virtual image, with alternative testing on Oracle under windows.   Both show the same behavior.  I am guessing that my interpretation on "JNI_COMMIT" is incorrect.  It appears that it is just allowing you to keep the memory that was passed rather than allowing for a second commit.  But that seems like a very odd feature.

Because we can't commit more than once, we would get very odd behavior if we write back more than one behavior.  thus I recommend disabling the write until I can resolve this.  
I came up with several odd cases on write testing that leads me to think it would be a misfeature because you can defeat it simply by passing to another array that opens a buffer (such as asking for numpy to share an array.)  The safest course is to just issue readonly buffer as we never supported write through before.

If the user does actually want a write through they can use a direct buffer or the shared memory solution which are much more stablely supported and do not require pinning support.

native/common/jp_array.cpp
Thrameos: The previous implementation was a copy once on request solution so by definition it was write through.  It did have the downside that readonly was not enforced thus you could potentially alter the array copy using the double slice operator but it was never reflected in Java so it leaves the impression that work was done that wasn't.

``` 
ja=JArray(JInt)([1,2,3,4)
b = ja[:]
b[1] = 5
print(ja)  <== produces [1,2,3,4]
```

The new implementation writes through because the slice is actually a view.  But it wont permit writing through of the memory view because we can't pin the array.  Hence you may be confused between the writing of the slice view and writing of the byte buffer view than backs it.  We can support writing the slice just fine now, but byte buffer which requires creating a memoryview, accessing the memory, then deleting the view is problematic.

Thrameos: Sorry I edited the last comment... should have read "read only" not "write through"

native/common/jp_class.cpp
Thrameos: For now I updated the code to delete local references after each use when traversing the array.  It is a bit slower but it keeps references in control.  We have a few spots that required that treatment.  Just grep for DeleteLocalRef to see all of the places that require it.

project/jpype_cpython/Makefile.linux
Thrameos: I have seen some hacks that make use of ccache, but none worked well for me.  I plan to just hack a copy of setup.py to make it dump out the flags and such into a templated Makefile so that I don't have to keep hand editing.  For now the canned make file is good enough for my developement.  Unfortunately after I lost it for the second time I decided I had to push it into git as it takes an hour or so to recompose every time it gets lost.

I will remove it once I get the setup pattern working,

native/common/jp_encoding.cpp
Thrameos: Netbeans is not a very good C++ GUI but for JPype 0.8 where I have much of the backend code in Java I need to keep bookmarks set between the two code bases it is useful.  I just wish they would allow a drop in for a  more genera in formatter so that I don't have to make horrible decisions.

native/common/jp_convert.cpp
Thrameos: Well the CPython native solution of sending things through a giant switch statement every single time that they want to convert each element is rather poor.  I was hoping to just call their type conversion routine rather than spinning my own.  (Done while watching my sons wrestling meet in the stands.)  But there isn't a lot of portable stuff in CPython in terms of support libraries.  They bury everything as statics even when it is something needed.  That is why you see me write likes like 

```
   PyLong_Type.tp_str(obj)
```

I can't call the actually method directly so I am forced to find someplace in the library where I can guarantee that the virtual table points to something I need to access and call it.  Heaven help us if someone monkey patches it.

Thrameos:
I will put in a followup pull request with the multidimensional array memory buffer support.  It appears like it is a very big speed up once I came up with a way to get most of the memory transfer performed on the Java side.   I may be able to speed it up more with a direct byte buffer but it should be enough to close #133.

marscher:
I have actually forgotten how to do the direct memory transfers on the Python side, but I'm sure it was possible. I just hacked this feature back then, since I needed to avoid memcpy calls at any cost, because the data would blow main memory easily. The depended project is dead for a long time, so I also personally believe nobody really relies upon it.

I think we can merge this now.

Thrameos:
Merged so that I can resume coverage checks (without a massive patch sitting cluttering everything up).   We have just a few outstanding patch requests that to get through and we can consider a release.

PR #582, opened 2020-02-17
Multidimensional memory buffer support

### LINKED ISSUE #133: Reading from Java byte array with np.frombuffer

Thrameos:
This pull request builds on the previous speed patch/memory buffer to improve the transfer of memory from Java to Python.   It uses a Java side support routine to create an arrays of 1D primitives that can be copied into a C continuous memory buffer as well as handling the bookkeeping of determining if the Java array is suitable for transfer (must be rectangular primitives no more than 5 dimensions).  The speed up is substantial though still slower than directly copying a numpy array.  This is to be reviewed after the main pull request review is complete.

Fixes #133


marscher:
With rectangular shape you mean shapes of the form (s_1,..., s_5), whereas s_i is always the same number? OK, square matrices are quite common use cases, but would it be difficult to support the n x m case?

Thrameos:
Both numpy.array and the Python buffer API do no support transfers of  irregular (called jagged arrays)  because such an array does not have a regular size.  Here is some test code to show what happens

```
a=np.array([[1,2,3],[4,5,6]])
print(a.dtype)
a=np.array([[1,2,3],[5,6],[7]])
print(a.dtype)
```

The first is a rectangular array and gets marked as contiguous with dtype `int64`  while the second is jagged and instead returns `object`.   Oddly numpy array does support a memoryview for the second as we can see for the next test case.

```
a=np.array([[1,2,3],[5,6],[7]])
print(a.dtype)
mv = memoryview(a)
print(mv.format)
print(mv.shape)
```

It gives type "O" with shape "(3,)"  so it really is just handling the first dimension.  I could not find any documentation on how to support Object backed memory views, but if I can then I can implement this behavior as well and we would be "complete".   Thus far I have only been dealing with primitives but there is no reason (unless the API is really broken) I can't do the objects as well.

I am a bit perplexed how numpy is getting type O out of the view.  It must be going through a fallback case.   On the other hand I have already found numpy doing things like sending an itemsize that does not match the type code, so it doesn't surprise me that there is undocumented stuff with this API.  See the reference code from 3.8 below

```
Py_LOCAL_INLINE(Py_ssize_t)
get_native_fmtchar(char *result, const char *fmt)
{
    Py_ssize_t size = -1;

    if (fmt[0] == '@') fmt++;

    switch (fmt[0]) {
    case 'c': case 'b': case 'B': size = sizeof(char); break;
    case 'h': case 'H': size = sizeof(short); break;
    case 'i': case 'I': size = sizeof(int); break;
    case 'l': case 'L': size = sizeof(long); break;
    #ifdef HAVE_LONG_LONG
    case 'q': case 'Q': size = sizeof(PY_LONG_LONG); break;
    #endif
    case 'n': case 'N': size = sizeof(Py_ssize_t); break;
    case 'f': size = sizeof(float); break;
    case 'd': size = sizeof(double); break;
    #ifdef HAVE_C99_BOOL
    case '?': size = sizeof(_Bool); break;
    #else
    case '?': size = sizeof(char); break;
    #endif
    case 'P': size = sizeof(void *); break;
    }

    if (size > 0 && fmt[1] == '\0') {
        *result = fmt[0];
        return size;
    }

    return -1;
}
```

marscher:
It seems to me like we are now in the phase of re-implementing lots of numpys features. Lets see how this will proceed, I have no doubt you will push this into the right direction.

Thrameos:
My next plan in this direction is to implement the fast transfer of memory from byte buffers back into Java.  This would take two forms.  

For transfers with a specific size (2d->2d) it is just a special transfer on the initializer `contents = np.array(....);  JArray(JDouble, dims)(contents)`.  But requiring the dims to match really is just extra cruft so I am thinking something like `JArray.of(contents, dtype=JDouble)`   may be a bit more useful in general so that the dims don't need to be specified explicitly.

PR #583, opened 2020-02-18
Collection fix

### LINKED ISSUE #578: java.util.Container del does not work
`__delitem__` is tied to remove but those mean very different things.

```
jcollection = JObject(java.lang.ArrayList([11,12,13,14], java.lang.Collection)
del jcollection[13]
print(jcollection)
```

gives `[11,12,14]` because we remove item 13 not the non-existant item 13.

I will remove the bad binding as this is not likely what is intended.

### LINKED ISSUE #579: java.util.List getitem throws a Java exception
The following code fails

```
jlist=java.util.List()
jlist.addAll([1,2,3,4])
try:
   jlist[6]
except IndexError:
   print("success")
```

The wrapper returns java.lang.IndexOutOfBoundsException which is unexpected for a Python operation.  We can fix this by making IndexOutOfBoundsException derive from IndexError internally.

### LINKED ISSUE #580: java.util.HashMap does not throw KeyError on bad item
The wrapper for HashMap does not produce the expected output on a bad item like Python.

```
hm = jpype.java.util.HashMap()
try:
    print(hm["foo"])
    print("fail")
except KeyError:
    print("success")
```

We should add a check for containsKey() if the map returned None so we can throw properly.  The user can call `get` if they want Java behavior.

### LINKED ISSUE #587: HashMap node does not unpack when used as iterator
We need a customizer on HashMap.Node to allow for unpacking when iterating elements.

Here is an example
```
import jpype
jpype.startJVM()
java = jpype.java
hm = java.util.HashMap()
hm['alice'] = 'pickles'
hm['betty'] = 'ice cream'
hm['catty'] = 'corn'
for p,v in hm.entrySet():
    print(p, v)
```

Gives 
```
  File "testMap2.py", line 8, in <module>
    for p,v in hm.entrySet():
TypeError: cannot unpack non-iterable java.util.HashMap.Node object
```

Thrameos:
Addressing issues found in review of collections.

- There are certain wrappers methods such as `__getitem__` in which the Java method we call can throw an expected exception which would not be caught properly by normal Python code.  To prevent this I have enabled the ability to add a Python exception to the Java exception tree so that a Java exception can be caught as a regular Python exception.  Obviously this has limitations because the Python and Java exception trees are different and thus overloading RuntimeException to RuntimeError would cause problems.  But in limited cases like java.lang.IndexOutOfBoundsException to IndexError this makes sense.
- One of the untested collection wrappers was calling remove() which was not by index but by item which gave strange results.  This has been removed.
- Java hashmaps do not distinguish between a missing key and a null value so there was a big difference when dealing with maps.  I added KeyError in that case.  The user can get the Java behavior by calling ``get()`` as expected of the Java Map.
- Again more tests for our hopeless matrix.

Fixes #578 
Fixes #579 
Fixes #580
Fixes #587


Thrameos:
If you don't like one of these changes I can resubmit.   It may break some code that was using buggy behaviors.

marscher:
Thanks for addressing all of these issues and increasing testing coverage. Untested code is broken code. That's why you deleted the remove method, but it will break existing code with a AttributeErrror. Please raise a NotImplementedError containing info how to change the code to be accepted. The alternative would be to find a solution for remove(e).

Thrameos:
I dont think the collection remove could be exercised as it was covered up by list which had a correct implementation.  The only way to hit it was cast or implement you own collection.  Typing `del ja["foo"]` and expecting it to remove the string foo would be strange.  I will check to see if the was a case that was not covered by the parents.

Thrameos:
@marscher I tested all of the classes that were coming off of collection wrapper.  It appears that only one wrapper was effected by the remove change.  I simply slide the wrapper down though it does cause a small difference in Python and Java.

For maps our behavior matches for single key removal.  We had an overridden ``__delitem__`` already.
```
hm = {}
hm['alice'] = 'alice'
hm['betty'] = 'betty'
hm['catty'] = 'catty'
del hm['betty']
print(hm.keys())

hm = java.util.HashMap()
hm['alice'] = 'alice'
hm['betty'] = 'betty'
hm['catty'] = 'catty'
del hm['betty']
print(hm.keySet())
```

For lists we already supported ``__delitem__`` so no changes there.  We also supported removal with slice so our coverage was good here.
```
ls = []
ls.append('alice')
ls.append('betty')
ls.append('catty')
del ls[1]
print(ls)

ls = java.util.ArrayList()
ls.add('alice')
ls.add('betty')
ls.add('catty')
del ls[1]
print(ls)
```

The only place that the bad wrapper leaked through was for Set.
```
try:
    st = set()
    st.add('alice')
    st.add('betty')
    st.add('catty')
    del st['betty']
    print(ls)
except Exception as ex:
    print("Not support in Python", type(ex))
    print(ex)

st = java.util.HashSet()
st.add('alice')
st.add('betty')
st.add('catty')
del st['betty']
print(st)
```

Python does not support ``del st['key']`` form though I do not understand the reason for this.  Thus we have two options here.  We can support the operation for Java types (like the original wrapper) or force it to a TypeError with the same message as Python.  

I have changed the pull request to be my best compromise between the two.  I have implemented a TypeError with the appropriate message for the case in which someone managed to use a custom container type (or something we never test like Queue), and I make is so that HashSet has the "extra" operation.  I can remove the extra operation if you feel it is unnecessary.

Thrameos:
@marscher With this pull request I think we should likely start a release process.  I will be having many more minor changes like this as I increase coverage, but the process will likely take a month or more.  Thus given the current broken state with the numpy I recommend that we release now.  

We are in a slightly awkward place as we are basically releasing the JPype 0.8 front end without the JPype 0.8 core. It is missing the new type conversion API that lets users define custom conversions which allows date, sql timestamp, paths, and even lists to integrate seamlessly between Java and Python without having to customized each method.  

I am going to leave the labeling to you on that front.  If we release this as 0.7.2 then 0.8 will be much more incremental of a release which is likely okay.

Thrameos:
Fixes #587

Thrameos:
@marscher any objections to merging this and starting a release?

marscher:
No, please go ahead. Thanks

Am 28. Februar 2020 15:11:07 MEZ schrieb Karl Nelson <notifications@github.com>:
>@marscher any objections to merging this and starting a release?
>
>-- 
>https://github.com/jpype-project/jpype/pull/583#issuecomment-592529426

-- 

PR #588, opened 2020-02-24
Link Python and Java GC 

Thrameos:
(out of date,  need to be freshened)
This is an experimental PR designed to tie the Java and Python GC together.  It still requires additional tuning and a resolution as to why it causes shutdown to pause for a prolonged period.

@pelson This is my first hack at solving the problem.  

I used a few strategies to try to suppress the aggressive Python GC.   
1) Don't force a GC until we exceed what is expected from just running a normal Java process.
2) Once the process is over the limit wait until the memory grows.  (Thus will slowly climb the memory as we will trigger this with larger and larger amounts of freed memory)
3) Record the time to hit the growth cap and then predicatively call  the GC before we would be forced to.
4) If we hit a hard cap then be more aggressive about it.

There are still some minor issues before this is ready for prime time and some tuning required, but lets first see how bad this busts the test matrix.

### DIFF
--- a/native/common/include/jpype.h
+++ b/native/common/include/jpype.h
@@ -152,4 +152,20 @@ class JPField;
 // Services
 #include "jp_proxy.h"
 
+namespace JPGarbageCollection
+{
+void init();
+
+/**
+ * Called when Python starts it Garbage collector
+ */
+void onStart();
+
+/**
+ * Called when Python finishes it Garbage collector
+ */
+void onEnd();
+
+}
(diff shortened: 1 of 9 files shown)


Thrameos:
When this is properly tuned and integrated with classhints, it should allow PyPy to pass a garbage collection check.

Thrameos:
This review will be freshened and rebased.  Please wait to review this request until it can be updated.

PR #589, opened 2020-02-26
Classhints

### LINKED ISSUE #242: Integration between java.lang.Date and python datetime
My coworker requests for plotting time based data an integration between java.lang.Date and datetime such that a java.lang.Date can be used in place of a datetime.  I have implemented this but it depends on the Boxed patch.

### LINKED ISSUE #431: Rewrite Typemanager in Java
Current JPype is holding all its C++ wrappers in a global table living managed by C++.  Neither Java nor Python which are memory managed languages have influence on this global pool other that requesting creation of a new wrapper.  Living deep inside the code these wrappers are hard to access and audit.  Further, they do not share resources with one another.  Thus every class that derives from another must implement yet another copy of all of the methods inherited from the parent even not one of them is overriden.  Further, JPype talks to Java though many small transactions each time it needs something from Java.  This is costly and slow and means JPype has to allocate a huge pool of JNI methods.  This goes against recommendations on how to achieve high performance JNI applications.

An effort to correct this was made in #323.  However, that was tabled in favor of a less heavy lift.  
We should revive this effort for JPype 0.8.

### LINKED ISSUE #464: Support for PEP-0519
Java has two path like objects `java.io.File` and `java.nio.file.Path`.  They should be able to quack like a Python path using `__fspath__` and should autoconvert from a Python path object.  Currently we can do one of these operations using a customizer.  But the other half requires Python base conversion API which is targeted for 0.8.

See
https://www.python.org/dev/peps/pep-0519/

### LINKED ISSUE #485: PyJPClassHints API
So one of our long standing (and actually asked for) feature requests is the ability to control the type conversion process.

This thread will cover the plans for and design behind the new type conversion system.

Thrameos:
This pull request is a back port of the JPype 0.8 classhints system and supporting Java classes.  In this pull request we can define custom conversion methods which can be applied whenever a specific class is requested in Java.   Conversions may be by instance or by attribute.   This reduces the need for class customizers and ensures that methods outside of those we have customized apply the same rules as those that were customized (such as Python list to java.util.List)

Fixes #242 
Fixes #485 
Fixes #464 
Fixes #431


Thrameos:
@marscher This one will have to be reviewed before others because it changes the type system so that  we can define implicit conversions from within Python.  

There are two major changes here.  First the customizer API was largely moved into C++ land so that all of the data stored by Python customizers is now in JPClassHints were it is accessible to C++.    Once that information is inside of C++, we can then switch from a two part to one part conversion method.  In the prior system we called ``canConvertToJava`` which would probe all the types and decide if the conversion was possible then if it was then called ``convertToJava`` which would then probe all the types to see which one to use.   Instead we have unified it into ``getJavaConversion``  which looks it up once and then caches the conversion method.  Thus when it is time to actually do the conversion it only has to call the method.   Without this the classhints API would be a speed impact.

Last this contains the code the moves all the remaining C++ code used to create objects from C++ to Java.  Java has a much better view of object relationships so it can decide which methods override others and what types need to be created in what order.  Thus rather than C++ being in control it just asks Java through the API in ``TypeManager`` to create the required type.  Java decides what to do and creates each of the types using ``TypeFactory`` which is a native methods of C++.
 The end result is we don't need to travel back and forth to Java to probe for information.  Thus all the JPJni stuff shrinks to just a few methods,  And it is now testable from within Java so I can use a Java test bench to debug and resolve issues in type creation.

There is one piece of baggage left over from the experimentation process which is ``JPContext`` which is my attempt at making it possible to start a second Java context.  Currently I have not found any JVMs beyond Java 1.2 that supported that ability.  Our implementation to support multiple contexts is sound, but I doubt we would find a way at the JNI level to support it.  It is a bit of a burden to have to haul around the context from place to place, but otherwise not harmful.   I will take a second short at embedded contexts later in the year using agents instead.

Thrameos:
I should also mention that this likely the last of the core rewrites that JPype will require.  Thus far we have achieved:

1. Removed the extra memory management  (0.7)
2. Removed the extra Host layer for Ruby (0.7)
3. Ported all the types back to C++ for speed (0.8 speed patch)
4. Correct the implementation of buffer handling (0.8 speed patch)
5. Ported all of the JNI management code back to Java (0.8 classhints)
6. Replaced the two part conversion process and add user conversions (0.8 classhints)

Thus we have completed the road map that I put down two years ago for the project.    It hasn't always been pretty as I brutally took apart the backend and reconstructed it bit by bit.   Most things took 2-3 shots to actually find workable patterns,  Priorities shifted to match user requests.  And often I was forced to stand the building back up again part way through the demolishing process just so that some vital piece could go out the door. 

Items that remain on the wish list going forward are (mostly from 0.9 list): 

- [ ]  Wrap Python objects so they can be used from within Java.
- [ ] Provide a reverse bridge that JPype can be used to spawn Python from Java.
- [ ] Embedded Python in Java jars and load it.
- [ ] Allow sub processes to spawn more than one JVM.

I would like to thank @marscher for putting up with me the last few years as I systematically put in one ridiculously large Pull Request after another.  I know it hasn't been pretty but we have moved this project a very long way.

native/common/include/jp_floattype.h
```diff
@@ -62,22 +62,32 @@ class JPFloatType : public JPPrimitiveType
 	virtual JPPyObject  getArrayItem(JPJavaFrame& frame, jarray, jsize ndx) override;
 	virtual void        setArrayItem(JPJavaFrame& frame, jarray, jsize ndx, PyObject* val) override;
 
+	template <class T> static T assertRange(const T& l)
+	{
+		if (l < -3.40282346638528860e+38 || l > 3.40282346638528860e+38)
```
marscher: I guess these values are coming from the Java itself and would most likely never change in the future. Before your proposed change, these limits where obtained by a reflection. Actually I do think that there are different limits in different Java versions/Virtual machine implementations for float. However in the documentation the value is computed like this (2-2**(-23) * 2**127. I'm not sure if your expression could cause rounding errors.

native/common/jp_bytetype.cpp
```diff
@@ -237,19 +229,29 @@ JPPyObject JPByteType::getArrayItem(JPJavaFrame& frame, jarray a, jsize ndx)
...
+	if (getJavaConversion(&frame, match, obj) < JPMatch::_implicit)
+		JP_RAISE(PyExc_TypeError, "Unable to convert to Java byte");
+	type_t val = field(match.conversion->convert(&frame, this, obj));
+	frame.SetByteArrayRegion((array_t) a, ndx, 1, &val);
 }
 
+string JPByteType::asString(jvalue v)
+{
+	std::stringstream out;
+	out << (int) v.b;
```
marscher: Why do you need to explicitly cast this to int? I would assume the b value in the union is already of some integral type capable of representing bytes.

native/common/include/jp_floattype.h
Thrameos: All values for type constants are defined in the Java language specification and thus not necessary to look up.  When i was pushing things arround the lookup got lost.

In the case of float we are actually implementing the wrong behavior here as I discovered in the testing.  Unlike ints where overflow makes sense,  floats are supposed to overflow to infinity when the go out of range.  

When i went to test it I discovered that some paths hit thus overflow test and other paths mapped to infinity.  When i tried to make it more consistent more tests broke so we appear to expect different behaviors in different locations.  Looking at the language specification it appears mapping to infinity is correct.  Thus this code appears to be buggy.  

I am fine with either but making this check appear on the other paths is slower, harder to achieve, and no correct according to the language spec.  Thus i recommend silently mapping to inf on out of range.  If you would like it to raise overflow we need some clear rules on the check and not just arbitrary.  

Ints have similar issues on the arbitrary application of overflow.  We test on construction of a wrapper type but not on an array assignment.

native/common/jp_bytetype.cpp
Thrameos: The cast to int here was to so it printed a number rather than the null char or one char.

As JNI defines it the union member is a char which unfortunately does the wrong behavior when going to a stream.  

This particular method was routed arround later and got left as dead code.  It is removed in the instrumentation PR.  
Instead of implementing our own i managed to route to Pythons internal method which is a much cleaner solution.

Sorry about the dead code.  When i had to apply the patches out of order (speed/array then jni/classhints) it was a rather painfull remerge and left dead code from both branches.

marscher:
I also appreciate the work with you. Without your aid and project management efforts this project would be in an idle state (containing lots of bugs). I hope we will continue like that.

Thanks for this large refactoring and fixing lots of issues. The new type lookup and conversion mechanism are superior to what we used to have. It makes a lot of sense to have this part on the Java side.

The JNI functions to determine the modifiers have been replaced with a static table which extends the quoted documentation of the OpenJDK, I haven't understood yet how or where the extra values are arising (e.g. you test for serializable, which seems undefined in the documents linked in the header).

The JPContext thing is an ongoing effort, right?

Thrameos:
The modifiers in Java are fixed and limited by the Jar specification to a fixed number of bits  (it is forced to be a short by the spec IIRC).   Thus I took unused bits and packed in the various tests that we were using in JPype and put them to flags.  The idea is that rather than having to cross the JNI boundary every time we need to test does this object have serialization, is this an exception, or does this have the Functional notation,  instead I test for all these things on the Java side once and then pack them into the upper bits of the modifier.  I can push them up another 16 and use a long long if you think it needs to be more future proof.  Or we can pass a second flags field if that would be more clear.  

JPContext is basically a wrapper for the Java Env concept which I was hoping to be able to use to handle more than one Java JVM at a time.  Most likely much of its work will be relegated to frame instead at some future point when I figure out how to create sub JVMs.   The problem is of course that JNI does not specify any restrictions regarding JVMs running and was clearly designed with more than one JVM being possible.  But the JVM is not currently capable of running more than one copy due to global variables and the like.   I was able to boot two copies of the JVM in an experiment that gave me some hope by doing some magic on the loader side, but it was too unportable.    

Thus I am left with two ways to try to implement the feature.   First I can use JNI as a protocol for talking to a second process.  Thus when the user wants to create a second JVM I spin up a new process communicating with sockets from C++ and proxy all the JNI calls to it.   Second option is to use the host JVM as the communications channel.   There are many technological tradeoffs in this decision.  JNI failed to place a closure object on callback hooks which is a serious deficit for using it as a protocol for communications.  Without a closure pointer you can't manage local resources on callbacks.   Using the main JVM as a host saves a huge amount of work on the back end (C++) but makes the front end Python much more messy as knowing which resource belong to which JVM is a challenge.   It does move all the coms into Java which is far more portable than C++ which is a huge  win. Either way the JPContext needs to stay until I can decide which path is implementable (or abandon all hope) at which point it will be much more clear what the proper implementation should be.  

For now JPContext just is a big ball holding all of the previous global variables in one place.  It is required to create a JPJavaFrame and is the "persistent" portion of a Java environment.   It is not possible to hold a Frame between returns to Python so there has to be some kind of context object.

Thrameos:
A more specific answer to the modifiers are in ``org.jpype.typemanager.ModifierCode`` and ``native/common/include/jp_modifier.h``.  I should likely document each of the flags somewhere so it is well defined.  

Most are pretty obvious.   SPECIAL means that something has a predefined type wrapper in C++ and needs to be looked up by name when created rather than just getting the JPClass wrapper.  The rest match specific properties well defined in Java.

marscher:
Thanks for the clarifications. Lets stick to infinity mapping then.

Thrameos:
@marscher I rebased this to try to resolve the issues with LGTM.   There are still some bugs in this PR (outside of the test bench) that I have been resolving in instrumentation PR, but we likely have to commit this first before we can make useful progress as the deltas with this big change are hard to deal with.

marscher:
Don't feel obliged to fix all the warnings now. I just introduced the service, since I thought it could be become useful. We should fix the serious ones before we cut the next release.

Thrameos:
I managed to find the pattern that got C++ and Java working and hit all of the most obvious errors.   The big list of FIXMEs is useful but most were deferred for a time reasons so they are likely to remain that way for now.  The problem I was having was that their service uses Python 2 by default and thus most stuff was dying during the build problem.   Once I added a lgtm.yml to both the base and the rebased to that base it appears to work reasonably.   They are flagging a lot of false alarms right now which looking through the blogs indicates that it is actually a bug in their system.   The other minor annoyance is the bot adds new comments rather than editing the original like codecov.   I looked for a switch and couldn’t find one.

marscher:
On 3/4/20 11:33 PM, Karl Nelson wrote:
> I managed to find the pattern that got C++ and Java working and hit all of the most obvious errors.   The big list of FIXMEs is useful but most were deferred for a time reasons so they are likely to remain that way for now.  The problem I was having was that their service uses Python 2 by default and thus most stuff was dying during the build problem.   Once I added a lgtm.yml to both the base and the rebased to that base it appears to work reasonably.   They are flagging a lot of false alarms right now which looking through the blogs indicates that it is actually a bug in their system.   The other minor annoyance is the bot adds new comments rather than editing the original like codecov.   I looked for a switch and couldn’t find one.

Please do not push such changes directly to master. I think we should 
discuss them first. I also traversed the list and most of these issues 
seemed reasonable (e.g. catching an exception by value, throwing in a 
destructor, allocating but never deleting resources in JPContext, where 
are they actually handled?).

I agree that the repeated comments are annoying, but this way one can 
observe progress. Eventually I'm going to request this feature at LGTM ;)

Thrameos:
Okay.  I only push the lgtm file after i got it mostly working because it was required for the tool to have a baseline and it wouldnt take anything but master.  Did something else go in?    We can reset to the earlier point and resubmit a PR if you have concern.  Unless it is completely administrative ( missing file, misspelling, comment removal demanded by employer, etc) I dont push to master.

The throw in dtor is a false alarm as the dtor is marked as throws.  The deleted items are controlled by Java as all resources in C++ except object and array are Java controlled.  They are killed by a JNI call in the cleanup routine with all other resources.  I moved them to type factory to be explicit.  So although they are suspect they are each correct code.  I put in the ignore code on the false alarm, but they do not take until they reach master.

I am fixing the issues and finishing the coverage on instrumentation PR.  Though we may want to break that up into pieces.  Based on current progress it is at least a month before I will complete the C++ coverage of Python section.  The actual Python module is over 90 on instrumentation PR after refactoring the find jvm code to make it easier to hit.

marscher:
I refuse to ever review such a massive pr in the future, if you gonna force push it over and over again. This also forces me to go over whole beast again. A simple merge of master should be sufficient to catch up with its state.

marscher:
You do not release the primitive types in JPTypeFactory:
``` c++
// in JPTypeFactory_destroy()
	for (int i = 0; i < sz; ++i)
	{
		// Do not delete the shared primitive types.
		if (dynamic_cast<JPPrimitiveType*> ((JPResource*) values[i]) != 0)
			continue;
```

Additionally the services like ProxyFactory or ReferenceQueue are not deleted, I couldn't spot it. If there is a friend class taking care of its disposal, it should be documented.

native/common/include/jp_context.h
```diff
@@ -0,0 +1,310 @@
...
+JPRef<jref>::JPRef(const JPRef& other)
+{
+	m_Context = other.m_Context;
+	if (m_Context != NULL)
+	{
+		JPJavaFrame frame(m_Context, m_Context->getEnv());
+		m_Ref = (jref) frame.NewGlobalRef((jobject) other.m_Ref);
+	} else
+	{
+		printf("NULL context in JPRef()");
```
marscher: Shouldn't that raise? I assume this should never happen

Thrameos:
Good catch.  I will correct the error.  The reason primitives were different was for bootstrapping issue which is now fixed.  Appears that I didn't follow through with the removal.

I try to avoid requesting review until I get the pull request finished and squished into a good state.  The reason this got repushed is that the coverage tool and lgtm both refused to get a base properly.  I know it causes problems for your review.   Both of those tools wouldn't work until they have the code was based properly with the head.   Merging in master didn't do it and just broke if further.   I believe the problem is because these PRs all started before those tools were introduced, so I don't think it is likely to be a problem again.  I will just need to rebase my private repos to master again. 

The other thing that forces a rebase is of course when something goes so far out of date that it requires to be freshened, but that only affects me as usually that is unreview code.   When I have 40 commits and there is a bad conflict, the only clean way to freshen that I found is resetting to origin, diffing to get a patch file, fast forwarding to the head and then applying the hand edited patch.

The "good news" is this is one and the following instrumentation PR should be the last PR of this size ever again which touches the core code like this.  After this point we can build up rather than tearing down.  

Looking over the private repo branchs there will be reasonably large PR for the following subjects:

1) Add support for multiple jvms.   Most of it will be bolt on with the exception of call and array access request.   Those unfortunately will touch every file with an "if" statement.   I still want to get this one by the end of summer as it seems to be a frequent request.
2) Reverse bridge from Java.   Most of the code is in Java land with support classes for accessing Python objects from Java, which changes for PR but it does require a change to the boot sequence so that the attachment is done possible.  The code for JNI part is pretty massive but it is all new code and not something touching existing.  This one have been back burner for a while as it is not currently on my need path. 
3)  Support of extending a class in Python.   This one is unfortunately very large.  To make it work we have to disassemble the Java byte code for the class and create an interface for each entry point we want to override, create an interface for each, and then reassembly the whole works back into Java byte code.  Despite a year on this one, it still isn't complete.  (unfortunately is origin is currently very dead so I need to rebase this one.)

If something does seem too massive to review, please ask me to break it down.  Also if you have a direction request (actually need something to happen before something else) it is great.  Backporting from my primary repo where all of these are being assembled together take a lot of time because I have to strip out all the unrelated changes.   Things like the changes to exception handling routines would be much better reviewed standalone (JP_RAISE_* to JP_RAISE), but they create a nightmare on my end if the pull requests go in the wrong order as it produces massive conflicts with development repo as these silly small improvements touch huge sections of code.  If I know the order of features I can structure much easier on my end.  That said, other than something like this where features are intertwined together like the type system it should be possible to single subject it.

Is there anything else on this current PR that you like addressed?  I really don't want to keep pushing on it with changes as it is already a beast, and all my other pull requests (quick multidimensional array to JArray, maps fix, and instrumentation are currently based on this point)

And if you ever really feel the need to just call me up and yell at me, you can contact me at my organization under Karl E Nelson.

native/common/include/jp_context.h
Thrameos: Agree I will change it to 

```
    JP_RAISE(PyExc_SystemError, "NULL context in JPRef()");
```

Unfortunately, I do not think I can actually test that code without instrumentation.

marscher: We can test it later, when the instrumentation thing is in place.

native/common/jp_context.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+JPPlatformAdapter* GetAdapter()
+{
+	static JPPlatformAdapter* adapter = new PLATFORM_ADAPTER();
+	return adapter;
+}
+}
+
+
+#define USE_JNI_VERSION JNI_VERSION_1_4
```
marscher: Quick question: why do you need to pin the version here (I guess it is already set somewhere else?)

native/common/jp_encoding.cpp
```diff
@@ -217,37 +217,6 @@ unsigned int JPEncodingUTF8::fetch(std::istream& in) const
 	return -1;
 }
 
-//*****************************************************************
```
marscher: Why did you delete this hilarious comment? I just tripped on it now after so many years - made my day

native/common/jp_field.cpp
```diff
@@ -17,113 +17,75 @@
...
-	m_Class = clazz;
-	m_Name = JPJni::getMemberName(fld);
-	m_IsStatic = JPJni::isMemberStatic(fld);
-	m_IsFinal = JPJni::isMemberFinal(fld);
-	m_FieldID = frame.FromReflectedField(fld);
-	m_Type = JPJni::getFieldType(m_Field.get());
-	m_TypeCache = NULL;
-
-	//	JP_TRACE_OUT
+	this->m_Class = cls;
```
marscher: the self reference code style is newly introdcued in this class. It is not mandatory and makes the code harder to read IMO. BTW what happened to good old initializer lists?

native/common/jp_javaframe.cpp
```diff
@@ -0,0 +1,1116 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+ *****************************************************************************/
+#include <Python.h>
+#include "jpype.h"
+
+#define USE_JNI_VERSION JNI_VERSION_1_4
```
marscher: What happens if this constant is changed in one unit (e.g. increased), while it remains the old value elsewhere?
Can we redirect the jni include via an extra header, which defines the used jni version once, and use it whenever jni is included?
If it does no harm to mix versions, I'd be fine with the current solution.

native/common/jp_encoding.cpp
Thrameos: My employeer was less than thrilled.    So sadly it must go away.

native/common/jp_javaframe.cpp
marscher: There is one define here, and the other in jp_context

native/common/jp_context.cpp
Thrameos: Not sure.  I think copied this frome somewhere.  Is there a benefit to using a later veesion?

native/common/jp_method.cpp
```diff
@@ -14,291 +14,394 @@
    limitations under the License.
 
  *****************************************************************************/
-#include <algorithm>
+#include <jpype.h>
```
marscher: duplicated include

native/common/jp_javaframe.cpp
Thrameos: This one is a stray and should be removed.  Only the one in jp_context is used.

native/common/jp_method.cpp
```diff
@@ -14,291 +14,394 @@
...
+		jmethodID mid,
+		JPMethodList& moreSpecific,
+		jint modifiers)
+: m_Method(frame, mth)
 {
-	m_Class = clazz;
-	m_hasStatic = false;
-	m_Cached = false;
-	m_IsConstructor = isConstructor;
+	this->m_Class = claz;
```
marscher: pls no this

native/common/jp_field.cpp
Thrameos: Removing it.  It is a stupid editor trick I use to help with auto complete while typing fresh code.  It should have been removed during my clean up pass.

Thrameos: Removing it.  It is a stupid editor trick I use to help with auto complete while typing fresh code.  It should have been removed during my clean up pass.

Thrameos: Fixed in 19 places.

native/common/jp_method.cpp
Thrameos: Fixed

native/common/include/jp_context.h
Thrameos: This one is harder than I expected.   The reason for the print was that I can't throw here because throw requires JPException but JPException requires JPRef creating a perfect circular dependency.   I will figure out how to break it.

native/common/jp_context.cpp
marscher: I do not know. My suspicion is, that it only matters during JNI_CreateJavaVM, whereas the version is specified. 
I think the definition in jp_javaframe is just not doing anything, so it can be deleted.

Thrameos: Okay so it should be set in jp_context and no where else.  This define is used only to get the ENV version when creating the JVM and when attaching threads all of which must happen in context.  The value 1.4 is correct as we do not use any 1.6 features.   We will have to update this with a new value once the JVM adds Value types.

http://cr.openjdk.java.net/~jrose/values/values-0.html


```
jp_context.cpp:#define USE_JNI_VERSION JNI_VERSION_1_4
jp_context.cpp: jniArgs.version = USE_JNI_VERSION;
jp_context.cpp: jint res = m_JavaVM->functions->GetEnv(m_JavaVM, (void**) &env, USE_JNI_VERSION);
jp_context.cpp: return JNI_OK == m_JavaVM->functions->GetEnv(m_JavaVM, (void**) &env, USE_JNI_VERSION);
jp_context.cpp: jint res = m_JavaVM->functions->GetEnv(m_JavaVM, (void**) &env, USE_JNI_VERSION);
```

native/common/jp_encoding.cpp
marscher: Everywhere popular politics. How did they become aware of this? Are there legal concerns?

native/common/include/jp_context.h
Thrameos: Finally got it so resolved.  The solution made unfortunately was to make a function just to raise an exception.  :sob:

marscher:
Thanks for incorporating the style changes. Could you please delete the jpype.jar, it should not live in the repository as it is being generated. And possibly put it in gitignore.

native/common/jp_encoding.cpp
Thrameos: I have been *ahem* slightly skirting the limits of my laboratory policies.  When I joined this line of work in 2001 I had an agreement that I could operate on open source so long as it was not related or had security concerns.  But since then we have changed  contracts and someone decided to publish a new policy that restricts my contribution to no more that "25%" of the total project code.  Of course no one pushed to policy to any group leaders so it apparently sat on some web site until I was pushing another open source project out the door, which me being a good citizen recognized as an issue, so I followed the procedure.   Lawyers were involved and way more time spent as work that I have ever spent on this project there (seeing as I only work on it at night and at 5 AM).   Needless to say I was given a list of items to immediately change, a trip to the ethics office to explain why doing stuff for free does not count as a conflict of interest, and after all that the lawyer concluded (paraphrasing) "no conflict, typical professional activity for a programmer, no concerns", but no one is willing to sign the paper to either release the code or claim the code.  Should they decide to claim the code a "code release number" will be assigned and the "auspices statement" will need to be added claiming this code was developed with gov't dollars, but thus far given that the review was the only thing that ever cost the gov't a penny noone appears willing to give me that stamp.  I make the call once a month (every time the automatic system sends me a notice) to all IP and ask them to complete the release paper but given noone will every make a cent on this project it is a very low priority.  Once "released" I am allowed to file with copyright office to assert that I own the code (not my employer) I may have to do just to make sure no one who gets our contract in the future can raise the claim again.  Then I will then stick it in a drawer to be found in my personal papers sometime in the year 2060.

Bottom line... I had to put it through review which required the comment and a handful of other off-color comments from other contributors to be removed.  Ethics office has agreed that I am operating in good faith contributing to software used by the lab.  IP office has concluded they will define "25%" as whatever number I have contributed and "Apache" as an acceptable license as a policy they arbitrarily defined is not a legal matter.  One tree died a pointless death. 

*LEGAL: This comment is a somewhat humorous restatement life at the laboratory, and does in no way impute the honor of the laboratory or the management.  Opinions expressed at 6 AM do not reflect the opinions of the laboratory nor the US government.*

Thrameos:
Okay that is new.... do we have two appveyor CI checks?

marscher:
Well, if you created the branch inside this repo, then yes. One CI run for the merge and one for the branch itself.

marscher:
But I'm also fine with merging it now.

PR #590, opened 2020-02-28
Release

Thrameos:
Up versions for release.  Not sure I hit everything properly.   

This pull request is on the collection_fix which is still pending.


Thrameos:
@marscher I did my best to make a release, but no guarantees.  Hopefully it will be smoother than last time.

PR #592, opened 2020-02-29
Instrumentation

Thrameos:
This pull request adds the ability to trigger faults during the coverage testing so that we can exercise each of the failure paths and verify their function.   Because there are many fault paths, this request is going to take a while to complete.  

This is currently a test pull built on classhints2 and serves as the test bench to verify that the final core rewrite is actually covering all of the paths.   

In order to trigger faults, the compiler is set to compile in fault hooks on enable coverage.   This converts all of the JP_PY_TRY and JP_TRACE_IN statements into fault hooks.  It uses C++ const expression to convert the name of the function into a hash code which can be triggered with a specific fault.  As a coverage method this doesn't help much as the coverage tool recognizes the difference between our fake fault triggers and the real condition and thus calls it just another untested branch.   But it does allow us to test some of the more deeply buried error handling.  Most faults will trigger a ``SystemError`` which represents Pythons error for something just short of fatal.  The test bench then attempt to trigger this fault with SystemError then again with any errors that are expected and finally again without an error to make sure we hit each path and managed to exit unscathed.

Thus far it has turned up about a dozen bad error handling paths and exposed about an equal number of bugs by allowing an easy way to generate off normal conditions or unusual conditions (using partially unpopulated objects).  Thus it is likely worth the effort.


native/common/jp_arrayclass.cpp
```diff
@@ -100,29 +100,5 @@ JPValue JPArrayClass::newInstance(JPJavaFrame& frame, int length)
...
-		JP_TRACE("Index");
-		Py_ssize_t sz = PyNumber_AsSsize_t(args[0], NULL);
-		if (sz < 0 )
-			JP_RAISE(PyExc_ValueError, "Invalid size");
-		return JPValue(this, m_ComponentType->newArrayInstance(frame, (int) sz));
-	}
-
-	JP_RAISE(PyExc_TypeError, "Arrays require int or sequence parameters");
-	JP_TRACE_OUT;
+	JP_RAISE(PyExc_SystemError, "Not used");
```
marscher: if it should not be used anymore, it can or will be deleted?

Thrameos: We have some things that are required to be overridden by C++.   I could let it fall it fall back to the base once I have finished coverage.  But these need to be errors until everything has cleared check.

Thrameos:
@marscher  I think this one should likely go to review and start another PR as I finally lost the coverage diffs which means I will need another baseline to establish further progress.  Or if you want me continue on this one until I run out of bugs (may be a few weeks at current rate) I would have to perform a force push.  Sadly 1100 tests is still not enough to get us over 60% coverage.

We also need to clear that appveyor branch.  I am not sure what it is doing but since it appeared nothing is clearing the CI.

Tasks done:
- Add fault system to trigger faults.  Faults are triggered for JP_TRACE, JP_PY_TRY,  JP_FAULT, and JP_JAVA_CHECK to try to get at parts of the code.
- Fixed float OverflowError to follow Java rules.
- Discovered a lot of strange edge cases in the type system when I pushed the test matrix out.  Still a fair number of FIXMEs that will need to be dealt with.
- Reorgs the tests to make it easier to use a diff tool to copy between different types (oh my was that boring)
- Removed piles of dead code.
- Reworked exceptions from Release* methods.  Exceptions are not permitted in those paths, so I had to do some rework to handle it better.
- Reorged a portion of the test harness.  We had a lot of redundant harness code that needed to be consolidated.   I put everything that had to do with fields and common methods into jpype.common.Fixture.

test/jpypetest/common.py
```diff
@@ -30,6 +30,44 @@ def version(v):
...
+
+class UseFunc(object):
+    def __init__(self, obj, func, attr):
+        self.obj = obj
+        self.func = func
+        self.attr = attr
+        self.orig = getattr(self.obj, self.attr)
+
+    def __enter__(self):
+        print("ENTER")
```
marscher: debug prints?

Thrameos:
Yes.  I will remove those.   Unfortunately, debug prints do not show in Pytest unless you give the “-s” switch so they tend to get lost.

test/jpypetest/test_fault.py
```diff
@@ -0,0 +1,1249 @@
...
+
+    This may be one of the most tedious files every written by human hands.
+    Coffee is advised.
+
+    """
+
+    def setUp(self):
+        common.JPypeTestCase.setUp(self)
+
+    @common.requireInstrumentation
```
marscher: I think this decorator could also be attached at class definition level

test/jpypetest/test_fault.py
```diff
@@ -0,0 +1,1249 @@
...
+
+    This test suite brutally tries to force an exception to be thrown
+    at each entry point and function call.  The exception is controlled
+    based on the name of the function in the JP_TRACE_IN and JP_PY_TRY
+    block.  Specific fault points are also triggered to produce
+    abnormal objects which can then be passed to trigger error handling
+    behaviors for off normal conditions.
+
+    This may be one of the most tedious files every written by human hands.
+    Coffee is advised.
```
marscher: I will get one :+1:

test/jpypetest/test_fault.py
Thrameos: My long term plan is to move these tests from the giant fault handler test back to the unit test for each of the corresponding portions.  I did that for the primitives.  So the decorator was used to make the tests more portable.

test/jpypetest/test_fault.py
```diff
@@ -0,0 +1,1249 @@
...
+#        @JImplements("java.util.function.DoubleUnaryOperator")
+#        class f(object):
+#            @JOverride
+#            def applyAsDouble(self, d):
+#                return d
+#        jo = JObject(f(), "java.util.function.DoubleUnaryOperator")
+#        raise RuntimeError(jo.toString())
+
+    @common.requireInstrumentation
+    def testJPProxy_dealloc(self):
```
marscher: shouldn't this trigger an exception? What am I missing here

test/jpypetest/test_fault.py
```diff
@@ -0,0 +1,1249 @@
...
+        i = None
+        _jpype.fault("JPClass::getStaticField")
+        with self.assertRaisesRegex(SystemError, "fault"):
+            i = jf.static_object_field
+        _jpype.fault("JPClass::getField")
+        with self.assertRaisesRegex(SystemError, "fault"):
+            i = jfi.object_field
+        self.assertEqual(i, None)
+
+    @common.unittest.SkipTest
```
marscher: Skip for which reason? Please add a FIXME or TODO

test/jpypetest/test_jvmfinder.py
```diff
@@ -66,34 +64,113 @@ def test_get_from_bin(self, mock_real_path, mock_path_exists, mock_os_walk):
...
             self.assertEqual(p, expected.strip())
 
         # this version has no java_home binary
         mock_mac_ver.return_value = ('10.5', '', '')
         p = finder._javahome_binary()
-
         self.assertEqual(p, None)
 
+    @common.unittest.skip
+    def testPlatform(self):
```
marscher: This tests a implementation detail.

test/jpypetest/test_fault.py
Thrameos: Assuming you are referring to the deallocation.  Python deallocation routines are special.  Because they trigger underneath everything else and with their own timing.  Therefore, dealloc must eat any exceptions that get triggered or it will result in a Python SystemError thrown at a random time.  Most of the time will be at the end of scope (and thus outside of the test case).   

To handle this properly during testing you will see me using ``def f(): do something;  f()``  which will trigger the action and force a clean up so that we can detect if the call triggers a SystemError inappropriately.  That is also the reason for all the ``assertRaisesRegex`` calls as I need to ensure we get the correct trigger.

Therefore the correct behavior is to trigger a fault and make sure that we DO NOT get a fault because if we do then we will trigger bad behavior in Python (interrupt GC or fail to free memory).

test/jpypetest/test_jvmfinder.py
```diff
@@ -66,34 +64,113 @@ def test_get_from_bin(self, mock_real_path, mock_path_exists, mock_os_walk):
...
+            pathmock.realpath = f
+            pathmock.abspath = f
+            pathmock.dirname = os.path.dirname
+            pathmock.join = os.path.join
+            finder._get_from_bin()
+            self.assertEqual(
+                pathmock.dirname.mock_calls[0].args, (finder._java,))
+
+    @common.unittest.skip
+    def testCheckArch(self):
```
marscher: this tests needs the same (reverse) implementation like the actual impl. I would recommend against this pattern, since when you need to make a change you need to double implement it for the test as well. I'm curious if still there are good reasons to do so.

test/jpypetest/test_fault.py
Thrameos: Will do.   Most of the skip tests are because something different happened in the CI than happened on my system.  In particular those for the jvmfinder are all working tests on my system that fail or hang the CI.  (I believe it is a problem with mock versions.)  

I am hoping when I finish this pass the ^C crash issue on Spyder will fix.

test/jpypetest/test_jvmfinder.py
Thrameos: This is mostly just to make sure that coverage tool hits all code.  I really don't care about the result other that making sure that we don't have a syntax error down one of the paths (which in one case there was a typo that only hit on one platform.)

Would you like me to put a decorator on all those tests to be "coverageOnly"  so that they only trigger during the coverage phase?

setupext/build_makefile.py
```diff
@@ -0,0 +1,199 @@
...
+class BuildMakefileCommand(build_ext):
+    """
+    Override some behavior in extension building:
+
+    1. handle compiler flags for different compilers via a dictionary.
+    2. try to disable warning -Wstrict-prototypes is valid for C/ObjC but not for C++
+    """
+
+    # extra compile args
+    copt = {'msvc': [],
```
marscher: should we unify the cflags handling in a base class, so it does not run out of sync once the central build_ext changes?

Thrameos: Yes.  That would be a good idea.   I am curious if the Makefile it is producing is working for you now.  I have pattern for VC as well but it is buggy because I don't have a pattern to make the dep section properly.

doc/CHANGELOG.rst
```diff
@@ -6,6 +6,9 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 - **Next version - unreleased**
   - java.util.Map conforms to Python abc.Mapping API.
 
+  - float properly follows Java rules for conversion from double.
+    floats outside of range map to inf and -inf.
+
```
marscher: I think you also added domain boundary checks for integral types. These just used to overflow, now the user will face an exception (ValueError).

test/jpypetest/test_jvmfinder.py
Thrameos: Same as the previous.   To make sure we don't have a problem with a typo or accidental edit I recommend those tests.  Obviously they need to be updated with the code changes rather than just a fixed static test like the others.  A decorator should help make it clear and also make it so that it is only tested in the coverage phase.

doc/CHANGELOG.rst
Thrameos: This is one of the parts that I am unsure about.   I believe the correct behavior is to throw overflow on an automatic conversion and not throw when it is a cast....

```
    i  = 2**34
    test = JClass('jpype.common.Fixture')()
    test.callInt(i)  <= this should throw OverflowError
```

but

```
    i  = 2**34
    test = JClass('jpype.common.Fixture')()
    test.callInt(JInt(i))  <= this should NOT throw OverflowError
```

The reason being that ``test.callInt(JInt(i))`` is equal to ``test.callInt((int)i)`` on the Java side.   Thus if Java doesn't overflow error then Python should not as well.   I don't think I have achieved this correct behavior yet.   

Any thoughts?

marscher:
The OSX Travis job fails, because we need to manually install ant on OSX now. Seems like they have changed the image/target.

doc/CHANGELOG.rst
marscher: You're right, that implicit conversion should raise, while explicit shouldn't. But I do not know how to achieve it - sorry.

test/jpypetest/test_jvmfinder.py
marscher: If it is just in place to test coverage/ e.g. correct syntax I would just ensure that the code paths are all hit. This code path is only used on windows, right? So it would be sufficient to test coverage on Appveyor as well. Codecov does combine reports from different platforms/configurations without objection.

test/jpypetest/test_jvmfinder.py
marscher: I don't see this point. It only adds more maintaining overhead + doesn't add new test cases as it only reproduces the implementation detail. Same as above, if you need to check for SyntaxErrors you need to run the code paths via an API call.

test/jpypetest/test_fault.py
marscher: In my experience the mock library is very stable. Can you please point me to a particular test, which fails only on CI, but succeeds on your system?

Thrameos:
The OSX image was pretty busted a while back.   I disabled one copy of it and forced the brew pattern to update which was the required kludge.   But that added 5 minutes to the build and then they updated it again.   I will make another run at it tonight.

test/jpypetest/test_fault.py
marscher: That was stupid, you have decorated the one already :)

test/jpypetest/test_fault.py
marscher: Thanks, this makes total sense now. Sometimes tests also deserve documentation (sometimes not).

test/jpypetest/test_fault.py
marscher: agreed

test/jpypetest/test_jvmfinder.py
Thrameos: Not sure I trust appveyor to run a code coverage.   It is a tradeoff.  Either I make a somewhat useless test on Linux (mocking to see if the paths get hit) with the other tests for coverage.  Or I face the challenge of getting code coverage to work on windows.   I think that given the hours we tend to burn on making the CI work (push, wait 15 minutes, see it didn't work, push, rinse, lather, repeat) making it work once on one system is likely the best option.  We have both put in a huge set of commit after commit pushing one line in over and over trying to get the CI working, so I favor the somewhat useless mock test approach.

setupext/build_makefile.py
marscher: it fails with some AttributeError on my system:

```
$ python setup.py build_makefile
/home/marscher/miniconda3/envs/jpype_nonp/lib/python3.8/distutils/dist.py:274: UserWarning: Unknown distribution option: 'use_scm_version'
  warnings.warn(msg)
running build_makefile
running build_java
Jar cache is missing, using --enable-build-jar to recreate it.
  ant -Dbuild=../build -f native/build.xml
Buildfile: /home/marscher/sources/jpype/native/build.xml

compile:
    [javac] /home/marscher/sources/jpype/native/build.xml:20: warning: 'includeantruntime' was not set, defaulting to build.sysclasspath=last; set to false for repeatable builds
    [javac] /home/marscher/sources/jpype/native/build.xml:27: warning: 'includeantruntime' was not set, defaulting to build.sysclasspath=last; set to false for repeatable builds

jar:

all:

BUILD SUCCESSFUL
Total time: 0 seconds
running build_thunk
Building thunks
Skip build thunks
building '_jpype' extension
  File "setup.py", line 9, in <module>
    setup(
  File "/home/marscher/miniconda3/envs/jpype_nonp/lib/python3.8/site-packages/setuptools/__init__.py", line 144, in setup
    return distutils.core.setup(**attrs)
  File "/home/marscher/miniconda3/envs/jpype_nonp/lib/python3.8/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/home/marscher/miniconda3/envs/jpype_nonp/lib/python3.8/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/home/marscher/miniconda3/envs/jpype_nonp/lib/python3.8/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/home/marscher/miniconda3/envs/jpype_nonp/lib/python3.8/site-packages/setuptools/command/build_ext.py", line 87, in run
    _build_ext.run(self)
  File "/home/marscher/miniconda3/envs/jpype_nonp/lib/python3.8/distutils/command/build_ext.py", line 340, in run
    self.build_extensions()
  File "/home/marscher/sources/jpype/setupext/build_makefile.py", line 195, in build_extensions
    self.compiler.write()
  File "/home/marscher/sources/jpype/setupext/build_makefile.py", line 69, in write
    library = os.path.basename(self.library)
AttributeError: 'Makefile' object has no attribute 'library'
```

test/jpypetest/test_jvmfinder.py
marscher: I agree that setting up CI can be tedious work. But to simply check for a well running function, you do not need to check the details, right?
M$ VC can also produce coverage (binary files), which can be converted to xml by another tool (yay). https://github.com/jsargiot/visual-coverage
But please do not waste your precious time on this BS :)

marscher: Sorry I got you wrong, you have to mock to actually trigger this code paths. I just didn't like the struct packing binary thing as it seemed redundant with the actual implementation.

Thrameos: Oh the mock packing.   It was just a way to try to fake a binary file that I don't have.  The images on the CI don't have the wrong architecture loaded (32bit on 64bit, etc)  Thus I either put those images into the test bench (dubious licence wise) or use mock to fake the bytes in the right place.  I was planning to the same thing for elf as windows pei files so that we cover it.    I suppose I could make short header file only versions of the libraries for each architecture.) 

I can make it go through the entry point rather than hitting the internal function directly if you want.   I just have to mock it to windows and call the top level function.  That one is no problem.

(I am guess you want to refer to testCheckArch)

setupext/build_makefile.py
Thrameos: Okay I will have to find out which disttools is causing the problem.  I checked on Ubuntu under windows and it works so will have to make a wider scan of versions to see if I can find a stable pattern.   

I am not very happy with Python disttools.   Their API requires overriding stuff and it doesn't seem very stable.  I can make it do stuff that I need but it almost always needs to go into the machinery  which seems like poor planning on their part.

doc/CHANGELOG.rst
Thrameos: I think that it is just adding a bypass flag on PyJPNumber_new.   But I want to do this as a separate PR rather than part of a huge change ball.

marscher:
On 3/9/20 10:53 PM, Karl Nelson wrote:
> Thrameos commented on this pull request.
> 
> 
> 
>> @@ -6,6 +6,9 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
>   - **Next version - unreleased**
>     - java.util.Map conforms to Python abc.Mapping API.
>   
> +  - float properly follows Java rules for conversion from double.
> +    floats outside of range map to inf and -inf.
> +
> 
> I think that it is just adding a bypass flag on PyJPNumber_new.   But I want to do this as a separate PR rather than part of a huge change ball.
> 
Sounds more reasonable.

marscher:
On 3/9/20 10:51 PM, Karl Nelson wrote:
> Thrameos commented on this pull request.
> 
> 
> 
>> +
> +-include $(DEPFILES)
> +""", file=fd)
> +
> +
> +# Customization of the build_ext
> +class BuildMakefileCommand(build_ext):
> +    """
> +    Override some behavior in extension building:
> +
> +    1. handle compiler flags for different compilers via a dictionary.
> +    2. try to disable warning -Wstrict-prototypes is valid for C/ObjC but not for C++
> +    """
> +
> +    # extra compile args
> +    copt = {'msvc': [],
> 
> Okay I will have to find out which disttools is causing the problem.  I checked on Ubuntu under windows and it works so will have to make a wider scan of versions to see if I can find a stable pattern.
> 
> I am not very happy with Python disttools.   Their API requires overriding stuff and it doesn't seem very stable.  I can make it do stuff that I need but it almost always needs to go into the machinery  which seems like poor planning on their part.
> 
the distutils lib is crappy indeed. We committed our fate into 
setuptools, as it tries to solve most of distutils issues. But as it 
tries to preserve as much (weakly designed) API as possible, it cannot 
solve all problems. The most horrible thing is, that distutils classes 
are not even new-style objects, so we cannot rely on super() calls for 
invoking parent implementations.

Just out of curiosity, would Ubuntu notice/behave differently when 
emulated under Win?

Thrameos:
Ubuntu under WLS is an improvement over cygwin as it offers all the capabilities of Ubuntu with good package manager and at the same time lets you run windows commands just by calling the exe.   The only differences have been with some kernel level stuff.  For example docker does not work from emulated linux, so I have to run a separate windows docker system.   The good news is I can run the docker command from within the linux, but it launches the windows container manager.  It is good enough to allow me to test both windows and linux stuff from the same window and often within the same command script so it is a big time saver.

test/jpypetest/test_jvmfinder.py
Thrameos: Promoting this to issue.

Thrameos:
I applied autopep8 and added comments to address review concerns.   The rest were transferred to issues. Then once the CI clears I will merge.

PR #596, opened 2020-03-05
Add ltgm to master

Thrameos:
Somehow my "clean" copy got the changes out of order.  

The reset point is cd49a2b6602d86e5506fa8f5dea701dc3a4e84ce.   Then this PR will add lgtm.yml to the top level.  Then I will rebase the other PR to the master head.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,23 @@
+path_classifiers:
+   test:
+     - setup.py
+     - setupext
+     - test
+     - examples
+
+   docs:
+     - doc
+
+extraction:
+   cpp:
+       index:
+           build_command: 
+             - python3 setup.py build
+
+   java:
+       index:
+           build_command: ant -f native/build.xml
+
+   python:
+       python_setup: 
+           version: "3"


marscher:
Thanks. These changes are already on master, so I don't get the point of this now. I just got concerned a bit, as you added these changes without a PR. Anyhow, why isn't the library section included, as it adds the relevant paths for inspection?

Thrameos:
Yeah sorry about that.  I tried a few times using pull requests on my own branch and their system would not accept it, so I got a bit frustrated and pushed my master back seeing as it was supposed to be a missing file.   Their tool expects that the master has run lgtm or it will report every PR as failed regardless of what happens in the PR.

Library appears to make things not show up in the list of warnings.  Their definition appears to be library=="outside library code that I do not want to be warned on, it isn't mine".   So I had to remove that section to make it work.

Thrameos:
I certainly don't mind new quality checks (the more the merrier), but I do need a bit of warning so that I don't wake up to a pile of red lights on the CI and a D- rating staring me in the face.  Just ruins the morning.  ;-)

marscher:
Sorry for ruining your morning by this bad grade. I think the code quality isn't in such a bad shape and you are doing an outstanding job on further improving it. I should have noticed you about arming that gunnery. I expected it to fire after after the merge.

PR #597, opened 2020-03-06
Changes to build makefiles

Thrameos:
This pull request clears a number of warnings from LGTM and introduces a command to create a makefile to help with incremental builds.   The makefile currently only works for linux type systems, but otherwise should be fully functional.

To create a makefile use 

```
    python setup.py build_makefile
```

It is tied to the build ext system so unfortunately it doesn't run properly if the ``build`` directory already exists.  You can then edit the makefile and use it to test builds.  Once this is is working we can remove the project/jpype_cpython/Makefile.

This is backport of a portion of instrumentation pull request.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-import warnings
 import setupext
 from setuptools import setup
 from setuptools import Extension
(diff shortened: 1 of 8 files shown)


Thrameos:
Ready to go.  This is safe to apply in any order.

Thrameos:
This Pull request is part of the approved instrumentation PR.  Thus I will apply this one first.

PR #600, opened 2020-03-08
Support for multidimension primitive conversion

Thrameos:
In today's edition of how you find the time, we add support for direct conversion of multidimensional primitive arrays from buffers.  To use this feature we call ``JArray.of(obj)`` where obj is any type that supports the buffer protocol.  There is an optional second argument that defines the type to convert to.  

Thus to take a numpy double array to a Java float array with the same shape use...
```
  a = np.zeros((5,2,6,10))
  ja = JArray.of(a, dtype=JFloat)
```

It will automatically figure out the dimensions and size the result properly.

Implementation details.  This converts the Python buffers into a series of Java arrays along the last dimension then calls a support routine that unpacks the arrays back into a multidimensional array.  It is split into 4 parts.  The front end just does arguments checks,  the buffer handler decides what types we are unpacking to, a per type unpacker does the conversion (using a template) and then the Java portion reorganizes the result.


Thrameos:
This review will be freshened and rebased.  Please hold of reviewing until this is complete.

test/jpypetest/test_array.py
```diff
@@ -176,3 +173,184 @@ def testJArrayCopyRange(self):
...
+        with self.assertRaises(TypeError):
+            JArray.of(a)
+        a = np.random.randint(0, 2*16-1, size=1, dtype=np.uint16)
+        with self.assertRaises(TypeError):
+            JArray.of(a)
+        a = np.random.randint(0, 2*32-1, size=1, dtype=np.uint32)
+        with self.assertRaises(TypeError):
+            JArray.of(a)
+        a = np.random.randint(0, 2*64-1, size=1, dtype=np.uint64)
+        with self.assertRaises(TypeError):
```
Thrameos: This test needs to require Numpy

Thrameos:
It is hard to hit every one of the patterns for the buffer conversion.   I had to make special versions to hit each path during development.   I need to find some way to make numpy supply buffers with different layouts.  Perhaps I should make should expand instrumentation to make it skip to different cases.

native/common/include/jp_primitive_accessor.h
```diff
@@ -64,5 +64,87 @@ class JPPrimitiveArrayAccessor
...
+
+	// Align with the first element in the array
+	char *src = buffer.getBufferPtr(indices);
+
+	// Traverse the array
+	while (true)
+	{
+		if (indices[u] == view.shape[u])
+		{
+			int j;
```
marscher: j can be declared outside the while loop

native/python/jp_pythontypes.cpp
```diff
@@ -458,3 +458,37 @@ JPPyBuffer::~JPPyBuffer()
...
+{
+	char *pointer = (char*) m_View.buf;
+	// No shape is just a 1D array
+	if (m_View.shape == NULL)
+	{
+		return pointer;
+	}
+
+	// No strides is C contiguous ND array
+	if (m_View.strides == NULL)
```
marscher: Out of curiosity, what if the user passes a Fortran aligned buffer?

test/jpypetest/test_array.py
```diff
@@ -176,3 +173,184 @@ def testJArrayCopyRange(self):
...
+        _jpype.fault("JPLongType::newMultiArray")
+        with self.assertRaisesRegex(SystemError, "fault"):
+            JArray.of(b, JLong)
+        _jpype.fault("JPFloatType::newMultiArray")
+        with self.assertRaisesRegex(SystemError, "fault"):
+            JArray.of(b, JFloat)
+        _jpype.fault("JPDoubleType::newMultiArray")
+        with self.assertRaisesRegex(SystemError, "fault"):
+            JArray.of(b, JDouble)
+
```
marscher: I cannot spot a test for the handling of memory fragmented (stridded) arrays.

Thrameos: I need to find a way to test fortran and strided arrays.  Both are sources of potential bugs.  Any idea how those should be tested?

native/python/jp_pythontypes.cpp
Thrameos: We request types in order based on what we can handle.  Strided, offset, c contiguous.  Assuming that the fortran buffer implements offset or strided we handle it properly.  If they dont implement either of those easy to do things than we would give a type error.

Numpy is a bit more thorough in probing.  But given i cant test it currently i didnt want to implement it.  I only implemented strided because python had the access pattern in docs.  

I debated a bit here as to whether to give it fancy things like transpose or reshape on transfer.  But decided that was the producers job.

test/jpypetest/test_array.py
marscher: what is the intended result of such a memory transfer? You could check, that a strided array created with numpy survives the transfer to Java and back to Python.

Thrameos: Is there a way to create a strided array in numpy?  I was able to create a normal array with different steps by slincing and a c contigous by disabling to hit the fall through.  I just dont know how to hit the others currently.  

Perhaps i should look through numpy tests and see.  Or maybe we need a cpython harness?

native/python/jp_pythontypes.cpp
marscher: I dont think we should try to implement whole of NumPys feature set.

marscher:
There a some merge conflicts now. Sorry, if I didn't spot your recommendation in which order to merge these. If you want I can leave the merging to you and will just leave some review comments in the future.

test/jpypetest/test_array.py
marscher: I think this function can become handy in this case, https://docs.scipy.org/doc/numpy/reference/generated/numpy.lib.stride_tricks.as_strided.html

Thrameos: Okay I will check it out and see if we can complete the test bench for this.  Many thanks.

Thrameos:
All good.  Now that we have the core modules complete we should be getting much fewer conflicts.   I am going to bite off the explicit vs implicit type conversion problem next.   It will be marked as broad pull because I likely have to touch all the converters to give them an extra argument.  But as we have no other broad pull requests in the queue I think it will be safe in any order.

After that I plan to make some adjustments to the gcov coverage pattern.   Most of what it is tagging right now is untested fault paths on exceptions for stuff like out of memory errors or Python failed on a checked type (meaning it can't fail unless the object is really misbehaving).   I have been able to trip some of them by doing stuff like creating a class which implements a function poorly, but others are absolutely mindbendingly hard to hit.  (How to get an object to derive from long but not have an index method?  I couldn't do it so I changed the pattern to numbers)

native/python/jp_pythontypes.cpp
Thrameos: Okay I will leave it as is for now.   If we come up with a better way to construct tests perhaps more probing will be needed.  I think I hit 97% of the use cases for working with libraries like numpy.

PR #611, opened 2020-03-11
Fix for exceptions.  Hot patch to v0.7.2.

### LINKED ISSUE #605: Segmentation fault in JPypeTracer::trace1 (jpype 0.7.2)
I'm experiencing segmentation faults after upgrading from version 0.7.1 to 0.7.2.
The code is quite intricate so I'm not sure I'll be able to provide a minimal example.
The problem seems related to exception handling and the signal is triggered on line 122 of JPypeTracer::trace1:
https://github.com/jpype-project/jpype/blob/602050f5dc77a9fbd734d55ffdc63e989c11b443/native/common/jp_tracer.cpp#L117-L123

This is a partial stack trace, let me know if I can be of more help.
```
#0  0x00007ffff6bf6337 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:55
#1  0x00007ffff6bf7a28 in __GI_abort () at abort.c:90
#2  0x00007fffe9295f69 in os::abort (dump_core=<optimized out>) at /usr/src/debug/java-1.8.0-openjdk-1.8.0.232.b09-0.el7_7.x86_64/openjdk/hotspot/src/os/linux/vm/os_linux.cpp:1572
#3  0x00007fffe94a0136 in VMError::report_and_die (this=this@entry=0x7ffffffedf10)
    at /usr/src/debug/java-1.8.0-openjdk-1.8.0.232.b09-0.el7_7.x86_64/openjdk/hotspot/src/share/vm/utilities/vmError.cpp:1107
#4  0x00007fffe929ffb5 in JVM_handle_linux_signal (sig=11, info=0x7ffffffee1b0, ucVoid=0x7ffffffee080, abort_if_unrecognized=<optimized out>)
    at /usr/src/debug/java-1.8.0-openjdk-1.8.0.232.b09-0.el7_7.x86_64/openjdk/hotspot/src/os_cpu/linux_x86/vm/os_linux_x86.cpp:541
#5  0x00007fffe9293128 in signalHandler (sig=11, info=0x7ffffffee1b0, uc=0x7ffffffee080)
    at /usr/src/debug/java-1.8.0-openjdk-1.8.0.232.b09-0.el7_7.x86_64/openjdk/hotspot/src/os/linux/vm/os_linux.cpp:4552
#6  <signal handler called>
#7  JPypeTracer::trace1 (msg=msg@entry=0x7fffe9f8bab8 "Fatal error in exception handling") at native/common/jp_tracer.cpp:122
#8  0x00007fffe9f64771 in trace (msg=0x7fffe9f8bab8 "Fatal error in exception handling") at native/common/include/jp_tracer.h:84
#9  JPypeException::toPython (this=this@entry=0x17e9dc0) at native/common/jp_exception.cpp:334
#10 0x00007fffe9f529f2 in PyJPModule_rethrow (info=...) at native/python/pyjp_module.cpp:567
#11 0x00007fffe9f48cbe in PyJPMethod_call (self=self@entry=0x7fffc1d6e5c8, args=args@entry=('####',), kwargs=kwargs@entry=0x0)
    at native/python/pyjp_method.cpp:104
#12 0x00007ffff795b443 in PyObject_Call (func=func@entry=<_jpype._JMethod at remote 0x7fffc1d6e5c8>, args=args@entry=('####',),
    kwargs=kwargs@entry=0x0) at /usr/src/debug/Python-3.6.8/Objects/abstract.c:2261
#13 0x00007ffff79fad7a in do_call_core (kwdict=0x0, callargs=('####',), func=<_jpype._JMethod at remote 0x7fffc1d6e5c8>)
    at /usr/src/debug/Python-3.6.8/Python/ceval.c:5120
#14 _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:3404
#15 0x00007ffff7a03c5a in _PyEval_EvalCodeWithName (_co=<code at remote 0x7fffc13a6150>, globals=<optimized out>, locals=<optimized out>, args=<optimized out>, argcount=<optimized out>,
    kwnames=0x0, kwargs=0x7fffc1fb0fe8, kwcount=<optimized out>, kwstep=1, defs=0x0, defcount=0, kwdefs=0x0, closure=(<cell at remote 0x7fffc1c2efa8>,), name=<optimized out>,
    qualname='AnnotationLibrary.keyword_wrapper.<locals>.<lambda>') at /usr/src/debug/Python-3.6.8/Python/ceval.c:4166
#16 0x00007ffff7a048da in fast_function (func=<optimized out>, stack=0x7fffc1fb0fe0, nargs=1, kwnames=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4992
#17 0x00007ffff7a04b63 in call_function (pp_stack=pp_stack@entry=0x7ffffffeede0, oparg=<optimized out>, kwnames=kwnames@entry=0x0) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4872
#18 0x00007ffff79f90a7 in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:3335
#19 0x00007ffff7a0482a in _PyFunction_FastCall (globals=<optimized out>, nargs=<optimized out>, args=<optimized out>, co=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4933
#20 fast_function (func=<optimized out>, stack=0x7fffc1eb21e8, nargs=<optimized out>, kwnames=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4968
#21 0x00007ffff7a04b63 in call_function (pp_stack=pp_stack@entry=0x7ffffffef030, oparg=<optimized out>, kwnames=kwnames@entry=0x0) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4872
#22 0x00007ffff79f90a7 in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:3335
#23 0x00007ffff7a050aa in _PyEval_EvalCodeWithName (qualname=0x0, name=<optimized out>,
    closure=(<cell at remote 0x7fffc1d51bb8>, <cell at remote 0x7fffc1d51be8>, <cell at remote 0x7fffc1d51c18>), kwdefs=0x0, defcount=0, defs=0x0, kwstep=2, kwcount=<optimized out>,
    kwargs=<optimized out>, kwnames=0x7ffff7faa060, argcount=1, args=0x7fffc1eb2060, locals=0x0, globals=<optimized out>, _co=<code at remote 0x7fffea241e40>)
    at /usr/src/debug/Python-3.6.8/Python/ceval.c:4166
#24 PyEval_EvalCodeEx (_co=<code at remote 0x7fffea241e40>, globals=<optimized out>, locals=locals@entry=0x0, args=args@entry=0x7fffba11e5a0, argcount=<optimized out>, kws=0x7ffff7faa060,
    kwcount=kwcount@entry=0, defs=defs@entry=0x0, defcount=0, kwdefs=kwdefs@entry=0x0,
    closure=closure@entry=(<cell at remote 0x7fffc1d51bb8>, <cell at remote 0x7fffc1d51be8>, <cell at remote 0x7fffc1d51c18>)) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4187
#25 0x00007ffff79736b2 in function_call (func=func@entry=<function at remote 0x7fffc1d5c620>, arg=arg@entry=('####',), kw=kw@entry={})
    at /usr/src/debug/Python-3.6.8/Objects/funcobject.c:604
#26 0x00007ffff795b443 in PyObject_Call (func=func@entry=<function at remote 0x7fffc1d5c620>, args=args@entry=('####',),
    kwargs=kwargs@entry={}) at /usr/src/debug/Python-3.6.8/Objects/abstract.c:2261
#27 0x00007ffff79fad7a in do_call_core (kwdict={}, callargs=('####',), func=<function at remote 0x7fffc1d5c620>)
    at /usr/src/debug/Python-3.6.8/Python/ceval.c:5120
#28 _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:3404
#29 0x00007ffff7a03c5a in _PyEval_EvalCodeWithName (_co=<code at remote 0x7fffea216270>, globals=<optimized out>, locals=<optimized out>, args=<optimized out>, argcount=<optimized out>,
    kwnames=0x0, kwargs=0x7fffb9f147e0, kwcount=<optimized out>, kwstep=1, defs=0x0, defcount=0, kwdefs=0x0,
    closure=(<cell at remote 0x7fffc14d13d8>, <cell at remote 0x7fffc14d1e58>, <cell at remote 0x7fffc14d1678>), name=<optimized out>,
    qualname='LibraryKeywordRunner._runner_for.<locals>.<lambda>') at /usr/src/debug/Python-3.6.8/Python/ceval.c:4166
#30 0x00007ffff7a048da in fast_function (func=<optimized out>, stack=0x7fffb9f147e0, nargs=0, kwnames=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4992
#31 0x00007ffff7a04b63 in call_function (pp_stack=pp_stack@entry=0x7ffffffef650, oparg=<optimized out>, kwnames=kwnames@entry=0x0) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4872
#32 0x00007ffff79f90a7 in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:3335
#33 0x00007ffff7a0482a in _PyFunction_FastCall (globals=<optimized out>, nargs=<optimized out>, args=<optimized out>, co=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4933
#34 fast_function (func=<optimized out>, stack=0x7fffc1eaf7e0, nargs=<optimized out>, kwnames=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4968
#35 0x00007ffff7a04b63 in call_function (pp_stack=pp_stack@entry=0x7ffffffef8a0, oparg=<optimized out>, kwnames=kwnames@entry=0x0) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4872
#36 0x00007ffff79f90a7 in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:3335
#37 0x00007ffff7a0482a in _PyFunction_FastCall (globals=<optimized out>, nargs=<optimized out>, args=<optimized out>, co=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4933
#38 fast_function (func=<optimized out>, stack=0x17eaff0, nargs=<optimized out>, kwnames=<optimized out>) at /usr/src/debug/Python-3.6.8/Python/ceval.c:4968
```

Thrameos:
This is a fix for v0.7.2 for the exception issue.  I am posting to to address the issue with the released version.   It should be mergable with the current master with some minor adjustments.

Fixes #605


marscher:
Thanks. Do you think this also applies for exceptions with just one argument, e.g. a message string?

Approved in any case.

Am 12. März 2020 02:55:15 MEZ schrieb Karl Nelson <notifications@github.com>:
>@Thrameos requested your review on: jpype-project/jpype#611 Fix for
>exceptions.  Hot patch to v0.7.2..
>
>-- 
>https://github.com/jpype-project/jpype/pull/611#event-3121131746

-- 

Thrameos:
Yes.  It applies to all.  The pattern had a bug that was creating a bogus exception and then trashing it thus requiring a no argument ctor.  Not sure how i missed this one, except that i faulted by depending too much on java implementations that all met the requirement.

fleiber:
Hi,

Thanks for fixing this indeed very annoying issue. We'll be happy if it could be released quickly in a 0.7.3 :)

Thanks

Thrameos:
I anticipate another release near the end of the month.   I am currently working on my 30 pull requests in 30 nights bug fiesta as we increase our coverage from a very sad 50% to near 90% if I can accomplish it.  I hope that will meet your needs.  (Please also all thank our reviewer marscher)

fleiber:
I've been following a bit your recent work, and indeed you seem to be progressing greatly towards 0.8, which is of particular interest for us because we are trying to optimize the Java/Python memory transfers (well Kotlin/Python to be precise ^^).
But increasing the coverage and improving the stability is also a must have !
Great job, thanks!

marscher:
@Thrameos do we still have an option to perform memory transfers without making copies available, e.g. nio buffers. I think that jparrays are now always taking copies, right?

Thrameos:
Yes.  You can create a direct buffer to a byte array.   But the sort is a bit more complex as we have three separate transfer paths each with different buffering patterns depending on what you want to do.

1) You can construct a memoryview and then ask the memory view to convert to a bytebuffer it should create a shared memory space which can be used for direct transfer.   This is not really intended as a buffering technique but rather was intended to allow the JNI to map shared memory pages.  As it gives you a ByteBuffer rather than an array on the other side, you have to copy it on Java side if you want to do anything with it.   Currently this pattern is a bit counter intuitive to use and we didn't have a test until last month.  I should make this path much more intuitive but making an autoconversion from memory view to Java ByteBuffer so that this can be triggered without special handling.  Second we can make use the to support file sharing of Python files to Java files which we currently only have very basic support.

JNI does not support anything other than bytes to be shared using direct byte buffers.  Therefore, rather than trying to do mixed behaviors all arrays do a copy on bytes we use the same patterns for all array types.   The type of copy is selected based on the support from the JNI.    This gets us to the two other patterns in the code.

2) When doing copies from Python to Java from a memory view we have a different pattern.  We use GetPrimitive locking on transfers from memory views to arrays.   Those should in principle make no more than one copy because you write directly into the Java memory space.  But the JNI controls if it is single/double buffered depending on its state.  Other transfers use GetElements/Release which has the down side that it is a double buffer copy most of the time.   Once to the memory area where JNI does staging and then again to the Java space.   I can try to remove the double copy by updating the pattern accordingly in a few places.

3) Returns from JNI to Python are a different story.  As the Python API does not specify how long it intends to keep the buffer transfers must be made using GetElement/Release pattern which will perform double buffer copies unless the JNI supports pinning.   

Most of this is outside the control of the programmer.  The JNI decides if it wants to support pinning for individual arrays and it can change its mind at any time between get and release calls.  Therefore, we support whatever is going on transparently.  If we can guarantee a short lock period we can use the more aggressive Primitive locking commands.   

Does that clear things up?   (or do you have a feature request that where an alternative pattern may be possible.)

fleiber:
Just checked #630, that's some efficient bug report handling :D

Maybe for this discussion on memory copies I can describe the way we use JPype, it should be a pretty standard use case: we have a data library coded in Kotlin, but as it is used by a wide variety of users, we also provide Python bindings so users can stay in their preferred environment. Data caches are on the Kotlin side, when called from Python we return a double[], which is made available in Python through a Numpy array. The problem of course is the copy done to convert the double[] to the jparray. I thought it would be possible to keep the data in direct buffers in Kotlin, then just transfer the memory address, and wrap it somehow with Numpy, but I'm not sure how this corresponds to what you are describing above...

Thrameos:
Unfortunately, JVM does not support pinning unconditionally thus we can't just map an array from the JVM into the address space of Python and use it for numpy.   We can provide high efficiency methods for transferring memory to and from the JVM.  But as far as I am aware there is no way to make numpy arrays wrap directly.   

It may be possible to make a read only numpy array point to a double[].   I haven't investigated it in depth.  An unpinned array can't be modified and written back more than once (I tried a few times on that front).  So it is pretty limited.

fleiber:
I don't understand, I'm not talking about pinning, which I understand would not be possible (I don't even see how it would properly work with all the different GC): I plan to allocate the caches out of the heap, so they don't move, and I'm handling their life-cycle on the Kotlin side so Python could safely access the pointer?

Thrameos:
The concept of just transfer a "memory address" is pinning if you mean transfer by from JVM to Python in which np.array is backed by JVM.   The specific calls in JNI are in

https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp17314

There is another call getPrimitiveArrayCritical which tries to operated even without pinning support, but is not supposed to traverse a frame boundary.  That would be a "memory address" type transfer but it is very limited.

If I am not understanding your question you will need to be a bit more explicit about what you are thinking about "wrap it somehow with Numpy".

I work primarily on a numerical library myself so I have worked a lot on this pattern but it is always possible that I have missed something.   If I can make read only numpy arrays then it would be possible to pass the array without one of the copies (possibly all if JVM is nice enough to pin it)

Obviously the other type of transfer is shared memory using direct buffers.  But Java does not support a cast operator so the copy ends up in Kotlin rather than in Python.  So no real savings in that route.

Thrameos:
BTW you can handle cross VM reference counting with the RefQueue.   Just tell Kotlin that it is managing a Python object by adding the Python object to the lifespan of a Java Object.   It will add an additional that is removed when the object is GCed.   The only place where we can't support is when you have a true reference loop where something is both holding a Java ptr and that is a holding a Python pointer.

To keep that from happening the exposed API only supports reference queuing on arrays, buffers and proxies.   I may have to reconsider when I have Python objects mapped into Java space.  ( Will be fun when I have to traverse.)

fleiber:
I'm talking about memory allocated out of the heap using Unsafe.allocateMemory/freeMemory. This returns the "real" memory address of the reserved memory block, not some gc-managed heap pointer. This is what I would pass to Python, just a long, not the memory content itself, so I don't see why you are talking about JNI array calls ? This method is not "managed" by the JVM (and its GC), it's a direct system call, so the JVM doesn't have anything to do nor anything to pin (but of course I have the responsability to call free).

Thrameos:
Well the only support that I have currently is JVM or Python controlled memory.  I am not aware of support for managing Unsafe memory as that is not something I can see from either the JVM or Python. 

I can certainly support additional assuming someone is willing to support the work for testing.  Often people ask me about other JVM using languages like Kotlin or Scala, and when you so it is necessary to point me to the relevant docs or I will automatically assume that it is the same as Java.  Hence the sideways answer as it must a be a language feature that I am not aware of.

fleiber:
jdk.internal.misc.Unsafe is a "standard" JDK feature. It's what is used behind the scene when you call ByteBuffer.allocateDirect, which creates a small object implementing ByteBuffer and just holding the memory address. The allocated memory is freed when the ByteBuffer is collected, using the Cleaner class. It's often used when you have a few long-lived large buffers with critical performance requirements. Not being handled by the GC can be an advantage for performance (same as accessing your process memory with a pointer), but creating/destroying them is much slower, you have to be very careful about freeing, you may end up with memory fragmentation, etc. Usual GC pros and cons :)
The memory region is allocated in the process memory space, but using a malloc() system call, so it's outside the heap (which is considered by the system as a large allocated memory block). Anything in the same process can access it, which includes the Python which runs the JVM.
I didn't check, but I'm pretty sure there must be a way to wrap this allocated memory in a Numpy array, that's why I was asking if you had already investigated that direction :)

fleiber:
Googling "Numpy malloc" gives me pages like https://stackoverflow.com/questions/25102409/c-malloc-array-pointer-return-in-cython, because obviously Cython has the same questions, of how to wrap a malloc-allocated memory region in a numpy array. Seems pretty straightforward, but I'm not a Cython expert, I'll have to ask the guy in my team who is :)

Thrameos:
Ah you are talking about internal functions.   They are hard to deal with as the Java module system prevents me from talking to them directly.  (I had a similar issue with caller sensitive which is also an internal API.)  But if I can figure out how to get the pointer, it can be wrapped directly.  Two minor issues that would need to be handled.

1) First it needs a closure like memory handling so that it is clear how keep it alive until the numpy array drops.  I haven't looked into that issue.
2) Second I currently am not supporting linking against the numpy api.  I would need to build two extension modules so that you can use the numpy api directly or some way of dynamically linking which is very painful.  (Cry,  I just got that stuff removed. Sob.)

Thrameos:
Looking at postings on Unsafe... yeah it is as bad as I thought.  You have to manually start the JVM with stuff module protection disarmed or it will prevent access.  It is doable, but that is again something better left to with Project Panama is included.

fleiber:
Yes, I know you removed the Numpy linking, but it did seem cleaner that way, so I support it :)

About Unsafe, that's why I quoted "standard", because it's definitely not some usual Java classes, contrary to what Panama aims to do. But it's internally called everywhere, including in the direct ByteBuffers, so I wouldn't call it unsupported either.

In my case, I don't really need some dedicated JPype functions, I can just return the memory address as a long and call the Numpy wrapping myself, but I was wondering if there was a simpler more Java/JPype way of doing it... because passing memory buffers without copies is certainly a critical requirement for many Python/Java projects.

fleiber:
> Looking at postings on Unsafe... yeah it is as bad as I thought. You have to manually start the JVM with stuff module protection disarmed or it will prevent access. It is doable, but that is again something better left to with Project Panama is included.

No, I think the module issue is something else: this class used to be sun.misc.Unsafe, and was moved to jdk.internal.misc with JDK9/Jigsaw (some wanted to simply remove it, but the perf-critical Java world rebelled), so now you can't access it unless you have a properly modularized project. It's annoying the first time you have to do it, but it's not limited to Unsafe, it's really just the initial pain of modules :)

Thrameos:
Okay I will need some more consideration.   Making custom wrappers is pretty easy internally.   But I don't have a good way of registering externally defined C++ extension wrappers that would make it easy for what you would like to have.   We have "magic" wrappers like "org.jpype.Proxy" which do things like take an long and unpack it to a Python object.   We would just need to generalize it and link it to an np.array to support it.  

The process is currently define a class as "special" in Java org.jpype.manager.TypeManager then placing the custom converter in native/common/jp_typefactory.  You can add custom converters and JNI methods to support exotic access to the object.   If it has to have additional presence on Python it requires either a C module in pyjp_ or it would require backdoor access (like talking to statistics in reference queue.)

Does that make it more clear?

Thrameos:
(BTW once we drop Java 8 support it would be great is some sends a PR to help get our internal stuff to be a proper module.)

Thrameos:
I have an idea (perhaps a not a good one).  It is pretty doubtful that I can do the required work to get you access to Unsafe right now because the amount of APIs, security layers, etc.   But what I can do is get you an API on the Java side to call ``MemoryView.create(long memory, int dims, int[] shape, char type, Object owner)``.   You stuff in your unsafe long (by whatever vodo was required to get it), and the corresponding information and I will get you an object that magically transforms into a memory view on the Python side.   We can then ask numpy to wrap it and therefore provide full access to shared space.  I will hold a Global reference to the owner so long as Python has the view object checked out.   We may need a bit of work on the memory management but I think it would satisfy the basic requirements.

fleiber:
I think we can make it simpler by just using Buffer: on the Java side you have some wrapper which would just hold the Buffer and shape, then in Python you should be able to properly create the memory view.

- You can use the Buffer type to determine the data type (for instance DoubleBuffer -> double64).
- If it's a heap buffer you would do your usual JNI array calls (with Buffer.array() which gives access to the internal array to avoid an additional copy)
- If it's direct buffer (Buffer.isDirect() to know, then Buffer.base() to retrieve the low-level DirectByteBuffer) then you extract the address.
- Not sure what you should do for mmaped buffers, maybe just retrieve the file descriptor and provide a Python memory view based on it...
- Buffers are not the most recent interface, but their API is pretty clean, and handles read-only mode, slicing, marking, flipping, etc, so additional requests can be easily done.

This way seems very Java-natural, with all the dirty and difficult operations hidden behind the standard java Buffer classes, and pretty straightforward in JPype. What do you think?

Thrameos:
It unfortunately has to be a special type rather than Java standard type.   We don't want to magic wrapper something that use is likely to use.   Buffers are often used in normal programming and thus we can't have a user we working with a Buffer for normal programming and have it magically change to the memory view. 

Just to be clear what the conversion process is: Every time we get an Java object back from a method or field we check for the C++ wrapper class using the object type.  In most cases this wrapper class just creates a Python wrapper object that holds reference to the Java object.  But some are "special" (Number, Box types, Throwable, String, Array) which builds a Python wrapper which is derived from a different Python base object (so it can have Python required behaviors like can be caught).  These still obey all the rules of the normal wrappers but they have a little extra.  But it can also be "magic" such as org.jpype.proxy.JPypeProxy which instead probes the object to retrieve a jlong that it reinterpret casts to pointer and thus can return the Python object rather and any type of wrapper.  Magic wrappers completely hide the Java object from Python as once the conversion process is done the handle to the Java object is lost (thus preventing reference loops and other bad things.)  In the case of the proxy object which is an invocation handler pointing back to Python and being a foreign element to Java, this make perfect sense.  Thus if you have some object which is foreign to Java (say an Unsafe buffer from a native C library), then the magic wrapper can unpack it to C internals rather than just a reference back Java. 

I can make buffers memory viewable like arrays pretty easily (though you can do that already by calling ``memoryview(buffer.array()`` but that will again be read only access because of JNI limits.  DirectBuffers that are shared (read and write) are currently support, but only if the memory was created on the Python side.  To really support a please take this object and create a memory view in Python from Java, we would need a marker that something has an Unsafe access and some way to retrieve the point, shape, and datatype if you want to be able to get direct access, and it must not be a class that is usable for other purposes so that it doesn't matter if we can't get a handle to it.

But if you have some Java code that can pull off that stunt with a regular buffer, then we do have another option.  While I can't magic wrap a normal Java type (as it hides the Java object making it inaccessible to Python) I can always add utility functions to create a view on the Python side.  It would require explicit action on the users part to get to the view, but it would serve much the same function.  

I hope this clears things up.

fleiber:
I didn't say we should directly pass the Buffer, I said we should have a small wrapper class, with the buffer and some metadata including the shape: it wouldn't interfere with those who want to work with Buffers as today (because you are right we don't want any magic on standard Java types), but it solves all use cases, it's almost no work on the Java side (since it's reusing the existing Buffers), and the Python side can do the necessary magic to make it appear as a memory view.

The solution with utility functions on the Python side, taking a Java Buffer and shape, also looks good, it's similar to what I was going to do on my side (retrieve pointer + shape in Python, then wrap it as a memory view or directly to a numpy array).

Thrameos:
Do you have some code that will get the void* pointer from a Buffer?   I thought that Buffers can have many types of backers.

Thrameos:
Okay it looks like the JNI call GetDirectBufferAddress may give what is needed.  I will run some tests to see if I can create a float buffer that is backed by direct access.

Thrameos:
I believe I have enough information to implement a Buffer which will create a memory view on request.   Estimated effort will be 8 hours.   Lets see how this goes.

Thrameos:
Your pull request is complete in #639.

fleiber:
Wow that was quick!
Let me have a look, keep you posted (not much time available today) :)

fleiber:
I asked someone in my team to test, and it looks good on buffers!
But your buffer2 branch crashes, as it does on the master (was working on 0.7.1/2): https://github.com/Thrameos/jpype/issues/43

Thrameos:
Thanks I will look it over this evening (9 hr from now)

PR #617, opened 2020-03-14
Support for java.lang.Number

### LINKED ISSUE #616: Support for java.lang.Number
java.lang.Number should have the same rules as found in Java.   

Python int -> java.lang.Long
Python float -> java.lang.Double
JByte, JShort, JInt, JLong, JFloat, JDouble - > to boxed types

If possible get the size of different types from numpy appropriately.
np.int8, np.int16, np.int32, np.int64 -> to boxed types

If possibly including Java scale up rules for unsigned if possible.
np.uint8 -> java.lang.Short
np.uint16 -> java.lang.Integer
np.uint32 -> java.lang.Long

Reference
```
  public static Number callNumber(Number n)
  {
    return n;
  }

    System.out.println(callNumber(1).getClass());
    System.out.println(callNumber(1l).getClass());
    System.out.println(callNumber(1.0).getClass());
    System.out.println(callNumber(1.0f).getClass());
    System.out.println(callNumber((byte) 1).getClass());
    System.out.println(callNumber((short) 1).getClass());
    System.out.println(callNumber((int) 1).getClass());
    System.out.println(callNumber((long) 1).getClass());
    System.out.println(callNumber((float) 1).getClass());
    System.out.println(callNumber((double) 1).getClass());
```

Gives 

```
class java.lang.Integer
class java.lang.Long
class java.lang.Double
class java.lang.Float
class java.lang.Short
class java.lang.Byte
class java.lang.Integer
class java.lang.Long
class java.lang.Float
class java.lang.Double
```

Current behavior is 
```
raceback (most recent call last):
  File "test.py", line 12, in <module>
    fixture.callNumber(JInt(1))
TypeError: No matching overloads found for jpype.common.Fixture.callNumber(JInt), options are:
        public static java.lang.Number jpype.common.Fixture.callNumber(java.lang.Number)
```

Thrameos:
In today's pull request, we add support for java.lang.Number so that it auto boxes with the same rules as Java.  We also box numpy types to the appropriate Java box sized rather than always going to Long or Double.

I am not particularly happy with having to compare strings to support numpy types, but I could not find any way to determine the size other than using the name.  But this way we don't need to link to numpy.

Fixes #616

### DIFF
--- a/native/common/jp_classhints.cpp
+++ b/native/common/jp_classhints.cpp
@@ -405,6 +405,39 @@ public:
 	}
 } _javaObjectAnyConversion;
 
+class JPConversionJavaNumberAny : public JPConversion
+{
+public:
+
+	virtual JPMatch::Type matches(JPMatch &match, JPJavaFrame *frame, JPClass *cls, PyObject *pyobj) override
+	{
+		JP_TRACE_IN("JPConversionJavaNumberAny::matches");
+		JPValue *value = PyJPValue_getJavaSlot(pyobj);
+		if (value == NULL || frame == NULL || value->getClass() == NULL
+				|| value->getClass() == frame->getContext()->_boolean
+				|| value->getClass() == frame->getContext()->_char)
+			return match.type = JPMatch::_none;
+		match.conversion = &_javaObjectAnyConversion;
+		JPClass *oc = value->getClass();
+		if (oc == NULL)
+			return match.type = JPMatch::_none;
+		if (oc->isPrimitive())
+			return match.type = JPMatch::_implicit;
+		if (oc == cls)
+			return match.type = JPMatch::_exact;
+		bool assignable = frame->IsAssignableFrom(oc->getJavaClass(), cls->getJavaClass()) != 0;
+		return match.type = (assignable ? JPMatch::_implicit : JPMatch::_none);
+		JP_TRACE_OUT;
+	}
+
+	virtual jvalue convert(JPJavaFrame *frame, JPClass* cls, PyObject* pyobj) override
+	{
+		jvalue  v;
+		v.l = 0;
+		return v;
+	}
+} _javaNumberAnyConversion;
+
 class JPConversionJavaValue : public JPConversion
 {
 public:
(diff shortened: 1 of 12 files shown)


test/jpypetest/test_number.py
```diff
@@ -0,0 +1,40 @@
...
+
+class JNumberTestCase(common.JPypeTestCase):
+    def setUp(self):
+        common.JPypeTestCase.setUp(self)
+        self.cls = JClass("jpype.common.Fixture")
+        self.fixture = self.cls()
+
+    def testJavaPrimitives(self):
+        self.assertIsInstance(self.fixture.callNumber(JByte(1)), java.lang.Byte)
+        self.assertIsInstance(self.fixture.callNumber(JShort(1)), java.lang.Short)
```
marscher: just for the sake of completeness, we seem to lack JCharacter?

Thrameos: Characters are not numbers as they have encodings.  There are places where we treat them as unsigned shorts but that is mostly incorrect.   So as you can see from the language definition neither Boolean nor Character are derived from Number

https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html

native/common/jp_objecttype.cpp
```diff
@@ -82,6 +53,17 @@ JPMatch::Type JPObjectType::getJavaConversion(JPJavaFrame* frame, JPMatch& match
 		return match.type = JPMatch::_implicit;
 	}
 
+	// Apply user supplied conversions
+	if (!m_Hints.isNull())
+	{
+		JPClassHints *hints = ((PyJPClassHints*) m_Hints.get())->m_Hints;
```
marscher: This seems ugly. Why do you need to cast and then access a private member?

native/common/jp_classhints.cpp
```diff
@@ -405,6 +405,39 @@ class JPConversionJavaObjectAny : public JPConversion
...
+			return match.type = JPMatch::_exact;
+		bool assignable = frame->IsAssignableFrom(oc->getJavaClass(), cls->getJavaClass()) != 0;
+		return match.type = (assignable ? JPMatch::_implicit : JPMatch::_none);
+		JP_TRACE_OUT;
+	}
+
+	virtual jvalue convert(JPJavaFrame *frame, JPClass* cls, PyObject* pyobj) override
+	{
+		jvalue  v;
+		v.l = 0;
```
marscher: I dont understand, I thought we want to convert the Java number. So why this is alsways zero?

native/common/jp_objecttype.cpp
Thrameos: We can make an accessor.  PyJPClassHints is being stored as a PyObject because we need to hold a reference count.  But we have to get to the C++ element in order to make the conversion.  This pattern was improved in numbers2 PR.

native/common/jp_classhints.cpp
Thrameos: This one is a wierd edge case.  We do the match here but there is no convert method, but because it must be implemented because it is a pure virtual function we have to fill in a stub method.   If you read a few lines up you will see that it instead points to javaObjectAny for the actually conversion.  

I will switch this to be derived from another class so that it has better clarity.

Thrameos: I correct this in numbers2 by changing the base type to JavaObjectAnyConversion so that the stub method can be dropped.

test/jpypetest/test_number.py
marscher: I missed that important difference.

native/common/jp_objecttype.cpp
marscher: Thanks

PR #618, opened 2020-03-14
Fix warning on lgtm

Thrameos:
This PR attempts to address remaining warning on LGTM.


marscher:
Thanks!

Thrameos:
YES! AT LAST WE GET AN A!

PR #619, opened 2020-03-14
More coverage

Thrameos:
Yet another coverage pull request.

### DIFF
--- a/native/common/jp_field.cpp
+++ b/native/common/jp_field.cpp
@@ -49,12 +49,6 @@ void JPField::setStaticField(PyObject *pyobj)
 {
 	JP_TRACE_IN("JPField::setStaticAttribute");
 	JPJavaFrame frame(m_Class->getContext());
-	if (isFinal())
-	{
-		stringstream err;
-		err << "Field " << m_Name << " is read-only";
-		JP_RAISE(PyExc_AttributeError, err.str().c_str());
-	}
 	m_Type->setStaticField(frame, m_Class->getJavaClass(), m_FieldID, pyobj);
 	JP_TRACE_OUT;
 }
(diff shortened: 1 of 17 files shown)


native/common/jp_field.cpp
```diff
@@ -49,12 +49,6 @@ void JPField::setStaticField(PyObject *pyobj)
 {
 	JP_TRACE_IN("JPField::setStaticAttribute");
 	JPJavaFrame frame(m_Class->getContext());
-	if (isFinal())
```
marscher: This condition could never happen?

test/jpypetest/test_jobject.py
```diff
@@ -166,3 +167,34 @@ def testObjectDouble(self):
...
+    def testObjectField(self):
+        with self.assertRaises(TypeError):
+            self.fixture.object_field = object()
+        with self.assertRaises(TypeError):
+            self.fixture.static_object_field = object()
+
+    def testArraySetRangeFail(self):
+        ja = JArray(JObject)(4)
+        with self.assertRaises(TypeError):
+            ja[:] = [ 1, 2, object(), 3]
```
marscher: pep8 pls (extra space in beginning of list definition)

native/common/jp_field.cpp
Thrameos: This was a problem with responsibilities.  This condition was never hit  and in tracing I discovered that the PyJPField had already checked against this condition.  Thus this check was redundant.

test/jpypetest/test_jobject.py
Thrameos: No problem.  I will pep8 prior to merge.

native/common/jp_field.cpp
marscher: Thanks for the explanation.

PR #621, opened 2020-03-15
Coverage and bug fixes

### LINKED ISSUE #620: Segfault on waiting for GIL
If a GC cycle is started before then end of execution of Python, the reference queue can abort while waiting for the GIL lock to be returned.  There is no obvious workaround at this time other than ignoring the fault.  For some reason this is triggering some kind of exception but I can't figure out its type as of yet.

Thrameos:
In tonight's extended edition, we try to move branch coverage to 80%.   During this process we discover a number of bugs that need correcting.

- GIL segfaults on exit
- C++ exception information not being written out
- uint64 transform to float/double

This PR makes some use of GCOVR_EXCL statements as there are many parts of our code that is purely defensive and cannot be executed in normal operation or is restricted to shutdown and prestart conditions.   

JPPlatform headers were moved to a source file as the level of code in the header was larger than I felt comfortable as a header.   A number of other header functions were moved to the source files so that it better represented a division between inline code and implementations.

This pull request is based on #619 thus it can't be included until the base PR is merged.  Also this likely conflicts with #617 and #618, so all three should be included first.

Fixes #620


jpype/_core.py
```diff
@@ -310,6 +312,17 @@ def shutdownJVM():
...
 
 
+# In order to shutdown cleanly we need the reference queue stopped
+# otherwise, we can experience a crash if a Java thread is waiting
+# for the GIL.
+def _JTerminate():
+    try:
+        _jpype.shutdown()
+    except RuntimeError:
+        pass
```
marscher: exception silencing at this point might be the right thing to do. What kind of error do you expect here and would it be considerable to show them?

native/common/include/jp_class.h
```diff
@@ -192,10 +186,10 @@ class JPClass : public JPResource
 
 	JPContext* getContext() const
 	{
+		// This sanity check is for during shutdown.
 		if (m_Context == 0)
-			JP_RAISE(PyExc_RuntimeError, "Null context");
-		return
-		m_Context;
+			JP_RAISE(PyExc_RuntimeError, "Null context"); // GCOVR_EXCL_LINE
```
marscher: is this the exception you are catching in the atexit handler? If so, I'd prefer a custom exception type (or if you object the overhead, leave it like this).

native/common/include/jp_classhints.h
```diff
@@ -22,10 +22,14 @@ class JPConversion
 public:
 	virtual ~JPConversion();
 
+	// GCOVR_EXCL_START
```
marscher: if the default is never called, why cant it be pure virtual then? Can't it be tested at all?

jpype/_core.py
Thrameos: The only case that yields a runtime exception here is that JVM is already shutdown.   Thus if the user has shutdown and then atexit does as well we get a unnecessary exception.   I considered just passing a flag to shutdown to make it conditional so that we don't have an exception thrown.   

If you would like me to write the alternative (cleaner) solution I can.

jpype/beans.py
```diff
@@ -65,14 +65,14 @@ def _extract_accessor_pairs(members):
         if name == "getClass":
             continue
 
-        if member.isBeanAccessor():
+        if member._isBeanAccessor():
```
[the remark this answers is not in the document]
Thrameos: It was always supposed to be private as it was a method introduced to help beans.  But if you feel it should be exposed I can move it back.

native/common/include/jp_classhints.h
Thrameos: It can be pure virtual, but we have a handful of implementations that are missing a match method because the logic for the match is done in the getJavaConversion method rather than a match (because resources are held privately).

There is no way to hit this method currently.

native/common/include/jp_class.h
Thrameos: Null context should be a SystemError rather than RuntimeError.  There is a screen at the PyJP level to prevent ever getting to the backend when the JVM is not running.  I have been working to correct that issue.

jpype/_core.py
Thrameos: There is one other obnoxious case that I have never resolved which is if someone calls shutdown from within a Proxy call back.   The whole shutdown pattern needs to reworked so that the JVM is in control and the master thread will wait for it to complete.   But that task has always been too much of a lift as it has to touch a lot of code, and the value of supporting it is low.

native/common/include/jp_primitivetype.h
```diff
@@ -49,7 +47,7 @@ class JPPrimitiveType : public JPClass
 			void* memory, int offset) = 0;
 
 	virtual PyObject *newMultiArray(JPJavaFrame &frame,
-			JPPyBuffer& view, int subs, int base, jobject dims);
+			JPPyBuffer& view, int subs, int base, jobject dims) = 0;
```
marscher: pure virtual FTW :+1:

native/common/jp_platform.cpp
```diff
@@ -77,4 +82,74 @@ class Win32PlatformAdapter : public JPPlatformAdapter
 	}
 } ;
 
-#endif // _PLATFORM_WIN32_H_
\ No newline at end of file
+#define  PLATFORM_ADAPTER Win32PlatformAdapter
+#else
+
+//AT's comments on porting:
+// this file should be better called jp_platform_unix.h
```
marscher: this comment is obsolete

Thrameos: Fixed.

Thrameos:
Due to COVID-19 unfortunately my business has shutdown by public health order.  Still waiting on details.  Perhaps that means I can plow through to JPype-0.8.

marscher:
some conflicting files need to be resolved.

jpype/beans.py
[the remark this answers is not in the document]
marscher: I do not think this is used publicly, so it can stay as it is now.

native/common/include/jp_classhints.h
marscher: I see

marscher:
Thanks again for this incredible testing coverage increase. And you squashed so many bugs along this path :+1:

originell:
holy smokes what an increase! tremendous job!

Thrameos:
@originell well 2/3 of the increase was a cheat.   The coverage system was counting all the exception paths as partials because there was a possibility that Python would pass in a null pointer into a reference handle and it would throw which would only happen if something really bad happened.  As the first thing we did when handling a Python object was put it into a reference handle so we could throw a C++ without losing a reference in many places, it was considering all the code as having an untested branch.   This was basically doubling our branch count.

Adding the flag ``--exclude-thrown-branches`` which the documentation indicates would produce more "sensible" branch coverage (not my word, but theirs) fixed most of our bad coverage problems.

PR #624, opened 2020-03-16
Numbers2

Thrameos:
This pull require is the ground work for further improvements in the type system.  In particular, we need to be able to control when a match requires checking and when it does not.  The current system with just a function and no communication between the caller and conversion other than through the argument list does not support that.  

This pull request is completely under the hood, but is a prerequisite for a lot minor improvements, and once merged will allow all the rest of the PRs to be reviewed cleanly.

- Fixes a nasty bug with the order that special C++ wrappers were created that was causing duplicates for Number and CharSequence
- Adds a closure and slot cache for match so that type lookups only occur once per conversion with time saving when matching against overloads.
- Corrects warnings from Visual studio.
- Renamed getJavaConversion to findJavaConversion and removed all of the arguments in place of the single match structure.  This was both a style and performance change.  Passing a huge list of arguments over and over was both repetitive and a waste of CPU.  Thus now everything is stored in match and the function was renamed to better represent the function.  It should look a lot cleaner.

### DIFF
--- a/jpype/_jclass.py
+++ b/jpype/_jclass.py
@@ -88,7 +88,6 @@ class JClass(_jpype._JClass, metaclass=JClassMeta):
             if len(params) > 0:
                 params = params.split(',')
             if len(params) > 0 and len(params) != len(ret.class_.getTypeParameters()):
-                print(params, len(params))
                 raise TypeError(
                     "Java generic class '%s' length mismatch" % (ret.__name__))
             return ret
--- a/native/common/include/jp_match.h
+++ b/native/common/include/jp_match.h
@@ -14,15 +14,10 @@ public:
 		_implicit = 2,
 		_exact = 3
 	} ;
+
 public:
 	JPMatch();
 	JPMatch(JPJavaFrame *frame, PyObject *object);
-
-	JPMatch::Type type;
-	JPConversion *conversion;
-	JPJavaFrame *frame;
-	PyObject *object;
-	JPValue *slot;
 
 	JPContext *getContext()
 	{
@@ -39,17 +34,23 @@ public:
 	 * @return the Java slot or 0 if not available.
 	 */
 	JPValue *getJavaSlot();
+
+	jvalue convert();
+
+public:
+	JPMatch::Type type;
+	JPConversion *conversion;
+	JPJavaFrame *frame;
+	PyObject *object;
+	JPValue *slot;
+	void *closure;
 } ;
 
 class JPMethodMatch
 {
 public:
-	JPMatch::Type type;
-	bool isVarIndirect;
-	JPMethod* overload;
-	char offset;
-	char skip;
-	std::vector<JPMatch> argument;
+
+	JPMethodMatch(JPJavaFrame &frame, JPPyObjectVector& args);
 
 	JPMatch& operator[](size_t i)
 	{
@@ -61,19 +62,13 @@ public:
 		return argument[i];
 	}
 
-	JPMethodMatch(JPJavaFrame &frame, JPPyObjectVector& args)
-	: argument(args.size())
-	{
-		type = JPMatch::_none;
-		isVarIndirect = false;
-		overload = 0;
-		offset = 0;
-		skip = 0;
-		for (size_t i = 0; i < args.size(); ++i)
-		{
-			argument[i] = JPMatch(&frame, args[i]);
-		}
-	}
+public:
+	JPMatch::Type type;
+	bool isVarIndirect;
+	JPMethod* overload;
+	char offset;
+	char skip;
+	std::vector<JPMatch> argument;
 } ;
 
 #endif /* JP_MATCH_H */
--- a/native/common/include/jp_match.h
+++ b/native/common/include/jp_match.h
@@ -15,8 +15,20 @@ public:
 		_exact = 3
 	} ;
 public:
+	JPMatch();
+	JPMatch(JPJavaFrame *frame, PyObject *object);
+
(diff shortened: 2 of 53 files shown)


marscher:
The style of PRs based upon another base PR becomes unfeasible to review in a sane amount of (free) time. It produces lots of unreleated changes in the review pane and conflicts.
I suggest that we create a base branch (e.g. numbers2) and you open PRs based on this branch (which can live in this repo).

Thrameos:
Fully agree.  I will make it happen.

Thrameos:
Though in fairness, when the master is updated all those commits will disappear from the other PRs.  And you will notice that I don't hit request review on a PR until time catches back up with the master. For example once this is merged the next 4 reviews will suddenly drop all the commits except the last.  If not then I just freshen the PR and then hit the review button.

If you keep jumping ahead in the book, you are going to spoil the ending.

jpype/_jclass.py
```diff
@@ -88,7 +88,6 @@ def __new__(cls, jc, loader=None, initialize=True):
             if len(params) > 0:
                 params = params.split(',')
             if len(params) > 0 and len(params) != len(ret.class_.getTypeParameters()):
-                print(params, len(params))
                 raise TypeError(
```
Thrameos: Stray debugging line was removed

native/common/include/jp_match.h
```diff
@@ -14,20 +14,43 @@ class JPMatch
...
+	jvalue convert();
+
 public:
 	JPMatch::Type type;
-	JPConversion* conversion;
+	JPConversion *conversion;
+	JPJavaFrame *frame;
+	PyObject *object;
+	JPValue *slot;
+	void *closure;
```
Thrameos: This is the heart of the PR as we are going to push a bunch of the brains back into the match.  

The frame and the object were basically being passed at every call when they were static to the match after creation.   
In addition every method was calling on the slot once in the match and then again in the convert method.  When we were doing a large overload we would get 5 or more times.  Thus caching it once is a performance issue.

Last the match often has something that needs to be passed to the convert (a class, proxy, or type) so we add a user defined pointer as the closure.

native/common/jp_arrayclass.cpp
```diff
@@ -35,34 +35,16 @@ JPArrayClass::~JPArrayClass()
...
-		{
-			m_ComponentType->getJavaConversion(frame, imatch, seq[i].get());
-			if (imatch.type < match.type)
-			{
-				match.type = imatch.type;
-			}
-		}
-		match.conversion = sequenceConversion;
-		return match.type;
-	}
```
Thrameos: These functional checks were moved to become the match method for the JPConversion, thus you will see a lot of displacements like this.

native/common/jp_booleantype.cpp
```diff
@@ -46,31 +46,40 @@ class JPConversionAsBoolean : public JPConversion
...
+	}
+
+	virtual jvalue convert(JPMatch &match) override
 	{
 		jvalue res;
-		res.z = JPPyLong::asLong(pyobj) != 0;
+		jlong v = PyLong_AsLongLong(match.object);
+		if (v == -1)
+			JP_PY_CHECK();
+		res.z = v != 0;
```
Thrameos: I am continuing to remove some of the minor utility functions from the Python layer.  When I rewrote the ruby/python layer there was a lot of doubly indirect stuff.  It served a purpose when we were still supporting Python2 and there were actually differences, but now that it is dropped I have been pushing those back as it just makes the code easier to follow.

native/common/include/jp_primitive_accessor.h
```diff
@@ -252,9 +269,9 @@ class JPConversionFloatWiden : public JPConversion
 {
 public:
 
-	virtual jvalue convert(JPJavaFrame *frame, JPClass *cls, PyObject *pyobj) override
+	virtual jvalue convert(JPMatch &match) override
 	{
-		JPValue *value = PyJPValue_getJavaSlot(pyobj);
+		JPValue *value = PyJPValue_getJavaSlot(match.object);
```
Thrameos: This should be using the cache.  Please fix it, you slacker.

native/common/jp_classhints.cpp
```diff
@@ -21,6 +21,55 @@
...
+	object = obj;
+	type = JPMatch::_none;
+	slot = (JPValue*) - 1;
+	closure = 0;
+}
+
+JPValue *JPMatch::getJavaSlot()
+{
+	if (slot == (JPValue*) - 1)
+		return slot = PyJPValue_getJavaSlot(object);
```
Thrameos: This is one part that I don't really like.  I need to be able to check if the slot has been checked and I am not willing to make another bool (which will cost a whole word because of alignment) to just indicate if the slot has been checked.   Thus we have a 3 state value (-1 not checked, 0 no java slot, and otherwise) has value.   

In the horribly unlikely scenario that the java slot is located at the unaligned address -1 we will look it up multiple times taking a performance hit.

native/common/include/jp_primitive_accessor.h
marscher: :laughing:

marscher:
apache.org is terribly slow, 8mb in 20 minutes. I'm going to skip the rest of the Appveyor jobs.

Thrameos:
Okay I admit it... I was wrong, the github system won't fast forward a PR when the master catches up.   I must have dreamed it (partly the result of doing work at 5 AM)....  either way all the PRs are now rebased and thus we have caught back up to the present.

PR #626, opened 2020-03-17
C++ stacktrace enable

### LINKED ISSUE #623: C++ Exceptions optional
Although C++ exceptions are very useful for tracing the source of an error in C++ code, it is likely going to be a distraction for the user.   Perhaps we should have a switch to enable this feature as it is mostly for debugging.

Thrameos:
This pull request adds a switch for enabling extended C++ exception stacktraces.   It is a private internal module function.  It will be enabled during testing (as it makes tracking problems in the CI much easier).

This is built on numbers2.  (It is only a 5 line change)

Fixed #623

### DIFF
--- a/jpype/_jcollection.py
+++ b/jpype/_jcollection.py
@@ -122,6 +122,7 @@ class _JMap(object):
     This customizer adds the Python list and len operators to classes
     that implement the Java Map interface.
     """
+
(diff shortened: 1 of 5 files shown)


Thrameos:
I will have to deal with this one once the master catches up with this point.

PR #627, opened 2020-03-17
Shutdown bug

### LINKED ISSUE #625: Extraneous Runtime error with atexit
The atexit solution works for well behaved code (and prevents a crash if there is something waiting in the reference queue.)  But it added one weird error at exit time if there is an active exception on the stack at the end of the program.

```
Exception ignored in tp_clear of: <class 'list'>
RuntimeError: Java Virtual Machine is not running
```

Likely this is a problem with dead resources being clear.  I will need to isolate the problem so that it can be addressed.

Thrameos:
This prevents an exception from being issued with ReleaseBuffer or freeing of resource left over after the JVM is shutdown.  There may be other bugs like this found.

This is built on #624

Fixes #625

### DIFF
--- a/native/python/pyjp_array.cpp
+++ b/native/python/pyjp_array.cpp
@@ -271,7 +271,14 @@ static int PyJPArray_assignSubscript(PyJPArray *self, PyObject *item, PyObject *
 static void PyJPArray_releaseBuffer(PyJPArray *self, Py_buffer *view)
 {
 	JP_PY_TRY("PyJPArrayPrimitive_releaseBuffer");
-	JPContext *context = PyJPModule_getContext();
+	JPContext* context = JPContext_global;
+	if (!context->isRunning())
+	{
+		delete self->m_View;
+		self->m_View = NULL;
+		return;
+	}
+	//	JPContext *context = PyJPModule_getContext();
 	JPJavaFrame frame(context);
 	if (self->m_View == NULL || !self->m_View->unreference())
 		return;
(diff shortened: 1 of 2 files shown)


native/python/pyjp_array.cpp
```diff
@@ -271,7 +271,14 @@ static int PyJPArray_assignSubscript(PyJPArray *self, PyObject *item, PyObject *
...
 	JP_PY_TRY("PyJPArrayPrimitive_releaseBuffer");
-	JPContext *context = PyJPModule_getContext();
+	JPContext* context = JPContext_global;
+	if (!context->isRunning())
+	{
+		delete self->m_View;
+		self->m_View = NULL;
+		return;
+	}
+	//	JPContext *context = PyJPModule_getContext();
```
Thrameos: This resolves the issue, but leaves one hole.   If there is a memory buffer left open like this and the JVM is gone then accessing the memory will segfault.   I looked through everything I could finding in the Python docs and it looks like the never considered the case of a memory buffer that they can't hold open with reference counting.

native/python/pyjp_value.cpp
```diff
@@ -140,10 +140,12 @@ void PyJPValue_finalize(void* obj)
 	JPValue* value = PyJPValue_getJavaSlot((PyObject*) obj);
 	if (value == NULL)
 		return;
+	JPContext *context = JPContext_global;
+	if (context == NULL || !context->isRunning())
+		return;
```
Thrameos: If the JVM is gone we don't need to dereference, so no need to to proceed further.

native/python/pyjp_array.cpp
marscher: Mhm, who allocated this buffer in first place?

Thrameos: Allocated pyjp_array.cpp at line 314 and line 361 on getBuffer methods.

PR #628, opened 2020-03-17
Pure virtual FTW

Thrameos:
This is just a style patch trying to improve clarity on conversions.  

Builds on numbers2.

### DIFF
--- a/native/common/include/jp_classhints.h
+++ b/native/common/include/jp_classhints.h
@@ -22,14 +22,7 @@ class JPConversion
 public:
 	virtual ~JPConversion();
 
-	// GCOVR_EXCL_START
-	// This is the default method used for any matches that have explicit
-	// type dependent checks.
-	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls)
-	{
-		return JPMatch::_none;
-	}
-	// GCOVR_EXCL_STOP
+	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls) = 0;
 
 	virtual jvalue convert(JPMatch &match) = 0;
 } ;
--- a/native/common/include/jp_primitive_accessor.h
+++ b/native/common/include/jp_primitive_accessor.h
@@ -206,6 +206,11 @@ class JPConversionLongWiden : public JPConversion
 {
 public:
 
+	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls) override
+	{
+		return JPMatch::_none;  // GCOVR_EXCL_LINE not used
+	}
+
 	virtual jvalue convert(JPMatch &match) override
 	{
 		JPValue *value = match.getJavaSlot();
@@ -268,6 +273,11 @@ template <typename base_t>
 class JPConversionFloatWiden : public JPConversion
 {
 public:
+
+	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls) override
+	{
+		return JPMatch::_none;  // GCOVR_EXCL_LINE not used
+	}
 
 	virtual jvalue convert(JPMatch &match) override
 	{
--- a/native/common/jp_boxedtype.cpp
+++ b/native/common/jp_boxedtype.cpp
@@ -46,7 +46,7 @@ JPMatch::Type JPBoxedType::findJavaConversion(JPMatch &match)
 	if (m_PrimitiveType->findJavaConversion(match) != JPMatch::_none)
 	{
 		JP_TRACE("Primitive", match.type);
-		match.conversion = boxConversion;
+		match.conversion = boxBooleanConversion;
 		match.closure = this;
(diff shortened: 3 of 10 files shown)


native/common/include/jp_classhints.h
```diff
@@ -22,14 +22,7 @@ class JPConversion
...
 
-	// GCOVR_EXCL_START
-	// This is the default method used for any matches that have explicit
-	// type dependent checks.
-	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls)
-	{
-		return JPMatch::_none;
-	}
-	// GCOVR_EXCL_STOP
+	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls) = 0;
```
Thrameos: We get pure virtual here.

native/common/jp_boxedtype.cpp
```diff
@@ -46,7 +46,7 @@ JPMatch::Type JPBoxedType::findJavaConversion(JPMatch &match)
 	if (m_PrimitiveType->findJavaConversion(match) != JPMatch::_none)
 	{
 		JP_TRACE("Primitive", match.type);
-		match.conversion = boxConversion;
+		match.conversion = boxBooleanConversion;
```
Thrameos: But it generates one weirdness here because we can't call the boxed conversion directly.

native/common/jp_classhints.cpp
```diff
@@ -663,10 +605,77 @@ class JPConversionBoxDouble : public JPConversion
...
-		return _boxConversion.convert(match);
+		return JPConversionBox::convert(match);
 	}
 } _boxDoubleConversion;
 
+class JPConversionJavaObjectAny : public JPConversionBox
+{
+public:
+
+	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls) override
```
Thrameos: This block has to slide down for declaration order considerations.

native/common/include/jp_primitive_accessor.h
```diff
@@ -206,6 +206,11 @@ class JPConversionLongWiden : public JPConversion
 {
 public:
 
+	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls) override
+	{
+		return JPMatch::_none;  // GCOVR_EXCL_LINE not used
+	}
+
```
Thrameos: But we end up paying for the pure virtual by requiring two extra stub methods here and in the next template because these require the type dependent switch statement and thus sadly cant become match methods.

native/common/jp_boxedtype.cpp
marscher: Doesn't the conversion implement all abstract methods? What's the compilers message (if you wouldn't mind, I'm just curious).

Thrameos: It gives the usual message about cannot create an instance of an object with a virtual method.  The problem is of course that some of the JConverters have match defined (when the convert will be something else, but we can derived in that case), some have convert defined (where the match rules require extra arguments that are special to the calling type), and most have both.  Something like...

```
native/common/jp_chartype.cpp:45:7: note:   because the following virtual functions are pure within ‘JPConversionAsChar’:
```

marscher: I'm scratching my head right now, if we really need derivation and polymorphic lookups here. Why can't we just use templates?

But don't waste more time on this, it is almost perfect now.

Thrameos: I managed to get templates for the Float and Long patterns.   But the Object converters are a mess because the match rules are often very complete.  I will merge this PR for now and take another run at making it cleaner after I have bit off a few of the bigger ones.

Thrameos:
At some point we will need to consider making a release.  I still have about 5 items sitting on my outgoing queue that warrant inclusion.

- [ ] Fix implicit/explicit conversions so that we don't get out of range unnecessarily.
- [ ] Rework JPackage and unify it with jpype imports so that both of them support checking if something is actually a path.  (I finally got the Java side working so now it is just cleanup in Python)
- [ ] Deprecate the use of Python types as type specifiers for ``JObject(obj, type)``.  I am not sure why this was in the original design, but it is going to make it hard to support the script engine form where you actual want a Python object to go to Java.   The problem cases are JObject(obj, int) and JObject(obj, float) which are in conflict. I don't see why something that doesn't control the casting at all was ever included or even useful.  
- [ ] One final coverage pass to push at the remaining paths.  These things are surprisingly hard to do at times.  Carpet bombing the test system only gets you so far.  But there are always edge cases that require seriously pretzel like convolutions to actually snake a test into hitting.   I suspect some of them are double responsibility where the reason we don't hit them is something else has already taken care of it.
- [ ] Take another shot at the GC linkage PR

Assuming you have no objections to the direction I will try to target a release for Friday next week,  Though we may want to run a week shake down it operations before the release.

PR #635, opened 2020-03-20
Attempt to remove ant for CI

Thrameos:
Tonight's edition is an attempt to make the CI more stable by using maven rather than the local package manager to pull some requirements.  Perhaps this will be more stable.


marscher:
Good job. Eventually it would have been easier to just set a better mirror for the ant download, rather than writing so much wrapper code. Personally I think maven is a overkill tool in the sense how many dependencies it needs to fetch (+the ugliness of XML). If you really observe a time/stability benefit of this, I can approve it.

Do you need the ant python wrapper just for the classpath separator?

Thrameos:
The main benefit of this is the osx image which was taken 5 plus minutes because the need to update homebrew.  Ant is our worst dependency because we need something different to get it on each system.   


I may try to make it build with just javac (no ant)  if ant isn't found.  It can't be that hard.

Sadly that path sep requires me to write a wrapper like that for all my deployment tools either by having a .bat and .sh or doing a .py version.  Why must systems have such trivial differences and no probe method?

Thrameos:
Yes maven is a huge overkill tool.  I could just use maven to do everything but as an overkill tool it likes to completely take over the project layout.  Thus my only use of it is a fancy dependency manager for my ant projects.

setupext/test_java.py
```diff
@@ -27,6 +28,8 @@ def run(self):
         distutils.log.info("Building Java testbench")
         buildXmlFile = os.path.join("test", "build.xml")
         command = [self.distribution.ant, '-f', buildXmlFile]
+        if self.distribution.ant.endswith(".py"):
+            command.insert(0, sys.executable)
```
marscher: wouldn't it be easier to set ant.py executable with a shebang? Dunno if this works on windows though (today I'm gonna remove every trace of this OS from my personal computer *happy*, farewell win7)

marscher: https://docs.python.org/3/using/windows.html#shebang-lines

Thrameos: Well given every time I try to do anything on the CI it burns 2-4 hours due to having to replicate the problem without terminal, lets just remove ant from the equation entirely.   We still need to keep the ant files as they are used by LGTM and some other tools, but there is no reason to require them to build.

Thrameos:
Merging this one so that I can get the CI working in the other pull requests.

marscher:
On 3/22/20 7:48 PM, Karl Nelson wrote:
> Merged #635 into master.
> 

One day we should consider switching to Azure pipelines in favor of two 
providers. It also provides all three major platforms and would unify 
the configuration (+likely speed benefit for windows builds).

Thrameos:
Yep.  Definitely on the list of things to do.  But still so many bug to squash.... just keep squashing... just keep squashing.

PR #636, opened 2020-03-21
Explicit cast

### LINKED ISSUE #602: Explicit casts should not trigger OverflowError
We should limit the OverflowError to calls in which the automatic conversion is applied.  Direct casting should bypass the check so that it acts just like Java where explicit casts are performed.

Thrameos:
Tonight's PR deals with explicit casting to primitives.  In Java casting will never produce an overflow error, thus we have replicated that behavior.  Implicit conversions still produce OverflowError as per Python conventions.   This was intended to be a very small pull request but unfortunately the casting operations required a different field for each type for proper sign convention which ran straight into the big bag of if statements in PyJPValue_create.   Those if statements were replaced with logic in the convertToPythonObject methods which were already part of the system.   A secondary bug with overflow from unsigned long was also deal with, but there may be an additional one with array assignments.

Fixes #602

### DIFF
--- a/native/common/include/jp_primitive_accessor.h
+++ b/native/common/include/jp_primitive_accessor.h
@@ -182,7 +182,7 @@ public:
 } ;
 
 template <typename base_t>
-class JPConversionLongNumber : public JPConversion
+class JPConversionLongNumber : public JPConversionLong<base_t>
 {
 public:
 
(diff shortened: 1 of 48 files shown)


native/common/include/jp_primitive_accessor.h
```diff
@@ -164,16 +164,25 @@ class JPConversionLong : public JPConversion
...
+			if (val == -1)
+				JP_PY_CHECK();
+			base_t::field(res) = (typename base_t::type_t) val;
+		} else
+		{
+			jlong val = (jlong) PyLong_AsLongLong(match.object);
+			if (val == -1)
+				JP_PY_CHECK();
+			base_t::field(res) = (typename base_t::type_t) base_t::assertRange(val);
+		}
```
Thrameos: We select the type of conversion here.  We need different functions based on the whether it is an implicit or a user requested cast.

native/common/jp_primitivetype.cpp
```diff
@@ -29,3 +29,27 @@ bool JPPrimitiveType::isPrimitive() const
...
+	if (newobj == NULL)
+		return NULL;
+
+	((PyVarObject*) newobj)->ob_size = Py_SIZE(tmp);
+	for (Py_ssize_t i = 0; i < n; i++)
+	{
+		newobj->ob_digit[i] = tmp->ob_digit[i];
+	}
+	return (PyObject*) newobj;
+}
```
Thrameos: Unfortunately the Python API does not have an easy way to call long_subtype_new from outside of Python so we have to replicate the functionality here.

PR #637, opened 2020-03-21
Link Java and Python GC

Thrameos:
This causes Java and Python gc to trigger together (when appropriate).  Java GC will trigger Python when it runs. Python is more aggressive and less costly, so we only trigger when certain conditions are met.  This has been tested on Window and Linux.  But I do not have access to OSX.


native/common/jp_gc.cpp
```diff
@@ -0,0 +1,182 @@
...
+		if (current > limit)
+		{
+			limit = high_water + DELTA_LIMIT;
+			run_gc = 1;
+		}
+
+		// Predict if we will cross the limit soon.
+		ssize_t pred = current + 2 * (current - last);
+		if (pred > limit)
+			run_gc = 2;
```
Thrameos: This section is the nasty part.  We have to make rules that predict when we are about to head into a bad territory and then call the GC.  But the accounting for the various operating systems are poor at best.  Why it is so hard to just get a tally of how much memory has been allocated and not freed is a mystery to me.

native/java/org/jpype/ref/JPypeReferenceSet.java
```diff
@@ -0,0 +1,141 @@
...
+    this.items--;
+    ref.cleanup = 0;
+    ref.pool = -1;
+  }
+
+  /**
+   * Release all resources.
+   *
+   * This is triggered by shutdown to release an current Python references that
+   * a being held.
```
marscher: are being held

native/java/org/jpype/ref/JPypeReferenceSet.java
```diff
@@ -0,0 +1,141 @@
...
+    for (Pool pool : pools)
+    {
+      for (int i = 0; i < pool.tail; ++i)
+      {
+        JPypeReference ref = pool.entries[i];
+        long hostRef = ref.hostReference;
+        long cleanup = ref.cleanup;
+        if (cleanup == 0)
+        {
+          System.out.println("Fault 2");
```
marscher: if you must print, I would prefer System.err

native/java/org/jpype/ref/JPypeReferenceSet.java
```diff
@@ -0,0 +1,141 @@
...
+
+import java.util.ArrayList;
+import static org.jpype.ref.JPypeReferenceQueue.removeHostReference;
+
+/**
+ *
+ * @author nelson85
+ */
+public class JPypeReferenceSet
+{
```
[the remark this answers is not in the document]
marscher: I do not believe we can perform add/remove method without synchronization. So this is perfectly fine.

native/common/jp_gc.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include <mach/mach.h>
+#else
+// Linux doesn't have an available rss tally so use mallinfo
+#define USE_MALLINFO
+#include <malloc.h>
+#endif
+#define DELTA_LIMIT 10*1024*1024l
+#define HARD_LIMIT 200*1024*1024l
+
+namespace
```
marscher: whats the matter with this empty namespace?

native/common/jp_gc.cpp
```diff
@@ -0,0 +1,182 @@
...
+	PROCESS_MEMORY_COUNTERS pmc;
+	GetProcessMemoryInfo(GetCurrentProcess(), &pmc, sizeof (pmc));
+	current = (size_t) pmc.WorkingSetSize;
+
+#elif defined(USE_TASK_INFO)
+	struct mach_task_basic_info info;
+	mach_msg_type_number_t count = MACH_TASK_BASIC_INFO_COUNT;
+	if (task_info(mach_task_self(), MACH_TASK_BASIC_INFO, (task_info_t) & info, &count) == KERN_SUCCESS)
+		current = (size_t) info.resident_size;
+	return (size_t) 0; /* query failed */
```
marscher: you assign current, but then return 0 unconditionally?!

native/common/jp_gc.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+#elif defined(USE_MALLINFO)
+	struct mallinfo mi;
+	mi = mallinfo();
+	current = (size_t) mi.uordblks;
```
marscher: uordblks contains the number of free memory chunks, I thought you want to obtain use memory blocks instead?
I think this sort of memory information platform thingy has already been solved elsewhere (header only at best).

native/common/jp_gc.cpp
```diff
@@ -0,0 +1,182 @@
...
+	JPPyObject collect(JPPyRef::_call, PyObject_GetAttrString(PyJPModule, "_collect"));
+	PyList_Append(callbacks.get(), collect.get());
+	JP_PY_CHECK();
+
+	// Get the Java System gc so we can trigger
+	_SystemClass = (jclass) frame.NewGlobalRef(frame.FindClass("java/lang/System"));
+	_gcMethodID = frame.GetStaticMethodID(_SystemClass, "gc", "()V");
+
+	running = true;
+	getWorkingSize();
```
marscher: why do you need to call it here, if you don't care about the return value?

native/java/org/jpype/ref/JPypeReferenceSet.java
Thrameos: Naw that was just a missed debugging line on an error on a sanity check for an error during development.  Removing and commenting appropriately.

native/common/jp_gc.cpp
Thrameos: Variables were there and it appears that I forgot to come back and remove it.

native/common/jp_gc.cpp
Thrameos: Gack!

native/common/jp_gc.cpp
Thrameos: uordblks  - The total number of bytes used by in-use allocations.

http://man7.org/linux/man-pages/man3/mallinfo.3.html

I looked and couldn't find much on this topic that was useful.  Lots of stuff pointing to /proc which required a ton of parsing but very little not much in the way of actual guidance.  If you have a header that addresses it then I can include it.

native/common/jp_gc.cpp
Thrameos: Darn that was supposed to set the initial highwater mark... that code got moved.  Need to fix.

native/common/include/jpype.h
```diff
@@ -178,8 +178,14 @@ class JPResource
...
 #define JP_RAISE(type, msg)                 { throw JPypeException(JPError::_python_exc, type, msg, JP_STACKINFO()); }
 
+#ifndef PyObject_HEAD
+struct _object;
+typedef _object PyObject;
+#endif
+
 // Base utility headers
 #include "jp_javaframe.h"
+#include "jp_gc.h"
```
marscher: why do all other classes need to know about GC? Isn't it an implementation detail or do other classes call GC directly?

native/common/jp_gc.cpp
marscher: As you said it seems like a hard task to do it in a portable fashion. I do not have a library at hand. At least we should ensure that the info is approximately the right thing and I'm not yet sure about it.

marscher: You're right about uordblks

native/common/jp_gc.cpp
```diff
@@ -0,0 +1,177 @@
...
+	if (task_info(mach_task_self(), MACH_TASK_BASIC_INFO, (task_info_t) & info, &count) == KERN_SUCCESS)
+		current = (size_t) info.resident_size;
+
+#elif defined(USE_MALLINFO)
+	struct mallinfo mi;
+	mi = mallinfo();
+	current = (size_t) mi.uordblks;
+#endif
+
+	return current;
```
marscher: When there is nothing available to obtain this number, we should raise, shouldn't we?
I think we will get bogus behavior later, if we rely on the wrong value.

native/java/org/jpype/ref/JPypeReferenceSet.java
```diff
@@ -0,0 +1,141 @@
+/*
```
marscher: proper header pls

native/common/jp_gc.cpp
Thrameos: Okay this one does demand a check.  The intended behavior is that if stuff fails badly we just go back to the old behavior of no linked gc.   Unfortunately the test script for this is way to complex for a test case currently.   I will dump it in project for now so it doesn't get deleted.

native/common/include/jpype.h
Thrameos: I remember there was some weird order dependency problem that I was having earlier that required it to be at that spot.  Lets see if it is resolved.... nope still there.  The kitchen sink (JPContext) demands GC.   Lets just forward declare it....  okay so module and reference queue are the users.... resolved.

native/common/jp_gc.cpp
Thrameos: I used a few random sources since I could not find anything definitive.  Do you happen to know if mallinfo functions on osx  as it would be best it we had a tested pattern.

chip.org/wiki/resident_set_size
https://stackoverflow.com/questions/38490320/how-to-query-amount-of-allocated-memory-on-linux-and-osx/43126663

My conclusion is that linux only gives the maximum rss in getrusage() which is okay if I don't mind it overflowing and then having to change my guess rate.  But that would be horrible for a program that has random behavior and so not really acceptable.  Once the high water hit the hard limit we have to call every time.  Thus a switch to mallinfo.  mallinfo apparently doesn't capture everything but it does get the majority of numpy and python.  

Windows page unfortunately gets both python and Java so the behavior is really odd.  Both languages are GC and they like holding on to pages so we get gross block changes rather than anything fine grain.   So the trigger is haphazard because we only see page level changes.  It keeps things under the we are leaking like crazy line, but a better predictor would help.  I am worried in real code that stuff staying in blocks will cause overall growth.  The alternative is to use HeapWalk, but that looked miserable and it doesn't give me much info.

OSX.... er don't have access so no clue what I am doing on that one.

I tried neutral methods at first by instrumenting the malloc/free call using Python and the hidden malloc size, but their auditing tools are not called until after the memory is freed so I can't get the free size and it is pretty dangerous for portability.  But it would have been the most direct as it would be locked to space managed by Python (and numpy).

Thrameos:
@marscher is this one ready to merge?

marscher:
If you add some note for the changelog that be great. I think we need to ask some people to test this on different OS, for which we do not have access to.

Thrameos:
Okay I will add a note and see if I can get one of my office mates to try test it.  It will be a bit difficult given we are all locked in our homes over here, but I should find one willing to do a phone debugging session.  

This one will thus be placed on hold for merging until testing is complete.

Thrameos:
OSX testing.   Currently showing minor growth.   Needs some retuning.
https://travis-ci.org/github/jpype-project/jpype/jobs/670971357

Thrameos:
It appear that the gc link is working for OSX.  It is a bit less aggressive than I would like as just like windows it is counting both Java and Python space so it tends to see only the large block releases rather than the finer scale changes.  This means that it will tend to grow by hitting the limit for a while before it starts cycling at a fairly stable rate.  This is still a big improvement over memory exhaustion so it is worth including even it the policy is a bit suboptimal.  The linux version is much more aggressive and keeps the size down, but this of course means more time spent running the GC.  We can refine the tuning more if it appears to be an issue.  I believe that we can apply once it clears the CI.

https://travis-ci.org/github/jpype-project/jpype/jobs/671072476

PR #639, opened 2020-03-22
Support for direct buffers

Thrameos:
Tonight's PR adds support for memory views on direct buffer types.  Direct buffers can be converted to numpy arrays and written to directly.  This is a basic functionality pull request and has not be run on production code.  Memory buffers held across shutdown will result in bad behavior as Python lacks an API to disable active buffers.  

This fixes an error discovered in JPArray.  When I duplicated the JPArray to JPBuffer I also managed to duplicate a Global reference leak bug.  A sanity check was added to prevent a similar leak from happening again.

This pull request builds on #636 and will be rebased after it is merged for easier review.


native/common/jp_arrayclass.cpp
```diff
@@ -60,7 +60,6 @@ JPPyObject JPArrayClass::convertToPythonObject(JPJavaFrame& frame, jvalue value,
 	}
 	JPPyObject wrapper = PyJPClass_create(frame, this);
 	JPPyObject obj = PyJPArray_create(frame, (PyTypeObject*) wrapper.get(), JPValue(this, value));
-	PyJPValue_assignJavaSlot(frame, obj.get(), JPValue(this, value));
```
Thrameos: This silly bug was hard to find.   It wasn't until I duped the code for buffer did it appear

native/common/jp_buffertype.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+JPPyObject JPBufferType::convertToPythonObject(JPJavaFrame& frame, jvalue value, bool cast)
+{
+	JP_TRACE_IN("JPBufferClass::convertToPythonObject");
+	if (!cast)
+	{
+		if (value.l == NULL)
+			return JPPyObject::getNone();
+	}
```
Thrameos: Compact me.

native/common/jp_context.cpp
```diff
@@ -246,6 +229,31 @@ void JPContext::startJVM(const string& vmPath, const StringVector& args,
 		m_ReferenceQueue->m_ReferenceQueue = JPObjectRef(frame,
 				frame.CallObjectMethodA(m_JavaContext.get(), getReferenceQueue, 0));
 
+		// Set up methods after everthing is start so we get better error
+		// messages
+		cls = m_ClassLoader->findClass(frame, "org.jpype.JPypeContext");
```
Thrameos: There is an ever present miserable bug that that when something goes wrong in the bootstrapping process the exception handler will segfault.  Hopefully it is finally resolved, but I rearranged the order to make it harder to reoccur.

native/java/org/jpype/JPypeContext.java
```diff
@@ -305,4 +307,23 @@ public Exception createException(long l0, long l1)
...
+    if (b instanceof java.nio.CharBuffer)
+      return ((java.nio.CharBuffer) b).order() == ByteOrder.LITTLE_ENDIAN;
+    if (b instanceof java.nio.IntBuffer)
+      return ((java.nio.IntBuffer) b).order() == ByteOrder.LITTLE_ENDIAN;
+    if (b instanceof java.nio.LongBuffer)
+      return ((java.nio.LongBuffer) b).order() == ByteOrder.LITTLE_ENDIAN;
+    if (b instanceof java.nio.FloatBuffer)
+      return ((java.nio.FloatBuffer) b).order() == ByteOrder.LITTLE_ENDIAN;
+    if (b instanceof java.nio.DoubleBuffer)
+      return ((java.nio.DoubleBuffer) b).order() == ByteOrder.LITTLE_ENDIAN;
```
Thrameos: This code was just miserable.   For reasons I find unfathomable there is no order on the base class so every single class needs its own lookup.  Thus better to do this in Java because otherwise we have to lookup the methodID over and over.

native/java/org/jpype/manager/TypeManager.java
```diff
@@ -368,6 +369,10 @@ private ClassDescriptor createClass(Class<?> cls, boolean special)
       modifiers |= ModifierCode.THROWABLE.value;
     if (Serializable.class.isAssignableFrom(cls))
       modifiers |= ModifierCode.SERIALIZABLE.value;
+    if (Buffer.class.isAssignableFrom(cls))
+    {
+      modifiers |= ModifierCode.BUFFER.value | ModifierCode.SPECIAL.value;
+    }
```
Thrameos: This is an edge case.  Most of the special wrappers get made at boot time, but this one did not have an order problem so we will use a lazy creation pattern.

native/python/pyjp_buffer.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+	sout << "<java beffer " << self->m_Buffer->getClass()->toString() << ">";
+	return JPPyString::fromStringUTF8(sout.str()).keep();
+	JP_PY_CATCH(0);
+}
+
+static void PyJPBuffer_releaseBuffer(PyJPBuffer *self, Py_buffer *view)
+{
+	JP_PY_TRY("PyJPBufferPrimitive_releaseBuffer");
+	JP_PY_CATCH();
```
Thrameos: I honestly have no clue what I am supposed to put here.  The object dereference is done automatically and we don't have to resource track.  Thus I left a stub for now.

native/common/include/jp_buffertype.h
```diff
@@ -0,0 +1,35 @@
...
+#define _JPBUFFERTYPE_H_
+
+/**
+ * Class to wrap Java Class and provide low-level behavior
+ */
+class JPBufferType : public JPClass
+{
+public:
+	JPBufferType(JPJavaFrame& frame,
+			jclass cls,
```
marscher: weird indention

native/common/jp_buffertype.cpp
marscher: yes please :D

native/java/org/jpype/JPypeContext.java
marscher: That's indeed tedious... Agreed to have it on the java side

native/python/pyjp_buffer.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+static void PyJPBuffer_dealloc(PyJPBuffer *self)
+{
+	JP_PY_TRY("PyJPBuffer_dealloc");
+	delete self->m_Buffer;
```
marscher: This assumes that JPBuffers are always direct. Sorry I didn't spot this check earlier.

native/python/pyjp_buffer.cpp
marscher: The buffer capi documentation says that the getBuffer and release functions are a pair like free/malloc. So if the view has been allocated via getBuffer it has to be cleared by calling release. Why is it de-referenced automatically in every case?

native/python/pyjp_buffer.cpp
```diff
@@ -0,0 +1,139 @@
...
+		if ((flags & PyBUF_FORMAT) != PyBUF_FORMAT)
+			view->format = NULL;
+
+		// Okay all successful so reference the parent object
+		view->obj = (PyObject*) self;
+		Py_INCREF(view->obj);
+		return 0;
+	} catch (JPypeException &ex)
+	{
+		PyJPBuffer_releaseBuffer(self, view);
```
marscher: why call the stub here, if it does nothing?

native/python/pyjp_buffer.cpp
```diff
@@ -0,0 +1,139 @@
...
+	JP_PY_CHECK_INIT();
+}
+
+JPPyObject PyJPBuffer_create(JPJavaFrame &frame, PyTypeObject *type, const JPValue& value)
+{
+	PyObject *obj = type->tp_alloc(type, 0);
+	JP_PY_CHECK();
+	((PyJPBuffer*) obj)->m_Buffer = new JPBuffer(value);
+	PyJPValue_assignJavaSlot(frame, obj, value);
+	return JPPyObject(JPPyRef::_claim, obj);
```
marscher: this claims the reference to the buffer? So when it goes out of scope it will be garbage collected??!

native/python/pyjp_value.cpp
```diff
@@ -287,6 +286,13 @@ void PyJPValue_assignJavaSlot(JPJavaFrame &frame, PyObject* self, const JPValue&
 	// GCOVR_EXCL_STOP
 
 	JPValue* slot = (JPValue*) (((char*) self) + offset);
+	// GCOVR_EXCL_START
+	// This is an sanity check that should never trigger in normal operations.make
```
marscher: a sanity check (sorry for being anal)

native/python/pyjp_buffer.cpp
Thrameos: This type is not references as unlike all of the other buffers we use this is direct buffers only so the memory can't move and never needs to be pinned.  The object counter is the PyJPBuffer and as far as I can tell this is automatically done.  So not much appears need to be here.

native/python/pyjp_buffer.cpp
Thrameos: This was left from the array pattern where it is required.  I kept the pattern in case there was actually some need to call the stub.  I can remove it if you think that the stub will never be required.

test/jpypetest/test_leak2.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def testArrayCall(self):
+        JA = JArray(JInt)
+        def call():
+            inst = JA(100)
+            Tracer.attach(inst)
+            self.fixture.callObject(inst)
+        for i in range(100):
+            call()
+        self.assertEqual(Tracer.leaked(), 0)
```
marscher: You have ensured that no intermediate layer has modified the refcount of the JPArray instance.

native/python/pyjp_buffer.cpp
Thrameos: Okay this line likely need to be explained....

Normally we are using the ``_call`` or ``_use`` pattern  not the ``_claim``.   In a ``_use`` we are given an existing reference that we need to ensure we don't lose the reference to (like stupid things like manipulating a map and losing the reference to ourselves.   ``_call`` is also obvious because we got a reference an immediately put into a container.  

``_claim`` is an edge case.  We constructed a new reference, but for whatever reason we didn't put it in a reference counter (special error handling etc), but we want it to enter the reference counting system at some later point.

```
   JPPyObject doSomething()
   {

     // We created a reference but we can't throw because it requires a special handling...
    PyObject* obj = PyDo_Something(...);
     //  Next we do whatever checks we need and the handling routines...
    if (obj==NULL)
    {  // handle error
    }
     ...
    // Okay we know that it isn't 0 and but it still isn't in a reference counting frame so it
    // still has a reference count of 1.
   return JPPyObject(JPPyRef::_claim, obj);
   // So something weird happens here.  We claim the object so now it has a reference count 
   // of one and will die at the end of scope, but before that happens the return value will 
  // be constructed which will copy the reference and increment it to 2.  then decrement it 
  // back to 1.
}
```

So the intent of the line is to hold grab the reference and then transfer control to the return.   That said I do not see any reason for using the claim pattern in this case.  There is no special handling of types nor an exceptional return path.  And in some very rare case assignJavaSlot may fail resulting in a leaked object, though given the type system I am nor sure it can be triggered here.  I will convert it to a ``_call`` pattern to be safe.

test/jpypetest/test_leak2.py
```diff
@@ -0,0 +1,80 @@
...
+            Tracer.attach(inst)
+            memoryview(inst)
+        for i in range(100):
+            call()
+        self.assertEqual(Tracer.leaked(), 0)
+
+    def testBufferCall(self):
+        byte_buffer = jpype.JClass('java.nio.ByteBuffer')
+        def call():
+            inst = byte_buffer.allocateDirect(10)
```
marscher: now the memory of the buffer is outside of JVMs heap. The tracer ensures that the inst object is cleared (which then calls dealloc and frees the buffer).

native/python/pyjp_value.cpp
Thrameos: It originally said "an idiot check" check, but my save censor converted it to "an sanity check".  Guess I need to add "an idiot" to the filter first.

I am unfortunately beyond dyslexic, as you may have noticed, resulting in anything that isn't being compiled (and thus checked by a computer) to often result in odd mispellings, weird word choice, and my favorite hilarious word substitutions.  I try to catch them by rereading what I write a day later as usually some distance allows me to interpret the words more clearly, but that still means I miss far more than one would expect from a native speaker.

test/jpypetest/test_leak2.py
Thrameos: Yeah.  This pattern was submitted by a user on the bug report for memory exhaustion.  By attaching a Python object to a wrapper it verifies that we didn't accidentally leak the Python portion of the wrapper.   It does check for one type of leak.  Leaks on the Java side remain difficult.  Although we can ask Java to run a garbage collection is it just a suggestion.  Thus writing tests to see if we leaked a Java object are more challenging as we can't guarantee that  the object is collected.

I have debated adding a atomic integer counter to JPContext (when compiled with instrumentation) so that we can actually get the current number of global references that we are holding.  We would then get the count before and after the operation.  But I haven't gotten around to it and like everything having to do with threads I worry about the portability of such a trick for code in general.   I got bitten once already by the JPTracer where  the function back referencing was causing a seg fault.  Obviously a counter for auditing is much less risk.

native/python/pyjp_buffer.cpp
Thrameos: You can have a buffer which is not direct in this code, but it won't have a memoryview currently.  I left the stubs in case we wanted to implement indirect buffers as well.   

But I am very reluctant to go down that path.   If we implement both direct and java array backed buffers on this interface, then we will suffer the problem that java array buffers can't be made into read/write buffers because the JNI pinning restrictions.   As it stands you can make a memory buffer for a direct buffer which will fail for an indirect buffer, but if it is indirect you can call the array method which will give you the memory buffer with array rules.

I believe this is the "best" compromise in terms of design.   Arrays have one set of rules and (direct) Buffers have a different set.    Obviously some docs would help.   The rules are pretty obvious at the implementation level given the JNI calls, but may be surprising for those not familiar with JNI.

native/common/include/jp_buffertype.h
```diff
@@ -0,0 +1,30 @@
+#ifndef _JPBUFFERTYPE_H_
+#define _JPBUFFERTYPE_H_
+
+/**
+ * Class to wrap Java Class and provide low-level behavior
+ */
+class JPBufferType : public JPClass
+{
+public:
+	JPBufferType(JPJavaFrame& frame, jclass cls, const string& name, JPClass* superClass, const JPClassList& interfaces, jint modifiers);
```
marscher: Actually it was fine to have one argument per line, I just thought it was not aligned with the first bracket. Sorry for wasting your time here.

native/python/pyjp_buffer.cpp
marscher: I agree, that we do not want to mix these concepts. Some documentation/examples would be great indeed.

native/python/pyjp_buffer.cpp
marscher: But you need to register the function as part of a mandatory set of the Python buffer protocol? 
A comment about this in the code wouldn't hurt anybody.
Also I don't understand the try catch as nothing will be thrown in this block, right?

native/python/pyjp_buffer.cpp
marscher: If it does no harm, lets just leave it. Eventually it will be needed for whatever reason.

test/jpypetest/test_leak2.py
marscher: keep it simple stupid :)

native/python/pyjp_buffer.cpp
Thrameos: Yeah as far as I can tell from looking at the C code it was a requirement.  There were places where it appeared not to be checking if the release was null and I wasn't going to risk it.   

The worthless try/catch in this block is for the tracer system not anything meaningful to the user.  In the case where I can't catch an error in the debugger and I have to find all the cases each object was born and died I have to use the tracer.  Unfortunately the JP_TRACE_OUT pattern implies that it will rethrow the exception which makes some C++ compilers unhappy because a extern "C" code can't have an exception.  Thus the next safest code to use the tracer is JP_PY_TRY/JP_PY_CATCH which should appear in all extern "C" Python calls.   I know that it ends up with some really odd code because it appears like we are handling exceptions in routines that don't have any C++ code or exceptions.   It also does one other function, which is these blocks can serve as failure points for instrumentation.   Yes, I release we have a lot of stuff hooked up to those macros (tracing, error handling, and instrumentation) but splitting them out just makes the code blotted and less readable.

native/common/include/jp_buffertype.h
Thrameos: Netbeans has some strange behavior when it comes to mixed spaces and tabs resulting in odd results.   I try when a list of arguments gets too long to follow to split them out, but it only lines up properly if the editor is set with the same tab width that I am using and even then sometimes it decides to randomly respace it after typesetting.   I know that netbeans is not a real C++ editor, but when diagnosing code in 3 languages at a time it does serve a purpose.  Like the "}  ;"  I just have to accept the good stuff it is doing like cross referencing between the three open projects and ignore the weird.

native/python/pyjp_buffer.cpp
Thrameos: This code was revised and should be good to go.

Thrameos:
@marscher  I think this one is ready for inclusion.   I am never quite sure if you want me to hit the review button again after I resolve issues or merge once it is a proper state.   

Either way other than the gc linker we should be getting very close a stable point for release.  I hit most of the low hanging fruit and even a few stretch goals in this 30 pulls in 30 nights campaign.   We still have a few critical bugs in 0.7.2 so another release is warranted.   

I have a pretty good list of items that I plan to hit in the next cycle.  But the one with the greatest issues is allowing bare Python objects to be passed to Java as org.jpype.PyObject entities with actual methods that can be called from Java.   The problem with that is there are certain cases I can construct in which the reference loop is unsolvable.  That is if a Python object which points to a Java object (Global reference) is placed an Python object which is then added to a Java container, then there is no way to resolve the reference loop and it will leak.  I could try to make back references from Java to Python to be weak references but that would leave a lot of undesireable behavior.     Support for this is needed if I ever want to be able to support the reverse bridge mode where Java is in charge.  But the inability to make if completely safe for the user remains a concern.

marscher:
We can schedule a release soon as there has been lots of progress so far.
I think the reverse direction is new major feature which we should postpone until we have a stable 1.0. Eventually it is not even possible to do it a fool proof fashion.

Thrameos:
I reviewed the remaining JPype 0.8 items.  2 of them remain distant stretch goals and one is at least partially dependent on those goals.  Thus I am pushing all of them to 0.9.  That means that the one remaining item is documentation.  Yeah I know the thing that every programmer would rather procrastinate on.  Better to cut code and blaze a trail than document what you have done.  

I will devote the week to trying to get that in better order.  At least I left detailed notes the last time I worked on it, so perhaps it won't be as bad as I dread.  At least procrastinating on it for the week will inspire me to put down the road map for 0.9 if nothing else.

PR #640, opened 2020-03-22
Work on coverage system

Thrameos:
This is a test pull to try to get the coverage system working.   The old pattern required the caching to skip rebuilding but the remove ant patch broke that.  This tries to make the pattern work again and add java coverage.


Thrameos:
I am merging this so that I can proceed with coverage work.

PR #643, opened 2020-03-23
Java coverage

### LINKED ISSUE #641: Comparisons to None result in Null Pointer exception
The logic for ``__eq__`` is not very aware of null pointer and thus produced null pointer exception where it should rather than just true or false as appropriate.

Thrameos:
More coverage testing.  This time for the Java layer.

There was a few bug discovered with ``__eq__`` that required addressing regarding comparisons to None. 

In addition a bug with using Proxies to interfaces constructed with custom classloaders was resolved.

Fixes #641


native/python/pyjp_class.cpp
```diff
@@ -545,7 +545,9 @@ static PyObject *PyJPClass_cast(PyJPClass *self, PyObject *other)
 	jobject obj = val->getJavaObject();
 	if (obj == NULL)
 		return PyJPValue_create(frame, JPValue(type, 0)).keep();
+	printf("A\n");
```
marscher: pls remove dbg prints

native/python/pyjp_object.cpp
```diff
@@ -51,6 +51,69 @@ static PyObject *PyJPObject_new(PyTypeObject *type, PyObject *pyargs, PyObject *
...
 
+
+static const char* op_names[] = {
+	"<", "<=", "==", "!=", ">", ">="
+};
+
+static PyObject *PyJPObject_compare(PyObject *self, PyObject *other, int op)
+{
+	JP_PY_TRY("PyJPObject_compare");
+	if (op == Py_NE)
```
KOLANICH: `switch`?

Thrameos: If you mean a switch statement then not sure it would apply given we are only testing NE,EQ, everything else, so I am not sure a switch statement would add any clarity.  NE is implemented by calling the method again and then checking if we got Py_False back. 

If you mean something else please elaborate.   The logic for the comparison is pretty messy because there are a bunch of edge cases (null first, null second, null both, java second, not java second, etc)

KOLANICH: I meant a `switch` statement.

>I am not sure a switch statement would add any clarity.

How about the rest of comparators via `compareTo`?

Thrameos: This section is for objects so we only have == and !=.   You are thinking about comparable and Numbers.   Those comparisons are both located elsewhere.   

Numbers is a delegated comparison with the one exception of null pointers. (again we only have EQ and NE in that limited case)
```
static PyObject *PyJPNumberLong_compare(PyObject *self, PyObject *other, int op)
{
        JP_PY_TRY("PyJPNumberLong_compare");
        JPContext *context = PyJPModule_getContext();
        JPJavaFrame frame(context);
        if (isNull(self))
        {
                if (op == Py_EQ)
                        return PyBool_FromLong(other == Py_None);
                if (op == Py_NE)
                        return PyBool_FromLong(other != Py_None);
                PyErr_Format(PyExc_TypeError, "'%s' not supported with null pointer", op_names[op]);
                JP_RAISE_PYTHON();
        }
        return PyLong_Type.tp_richcompare(self, other, op);
        JP_PY_CATCH(NULL);
}
```

And for Comparable it is currently in Python
```
@_jcustomizer.JImplementationFor('java.lang.Comparable')
class _JComparable(object):

    def __eq__(self, o):
        return self.compareTo(o) == 0

    def __ne__(self, o):
        return self.compareTo(o) != 0

    def __gt__(self, o):
        return self.compareTo(o) > 0

    def __lt__(self, o):
        return self.compareTo(o) < 0

    def __ge__(self, o):
        return self.compareTo(o) >= 0

    def __le__(self, o):
        return self.compareTo(o) <= 0

    __hash__ = _jpype._JObject.__hash__
```

So if this was for comparable then a switch would definitely be warranted as we have a lot of cases to cover.  But java.lang.Object is pretty limited.  In general we are using the Python method resolution, thus we hide methods by implementing them on derived classes.  That does lead to some confusion.

That said is it a very good idea that we need to investigate comparable for the same bug as the object one.  It likely throws NullPointerException.   My question is what should be the correct behavior for LE, LT, GT, and GE when one or both the pointer are null.  I will tag this in the issue tracker and get back to it a bit later.  I am trying to not push all of the JPype code back into the CPython layer if possible, but there are some things like this horrible equals logic that are just too huge and slow if written in Python.

Thrameos: Okay I added code for comparable as well (and yes that one needs to use switch).  It has a lot of edge cases and a really horrible difficulty. When you do a comparison of a Python type we have to figure out what the type to convert to, but because of type erasure that is a total guessing game.  Hopefully it works.

Again thanks for taking the time to review.

PR #644, opened 2020-03-23
Inner class bug

Thrameos:
Addresses the problem that inner classes are created by the outer class.  So if the inner class is requested first it attempts to create itself twice.   This needs some test code to verify that it is being dealt with properly, but for now I just have time a quick patch job.

This pull request deals with ``_hints`` cannot be set bug.

This still needs a test bench before it can be included.

### DIFF
--- a/test/harness/jpype/types/Outer.java
+++ b/test/harness/jpype/types/Outer.java
@@ -1,0 +1,15 @@
+package jpype.types;
+
+public interface Outer
+{
+  Object get();
+
+  public class Inner implements Outer
+  {
+	  public Object get()
+	  {
+		  return null;
+	  }
+  }
+
+}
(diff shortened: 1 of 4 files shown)


test/harness/jpype/types/Outer.java
```diff
@@ -0,0 +1,15 @@
+package jpype.types;
+
+public interface Outer
+{
+  Object get();
+
+  public class Inner implements Outer
```
marscher: programmers who would ever use such a coding style should be tared and feathered :)

Thrameos: It seems to come up pretty often with interfaces where an interface contains a collection of concrete implementations.  Certainly not something that I endorse as I never managed to trip this bug in in the 100 repos that I work on (spent the week converting from bitbucket to gitlab and the system admin through it was some sort of joke when we pushed 100 repos in), but am not sure dropping a coder into boiling tar is a good solution.  I have seen much, much worse styles than this.

Fun fact, did you know that this code will compile in Java...
```
  {
    List List = new ArrayList();
    List.add(1);
  }
```

PR #646, opened 2020-03-24
Weak proxy

### LINKED ISSUE #585: Support for weak references on Proxies
Proxies currently cause Python objects to assume the lifespan of the Java object used to reference it.  Unfortunately there are cases in which this leads to memory exhaustion as the Java garbage collection does not trigger when Python consumes too much memory due to stale objects.  

We need a way to discover if Python memory is growing excessively and deal with it.

See test pull request #581

Thrameos:
This PR makes it so that Proxies will reuse there instance handlers rather than creating new ones over and over.  We test this by looking at the system ID which changes every time the InvocationHandler is recreated.

Fixes #585

### DIFF
--- a/test/jpypetest/test_proxy.py
+++ b/test/jpypetest/test_proxy.py
@@ -457,11 +457,6 @@ class ProxyTestCase(common.JPypeTestCase):
             return i, obj
         i0, obj = f()
         i1 = hc(obj)
-        java.lang.System.gc()
-        time.sleep(0.5)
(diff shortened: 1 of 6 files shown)


test/jpypetest/test_proxy.py
```diff
@@ -17,6 +17,7 @@
 from jpype import *
 import common
 import sys
+import time
```
marscher: unused import?

PR #648, opened 2020-03-26
Add shutdown at end to make sure that we dont crash on shutdown.

Thrameos:
This ends our test suite so that we properly finish our coverage tests and ensure that shutdown does not result in a crash.  Using zzz should ensure that this is always the last test run.

### DIFF
--- a/test/jpypetest/test_zzz.py
+++ b/test/jpypetest/test_zzz.py
@@ -1,0 +1,21 @@
+import sys
+import _jpype
+import jpype
+from jpype.types import *
+from jpype import JPackage, java
+import common
+import pytest
+try:
+    import numpy as np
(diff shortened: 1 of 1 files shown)


Thrameos:
The weird loss of coverage has to do with the adding of the Java test bench.  So how does adding more test decrease coverage?  Well it has to do with what is the references base branch for the comparison.  If we push a bunch of tests into the system and then a pull references those  added tests,  (but it doesn't have them because it came from an earlier point in time)  then it show a big reduction.   In this case all the java files not show zero coverage.

We can fix this by rebasing patches after each merge but there are downsides to that (especially for big pulls).   For now I don't think there is anything to worry about.

PR #655, opened 2020-03-28
Memory leak fix on JPClassHints

### LINKED ISSUE #654: jp_classhints memleaks JPAttributeConversion
Just found it during valgrinding the direct buffer branch. In JPClassHints::addAttributeConversions a JPAttributeConversion pointer is created, which is never deleted (jp_classhints.cpp:188)

Thrameos:
Adds a clean up routine to JPClassHints.

Fixes #654

### DIFF
--- a/native/common/jp_classhints.cpp
+++ b/native/common/jp_classhints.cpp
@@ -80,6 +80,12 @@ JPClassHints::JPClassHints()
 
 JPClassHints::~JPClassHints()
 {
+	for (std::list<JPConversion*>::iterator iter = conversions.begin();
+			iter != conversions.end(); ++iter)
+	{
+		delete *iter;
+	}
+	conversions.clear();
 }
 
 JPMatch::Type JPClassHints::getConversion(JPMatch& match, JPClass *cls)


native/common/jp_classhints.cpp
```diff
@@ -80,6 +80,12 @@ JPClassHints::JPClassHints()
 
 JPClassHints::~JPClassHints()
 {
+	for (std::list<JPConversion*>::iterator iter = conversions.begin();
+			iter != conversions.end(); ++iter)
+	{
+		delete *iter;
+	}
+	conversions.clear();
```
marscher: why would you clear the vector of this dying object?

Thrameos: It is a habit of mine to never leave memory allocated or a list with contents.   Just like Py_CLEAR, if you leave something with a valid point in the destructor and it somehow ends up back in the dtor again (race condition, stack ordering, etc) then the object will delete the memory again.    Now this policy does hide those really evil errors when that condition does come up, but as those are usually transient edge cases and the errors that they cause are almost always going to be of the "heap corruption" type (which burn a week or more to solve), clearing a list on a dying object is usually the safest option.

marscher: I was wondering if any c++ standard would allow calling dtors twice

Thrameos: Back in 1997 when I was working on libsigc++ I hit lots of edge cases (I think at one point I had 6 active bug reports in the gcc ticket system).  In part that was because it was signals and slots and you would get users that would construct really nasty edge cases where the slot object was on the call stack while being deleted.  Made for some absolute head scratching errors.  At some point I just came to accept that calling the dtor only once was more of a suggestion than a rule and decided that defensive programming was a better policy than efficiency.

PR #658, opened 2020-03-29
Dealing with relative paths

### LINKED ISSUE #652: importing the .py use Jpype will be reported as error
my question is I have competed a example.py with Jpype and it works well。
I want to import the  example.py  to test.py, use the function in example.py,but the pycharm report that '' jpype._jclass.NoClassDefFoundError: java.lang.NoClassDefFoundError: org/openscience/cdk/smiles/SmilesParser ''
i want to know how to fix it 
thank you

### LINKED ISSUE #657: Relative paths for add path
Java and Python have very different notions of paths.  This can lead to strange issues in two cases:

1) Relative paths that don't match the current notion of Python.  For example a jar file in a Python module where the module is referencing relative to the Python directory and Java is referencing the starting directory.
2) Operating systems for which the JVM path system does not match the python one (Cygwin).

The safest course is to use Python path lib to convert relative paths into absolute paths.  We can do this for many systems automatically.  But this has difficulties for Cygwin where its path structure differs from Java and to a lesser extent WSL. 

The user can work around the issue by manually converting the path to an absolution one but we may be able to improve this.

Thrameos:
This deals with an issue in #652.  The user was attempting a path which was relative but the way that the addClassPath call was structured it was performing the lookup relative to the wrong location.   This pull request corrects that flaw so that paths are always relative to the call location rather than some arbitrary point like the starting directory or the installation directory.   We use inspect to get the callers file location.

This needs additional testing before it can be included.

Fixes #652
Fixes #657

### DIFF
--- a/jpype/_classpath.py
+++ b/jpype/_classpath.py
@@ -1,6 +1,5 @@
 import os as _os
 import sys as _sys
-import glob as _glob
 
 __all__ = ['addClassPath', 'getClassPath']
 
@@ -34,8 +33,9 @@ if _sys.platform == 'cygwin':
         return parts
 
     def _posix2win(directory):
+        directory = str(directory)
         if len(directory) > 3 and directory[1:3] == ":\\":
-            return directory
+            return Path(directory)
         root = _get_root()
         directory = _os.path.abspath(directory)
         paths = _splitpath(directory)
@@ -43,9 +43,9 @@ if _sys.platform == 'cygwin':
             paths.pop(0)
             drive = paths.pop(0)
             paths.insert(0, "%s:" % drive)
-            return '\\'.join(paths)
+            return Path('\\'.join(paths))
         paths.insert(0, root)
-        return '\\'.join(paths)
+        return Path('\\'.join(paths))
     # needed for testing
     __all__.append("_posix2win")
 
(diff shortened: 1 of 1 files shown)


marscher:
```
    def _posix2win(directory):

        directory = str(directory)

        if len(directory) > 3 and directory[1:3] == ":\\":

            return Path(directory)

        root = _get_root()

        directory = _os.path.abspath(directory)

        paths = _splitpath(directory)

        if paths[0] == "cygdrive":

            paths.pop(0)

            drive = paths.pop(0)

            paths.insert(0, "%s:" % drive)

>           return Path('\\'.join(paths))

E           NameError: name 'Path' is not defined
```

Thrameos:
@marscher does this one require additional work.   The cygwin test bench is still down so it is hard to say if all the issues have been dealt with.   I am not sure if azure pipelines has cygwin/mingw.   Both platforms are pretty broken due to problems with the path separator and issues with multithreading in the Python implementation.   I am not sure there is any compelling reason to continue to support those broken platforms.

marscher:
I agree we should not support cygwin. There are better options nowadays.

Thrameos:
So should I put in a pull request to remove cygwin dependent logic?

marscher:
On 4/7/20 11:20 PM, Karl Nelson wrote:
> So should I put in a pull request to remove cygwin dependent logic?
> 

Please go ahead and make it so.

Thrameos:
Yes, Captain.

See #674

PR #662, opened 2020-04-01
Revert travis to earlier point

Thrameos:
test to see if can unstick travis using earlier version


Thrameos:
@marscher hmm, no luck getting travis to report its status back to github.   I tried known good configurations but it is still stuck.  You had mentioned a log someplace, but I wasn't able to located it.   I don't want to start messing with the action settings as I don't know what I am doing.

PR #663, opened 2020-04-01
Proxy lambda

Thrameos:
On the list of items for JPype 0.8 that I discovered was interactions between Java classes that are annotated with ``@FunctionalInterface`` and Python callables.   These methods are intended for Java lambdas to implement functionality and we should do the same for  Python.  Therefore, we have added an implicit type conversion that automatically figures out what the SAM (single abstract method) is in a Java interface and creates a proxy for it on the fly.  

This allows
```
    def Func(x):
       return 2*x 
    ja = JObject(Func, "java.util.function.DoubleUnaryOperator")
    print(ja.applyAsDouble(2.0))

    ja = JObject(lambda x: 2*x, "java.util.function.DoubleUnaryOperator")
    print(ja.applyAsDouble(2.0))
```

We have not added anything to make FunctionalInterfaces appear as Callable for the purpose of Python, but it can be done in another PR.


Thrameos:
This one is a complete mystery to me.  I have tried 10 different machines/versions of Python and nothing thus far replicates the errors shown in the CI.  And without a way to replicate it I can't correct the issue.  So I am placing a hold on this until after the next release.

Thrameos:
@marscher This one is ready for review.

Thrameos:
This one is exercising a very weird edge case that I need to solve.   Best guess is that something can hop over proxy barrier and thus be called while things are being shut down.   The stack trace indicates that something was calling a method while the Destroyer was operating which is a sure way to crash.  

I believe the issue is daemon threads.   Java kills everything but daemons before proceeding to destroy the C++ resources.  Unfortunately, proxies can be called from daemons.   Thus, it can start a race.  If the proxy is called before shutdown flag completes it can sneak in between the flag check and be in a live state as resources in TypeManager are being killed off.

Thrameos:
@marscher okay back to ready for review status on this one.   The shutdown upgrades fixed the issue that was happening in the test bench. 

I have one additional upgrade that I would like to do with lambdas.   That is if something is a SAM it should appear on Python side as a callable.   There is still a bit more wiring to make that happen as although Java currently knows the name of the SAM it is not communicating that to the Python object wrapper so we can't call the method.   If it just a matter of passing that info down the chain.

But that is a PR for another day.

test/jpypetest/common.py
```diff
@@ -103,6 +104,7 @@ def setUp(self):
 
             classpath_arg %= jpype.getClassPath()
             args.append(classpath_arg)
+            _jpype.enableStacktraces(True)
```
marscher, line 107: Is "enableStacktraces" a hidden feature, we should advertise in the documentation? I guess it becomes handy during all kind of debug scenarios.

Thrameos: Yes.  It is an internal feature documented in the section "Getting additional diagnostics" along with a handful of developer features that I use for debugging.  

https://jpype.readthedocs.io/en/latest/userguide.html#c-exceptions-in-jpype

The only section in that area I am aware is currently missing are tricks to use jlink and debugging threading issues.

marscher: That addition sounds interesting. The guide is already super useful! Thank you

Thrameos: I am trying set a new record for largest documentation to user base ratio.  Helps that denominator is small. :wink:

PR #668, opened 2020-04-05
Updates on deprecation

Thrameos:
A few undocumented features need to be dealt with.   In order to support placing Python objects into the Java object system, we need to be able to apply JObject with a specified type for the cast unless we want it to go implicitly.   Given the implicit conversions would introduce a lot of errors, explicit conversions seems like a better idea.  Unfortunately, the existing JObject has casting types that map to existing classes as an undocumented legacy feature.   This doesn't provide any meaningful utility as casting to specific Java types (JObject, JInt, JString, etc) already covers this usage.  Thus we are going to deprecate it so we can replace this with useful functionality.  For the most part the replacements will still function almost identically as we will be overloading the casting system so that we can convert these Python wrapped items through, but deprecating this misfeature now would be safest option.

In addition there are a few deprecated features that are still floating around that should be dealt with 
- JException creating a Class.  This was tagged as causing conflicts with potential support of multiple JVMs.
- jpype.reflect was left from 0.6 series before ``class_`` was supported
- convertStrings defaulting to true.

It is not clear yet if 0.8.0 will be the next release or the release following.  I added a notice of features that are bound of removal to the changelog.

### DIFF
--- a/jpype/_jobject.py
+++ b/jpype/_jobject.py
@@ -104,6 +104,10 @@ def _JObjectFactory(v=None, tp=None):
     elif isinstance(tp, str):
         tp = _jpype.JClass(tp)
     if tp in _jpype._object_classes:
+        if not isinstance(tp, _jpype.JClass):
+            import warnings
+            warnings.warn("Using JObject with a Python type is deprecated.",
+                              category=DeprecationWarning, stacklevel=3)
         tp = _jpype._object_classes[tp]
 
     # Given a Java class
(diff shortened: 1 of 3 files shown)


jpype/_jobject.py
```diff
@@ -104,6 +104,10 @@ def _JObjectFactory(v=None, tp=None):
     elif isinstance(tp, str):
         tp = _jpype.JClass(tp)
     if tp in _jpype._object_classes:
+        if not isinstance(tp, _jpype.JClass):
+            import warnings
+            warnings.warn("Using JObject with a Python type is deprecated.",
+                              category=DeprecationWarning, stacklevel=3)
```
marscher, line 110: :+1: for the correct stacklevel! Usually people just ignore this setting

PR #670, opened 2020-04-06
Upgrades to JPackage and jpype.imports

### LINKED ISSUE #537: Improvements for JPackage and imports
Using the jrt system from Java 9 I was able to get a complete list of the available java/javax packages in the JVM.  In addition I found a pattern using set resource that checks if a package exists in user supplied jars.  With this we could improve JPackage or add import * capabilities to JPype.  It would certainly improve our integration level with Python.  There are certain restrictions such as not being able to find the contents if two jars supply the same package.  And this would only work on Java 9 or later, as Java 8 lacks this support.  But Java 8 is currently past end of life so we may be safe to drop support at some point.

The integration requirements would be to add a new support class in the JPype jar that would 
1) Construct a list of module supported packages.
2) Have a method called `isPakkage` which would check the module list, and if not found check for a jar supplying it.
3) Have a method called `getPackageClasses` which would either open the module or the jar and list the contents of the directory to generate a list of classes.
4) Modify JPackage to check with the support class to decide if a package is real before creating a branch.
5) Modify JImports to check with the support class when there is a wildcard to load all classes.

Does it seem worth the effort to try to include this feature?

```
  static public void getPackages() throws IOException
  {
    FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
    Path modulePath = fs.getPath("modules");
    ModuleFinder finder = ModuleFinder.of(modulePath);
    List<String> out = new LinkedList<>();
    for (Path module : Files.newDirectoryStream(modulePath))
    {
      Path moduleName = module.getFileName();
      if (!moduleName.toString().startsWith("java"))
      {
        continue;
      }
      ModuleReference omref = finder.find(moduleName.toString()).get();
      omref.descriptor().packages().stream()
              .filter(r -> r.startsWith("java"))
              .forEach(r -> out.add(r));
    }
    out.sort(String::compareTo);
    out.forEach(q -> System.out.println(q));
  }
```

Thrameos:
This PR revised the import system.  It adds a Java support class which can detect what is a Java package and its contents.  This allows for wildcard and recursive imports.  There is some additional overhead at startup to compose the list of available modules.  jpype.imports still provides better diagnostics than JPackage but they now share the same machinery.   This is a fairly significant upgrade.

There is one bug fix.  Apparently JPackage was able to import private and protected classes without any warning.  A few of the tests were in fact supposed to be unreachable.  This bug was corrected.

This is actually one of the oldest of my pull requests as the original date for the package determination code was from JPype 0.6.3.   Each year I have taken it out, dusted it off (or started again after my machine got formatted), tried to get it to work, and then been forced to shelve it because of the issues with working with the pre Java 9 issues.  But finally I came across a technical documentation on the replacements made in Java 9 which contained enough technical specs to make it possible to write it for Java 8.  Thus it is finally complete.

Fixes #537


native/common/include/jp_javaframe.h
```diff
@@ -348,6 +348,12 @@ class JPJavaFrame
 	jsize GetStringUTFLength(jstring a0);
 
 	jint compareTo(jobject obj, jobject obj2);
+
+	jboolean isPackage(string str);
```
marscher, line 352: Can't we pass a const string reference here?

Thrameos: Fixed

native/java/org/jpype/pkg/JPypePackageManager.java
```diff
@@ -0,0 +1,322 @@
...
+  private static Path getJarPath(URI resource)
+  {
+    if ("file".equals(resource.getScheme()))
+    {
+      return Paths.get(resource);
+    }
+    if ("jar".equals(resource.getScheme()))
+    {
+      String[] q = resource.toString().split("!");
+      FileSystem zipfs = getJarFileSystem(URI.create(q[0]));
```
marscher, line 286: zipfs cannot be null here?

native/java/org/jpype/pkg/JPypePackage.java
```diff
@@ -0,0 +1,126 @@
...
+    for (String key : contents.keySet())
+    {
+      Object o = this.getObject(key);
+      if (o != null)
+        out.add(key);
+    }
+    return out.toArray(new String[out.size()]);
+  }
+
+  static boolean isPublic(Path p)
```
marscher, line 58: the method returns true if the path contains a public package/class? A short docstring wouldn't hurt. The magic going on here is fascinating :)

native/java/org/jpype/pkg/JPypePackage.java
```diff
@@ -0,0 +1,126 @@
...
+      header.getShort(); // skip major
+      header.getShort(); // skip minor
+      int cpitems = header.getShort();
+
+      // Traverse the cp pool
+      for (int i = 0; i < cpitems - 1; ++i)
+      {
+        is.read(buffer3.array());
+        buffer3.rewind();
+        int type = buffer3.get();
```
marscher, line 80: aint't it a byte?

native/java/org/jpype/pkg/JPypePackage.java
Thrameos: Done.

native/java/org/jpype/pkg/JPypePackage.java
Thrameos: Yep.  I am too used to files where we get an int with a value of -1, but this throws instead.

native/java/org/jpype/pkg/JPypePackage.java
Thrameos: Fascinating because it looks useful, or fascinating because I spent multiple weeks digging through technical documents for over 3 years figuring it out?  😪

native/java/org/jpype/pkg/JPypePackageManager.java
Thrameos: Fixed.

marscher:
From an engineering point of view. These technical documents are publicly available? What a gigantic effort you have taken.

I admire your endurance on this. It is big step into a bright future.

Thrameos:
The key document turned out to be

Thrameos:
The key document turned out to be JEP 220.  It it was discussing the problems that removing the system and introducing the new jrt system.   I was able to get a pattern that worked for Java 9 before but as we were supporting Java 7 and 8 at the time I didn't pursue it as I wanted something universal.   This document gave me enough keys to open the older version once I was able to open the bootloader resources as well as jar resources.   It is always sad when I discover something was possible much earlier, though the difficulty level is much lower now that we have much more code in Java.  I don't think it actually was possible until Java 7 when the zip file system was introduced.

https://openjdk.java.net/jeps/220
https://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html

I am slightly worried that I will run out of open files with the current implementation.  I didn't really want to leave the file systems open but to get to the paths requires access to an open filesystem.  

i guess I could close it and then reopen it again when I need it.  But will that invalidate the paths?

project/jpype_java/test/org/jpype/pkg/JPypePackageNGTest.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Test
+  public void testGetPackage()
+  {
+    assertTrue(JPypePackageManager.isPackage("java.lang"));
+    JPypePackage pkg = new JPypePackage("java.lang", JPypePackageManager.getContentMap("java.lang"));
+    for (Map.Entry<String, Path> e : pkg.contents.entrySet())
+    {
+      JPypePackage.isPublic(e.getValue());
```
marscher, line 32: is that an assertion?

Thrameos: It is just a check for robustness.  The goal of the statement was to make sure nothing would generate an exception while reading a whole bunch of constant pools.  Unfortunately there is no way to know in jar files if you are actually synced up as the only guaranteed structure is the magic number.

I guess we could toss a bunch of known values in, but it would be better to use a structure we control as java/lang can have anything in it.

Thrameos:
I made an attempt at not keeping the Jar files open, but it was really slow (32s per 10k vs 600ms per 10k).  Jar files have to be parse completely when they are opened as file systems due to lack of indexing in the compressed files.  Thus it is pretty prohibitive to reopen them multiple times.  I can add a some better resource management in a second pass after this release.

Thrameos:
Fixed the issue with too many jars open.
Fixed the issue with Java 9 mis-compiling Java 8 binaries.

https://github.com/eclipse/jetty.project/issues/3244

native/common/jp_javaframe.cpp
```diff
@@ -1108,3 +1114,34 @@ jint JPJavaFrame::compareTo(jobject obj, jobject obj2)
...
+
+jobject JPJavaFrame::getPackage(const string& str)
+{
+	jvalue v;
+	v.l = fromStringUTF8(str);
+	JAVA_RETURN(jobject, "JPJavaFrame::getPackage",
+		CallObjectMethodA(m_Context->m_JavaContext.get(), m_Context->m_Context_GetPackageID, &v));
+}
+
+jobject JPJavaFrame::getPackageObject(jobject pkg, string str)
```
marscher, line 1134: const str ref please

marscher:
Please merge this one after resolving the conflicts. It is a major improvement! Thanks a lot!

PR #674, opened 2020-04-08
So long, Cygwin.

Thrameos:
Unfortunately Cygwin has been a bit long in tooth for a while.  Long ago it was the best way to run a Posix like system under windows for those of use who were required run a Window environment.   Today between virtual machines and WSL, there are many better options.

Cygwin has always been problematic.  The Cygwin concept of paths does not play well with Windows paths and the cygpath command is not accessible inside of Python so a user is stuck manually translating pass file paths between Java and Cygwin Python.  Further numerous bugs have shown up which ultimately traced down to bad behaviors in the Cygwin's supplied Python build forcing us to disable tests around broken behaviors in their build.   Last we simply can't get it work properly in the CI tools any longer so it is no longer possible to reliably test what is working.   So three strikes and it is out of here. 

There are still edge cases in which Cygwin as a shell utility is usable but not as when running Python.  One can use Windows Anaconda Python under Cygwin which does not have all the issues.


marscher:
Do you need to add extra bindings for the python class WindowsPath?

Thrameos:
Yeah it looks like I need to test that case.  It must be failing the check for ``__path__`` on 3.5.   I can add a pattern for that case.  I checked my usual windows builds, but I don't usually hit 3.5

PR #676, opened 2020-04-09
Switch over CI

Thrameos:
Removes travis and appveyor from CI.

We will need an additional job for 32 bit windows but lets start by moving coverage and the test.

This PR will be rebased onto #680 as we need the shutdown working to complete the coverage task,


Thrameos:
I can't seem to get the codecov token so that azure can post the coverage.   

https://docs.codecov.io/docs/frequently-asked-questions#section-where-is-the-repository-upload-token-found

PR #680, opened 2020-04-11
Shutdown

### LINKED ISSUE #678: Thread freeze on destroy
Current we do not allow any Destroy hooks to run in Java because we are not terminating the JVM properly.  Investigating the issue, it turns out that Python threads that were attached to Java stop being serviced when the JVM DestroyJavaVM is called.  This creates a deadlock issue as the attached threads are unable to be serviced but shutdown will not proceed until all threads are terminated.   

This is one of the remaining long standing issues as we can't properly handle a shutdown other than manual termination.

Thrameos:
This PR deals with various issues found in a review of shutdown mechanics on JPype.  Previously we did not force a shutdown until an edge case in which a memory handler crashed during a load test.  Now that we are forcing a clean shutdown there are some robustness issues that need to be addressed.

Since some of the earliest releases of this project the shutdown process has been busted... 

```
        // No idea why I can't find this symbol .... no matter, it does not work anyway.
        //JNI_DestroyJavaVM = (jint (__stdcall *)(struct JavaVM_ *))GetAdapter()->getSymbol("DestroyJavaVM");
```

The problem is that JPype is not working with the JVM to shutdown but trying to force the process.   The correct process is to ask for the JVM to shutdown and then wait for each of the Java threads to die, then use a shutdown hook to remove our resources.  But due to a number of flaws this process would deadlock leaving us with no choice but to simply halt the JVM.

This became an issue when we started running code coverage tools.   Those tools need the JVM to shutdown gracefully.   Thus it is now time to address this issue.

This PR fixes the following:
- Threads that are attached automatically are now attached as daemon so that Python spawned threads will not stop a JVM shutdown unless explicitly attached to the JVM.
- ShutdownJVM can only be called from the master thread.
- Attaching Python routines to the Java shutdown using Runtime now works.  
   - Those attached using JPype are called as part of the JPype shutdown sequence
   - Those attached using Java have a race condition so we have to try to force a race loss.
   - Any that fail the race will be given RuntimeError
- The shutdown sequence now is called by the JVM once all but the master thread have been terminated.  
- As the shutdown is now internal to Java, a special flag has to be used to communicate with Python to inform when is the last time a Java method can be called.
- A large number of deadlocks have been removed.
- Jacoco now properly reports and no longer requires any special hooks.
- Some special flags needed to be added to proxies to communicate state with the shutdown hooks.

This PR may produce some issues.  In the old version, we were yanking the plug on Python while Java was fully running thus there were no issues with Java.  But the consequence was that because Java was fully operational, things like timers and threads were still running and could potential crash the whole process.   Now we wait until Java is on its last leg before removing Python resources.  The process is more robust as the number of things that can happen are severely reduced, but by the same token our routines are have to be hardened to operate even when the whole place is about to blow down.

Fixes #678


Thrameos:
@marscher I am not sure if this one addresses the bugs in Kotlin wrappers that was reported on Thrameos/jpype, but it seems like a necessary step for the next release.   Unfortunately we have a lot of PRs that change some fundamental (as well as trivial) things.  Thus we may want to try to gather up a current repo and run with it for a week before.  

I am currently tracking two very difficult issues.   On a German version Visual Studio compiler a user is failing to start in the boot up sequence.  We have had a report like this before, but as it is purely a site issue it very hard to resolve.  Secondarily some users are discovering bugs released to the mandatory shutdown on exit.  The issues were always there but they are only now exposed.   I am not sure if either should hold up a release as we have some serious issues with the last release that need to be resolved.

native/java/org/jpype/JPypeContext.java
```diff
@@ -100,17 +104,131 @@ public static JPypeContext createContext(long context, ClassLoader bootLoader)
...
+    }
+
+    // Kill all remaining Python resources
+    shutdown();
+  }
+
   /**
    * Stop all JPype resources.
    *
+   * This code is parr of a shutdown hook so it should be bomb proof.
```
marscher, line 192: is part of

test/harness/jpype/proxy/Wait.java
```diff
@@ -0,0 +1,16 @@
+package jpype.proxy;
+
+public class Wait {
+    public static void snooze() throws Throwable {
+	    try {
+	    System.out.println("Waiting");
```
marscher, line 6: Debug prints?

native/java/org/jpype/JPypeContext.java
```diff
@@ -71,6 +72,9 @@
   private TypeManager typeManager;
   private JPypeReferenceQueue referenceQueue;
   private ClassLoader bootLoader;
+  private AtomicInteger shutdownFlag = new AtomicInteger();
```
marscher, line 75: this gives birth to a beautiful zero? Java flavored obfuscation :D
Later this is checked for being greater zero, so I assume it does.

native/java/org/jpype/JPypeContext.java
```diff
@@ -100,17 +104,131 @@ public static JPypeContext createContext(long context, ClassLoader bootLoader)
...
+   * This routine will try to take out the last connections in an orderly
+   * fashion. Inherently this is a very dangerous time as portions of Java have
+   * already been deactivated.
+   */
+  private void watchShutdown()
+  {
+    // Try to yield in case there is a race condition.  The user
+    // may have installed a shutdown hook, but we cannot verify
+    // the order that shutdown hook threads are executed.  Thus we will
+    // try to intentionally lose the race.
```
marscher, line 136: lose the race by waiting (n**2 + n)/2 = 15 milli-seconds?

marscher: Its late already, I replaced the 1 with an i :) Where the heck are my glasses=)

native/java/org/jpype/JPypeContext.java
```diff
@@ -100,17 +104,131 @@ public static JPypeContext createContext(long context, ClassLoader bootLoader)
...
+        {
+        }
+      }
+    }
+
+    // Disable all future calls to proxies
+    this.shutdownFlag.incrementAndGet();
+
+    // Wait for any unregistered proxies to finish so that we don't yank
+    // the rug out from under them result in a segfault.
```
marscher, line 174: there seems to be a lot of things which can go horribly wrong :)

native/java/org/jpype/manager/TypeManager.java
```diff
@@ -48,6 +48,9 @@
   public TypeAudit audit = null;
   private ClassDescriptor java_lang_Object;
   public Class<? extends Annotation> functionalAnnotation = null;
+  // For reasons that are less than clear, this object cannot be created
```
marscher, line 51: Doesn't it make sense for the JVM to block the creation of new objects during shutdown?

test/harness/jpype/proxy/Wait.java
```diff
@@ -0,0 +1,16 @@
...
+	    Thread.sleep(1000);
+	    System.out.println("Wake");
+	    }
+	    catch (Throwable th)
+	    {
+		    System.out.println("Catch "+th);
+		    throw th;
```
marscher, line 13: is this exception handled elsewhere now? Why the rethrow?

native/java/org/jpype/manager/TypeManager.java
Thrameos: I would have assumed it would have run it immediately if we are in shutdown.  Preventing another hook means we generate an exception during a fragile time.  But maybe that I think about it differently.    It creates lots of threads and objects at shutdown.

test/harness/jpype/proxy/Wait.java
Thrameos: Wait is a test code that I was using to print a count down..  I will remove it.

native/java/org/jpype/JPypeContext.java
Thrameos: Yeah.  There are hoards of things that can break.   Which is way I had to try to redirect them to a list I know so that I can join all of them.

native/java/org/jpype/JPypeContext.java
Thrameos: Yep

```
Creates a new AtomicInteger with initial value 0.
```

Everything is born as nothing in Java.  I was a little hesitant to add a counter to the proxies.   Unfortunately a few of my test cases found cases in which I would seg fault if I didn't make sure the proxy was exited.   A full lock system was overkill and potentially rather slow.  So I opted for atomics.

test/harness/jpype/proxy/Wait.java
Thrameos: Debug file.   Not used.

native/java/org/jpype/JPypeContext.java
Thrameos: I know that it seems odd to wait 1 ms but when the threading scheduler sees a sleep or a yield it should check to see if there are any other threads ready before coming back to this thread.    All of the threads are born during the shutdown hook and we just need to stall long enough for any unaccounted threads to hit the atomic integer increment.   Once the first one does we are safe to proceed.  

I debated pulling a longer sleep.  But most cases don't need this and it is just adding time to our shutdown.  I could also have implemented a find all Java threads code here.   But I didn't want to risk an exception should module protections be tripped.   (Yes there are a lot of horrible things to worry about here.)

marscher:
This is a major improvement, as it gracefully shuts down the JVM and (parts?) of the C++ layer. I think we should carefully test this. Eventually we should provide some release candidates for 0.8 and ask big projects to give us feedback. I hope this will not delay the final release too much.

Lets begin and tag everything we want to have in 0.8 and mark the rest for another milestone. By this I mean a feature freeze (I like the proposal for the new cast operators etc., but as you said, it might be wiser to postpone it).

Thrameos:
Sounds good.  I anticipated that we would do a release 0.7.3 shortly which would be the last with the deprecated functions rather than just jumping to 0.8 where we actually remove some bad stuff, but it is pretty muddy as we are sort of half way there.  I don't really want to stall 0.7.3 as 0.7.2 has a bad bug with Exceptions.  My customer is having a big issue with the memory exhaustion so I need to get a release to support them.

So one plan would be reapply just bug fixes starting from 0.7.2 and make a release 0.7.3 with minimal changes.  That way we buy a little time to settle issues for 0.8.  I am trying to get the documentation complete, but as I am writing I am turning up a lot of additional edge cases (Lists aren't sequences, null fails here, etc).  I did a very hard press in March but the doc review is turning up another round that needs to be dealt with.

Alternatively we can try to push to 0.8.0 and then we finalize the API at 0.8.1.  I am not much in favor of that.

Just point me which direction you would like and I will try to make it happen this weekend.

marscher:
I just thought to do a minor since the shutdown thing really changes some behavior users might implicitly depend on (which would be horrible of course). A minor would at least indicate some braoder functionality changes. 
But I can also see your need for a micro on 0.7. And the road to 0.8 seems more reasonable and should produce a more stable result. Is the change log file actually up to date?

Thrameos:
I believe the changelog is mostly up to date.  I have been putting everything at user level in (though sometimes that is causing conflicts as most stuff is added at the end).  I should a scan of the PR to see if anything was missed.

How about we do a major/minor split using the change log?    If it seems like something should be held off we move it to the "next, next release" section and leave all the rest where they are.    That will give me a plan to issue the micro release.  I can then do one merge step and push a release candidate branch into place. 

I think the shutdown behavior and the terminate should likely push to 0.8.

marscher:
Using the changelog sounds like a good option as it reflects the users perspective.

PR #682, opened 2020-04-12
Compact syntax

Thrameos:
This pull request contains a few items.

- A fix for a minor problem with the coverage tool.  (it should have marked the lines as GCOVR_EXCL
  rather than bypassing the check)
- A bad entry in PyJPArray table (GetItem does not implement the proper contract, but Python is not checking that slot as there is another with a higher presidence)
- Introduces an easy way to define arrays using operator index ([]) with syntax much more similar to Java.
- Introduces a cast operator using the matmul operator (@).

Some of the notation that we use in JPype is a bit awkward when compared to Java.  Consider the following code:

```
    int[][] i = new int[10][5];
    int[][] j = new int[10][];
    Object o = new String("foo");
    bar( (String) o);
```

This pull request uses the Python operators to introduce two new operations.  Which are cast (@) and type index ([]).  Thus translating the above code to...

```
    i = JInt[10,5]       # i is an int[][] with dimensions 10x5
    j = JInt[10,:]        # j is an int[][] with dimension 10xunspecified
    o = JObject@JString("foo")  # o is an java.lang.Object
    bar(JString@o)       # cast to String
    bar(JString@None)       # cast null to String
    bar(java.lang.String@None)       # cast null to String with explicit type
    bar(JObject@JInt(10))     # create a java.lang.Integer with value 10  
```

Obviously there is more than one way to skin a cat here as we could use the constructor rather than casting in some places.  

as opposed to the old notation.
```
    i = JArray(JInt,2)(10)
    for q in range(10):
        i[i] = JArray(JInt,2)(5)
    j = JArray(JInt,2)(10)
    o = JObject(JString("foo"), JObject)
    bar(JObject(o, JString))
    bar(JObject(JInt(10)))     # create a java.lang.Integer with value 10
```

I am hoping this will make the code look much more clear to the user.  Obviously we can't actually define casting directly with Python syntax, but the mat multiply is pretty similar.  We have disabled ``@=`` so that it doesn't get used for Java types.

With this types have the following defined operations
 - ``Class(*args)`` - construct an object from the class
 - ``Class@obj`` - cast an object to the class
 - ``Class[#]`` - create a defined one dimensional array
 - ``Class[:]`` - define a array type with one dimension
 - ``Class[#,#..]`` - create a multi dimensional array
 - ``Class[:,:,...]`` - define a array type with multiple dimensions
 - ``Class[#,.., :,..]`` - create a multi dimensional array with uninitialized dimensions

I am not sure it this one should wait for the next release.  Of perhaps this one isn't really a good idea.


Thrameos:
@marscher I am putting this up for review.   If it looks like a good idea I will update the docs and tests appropriately and push additional commits.   I will also pack port some of the bug correction into another PR.

Thrameos:
Another option if we don't want to use matmul would be to use floor div.

Consider Java

``` java
  obj.callMethod( (java.lang.List) p1, (int) p2);
```

Written in Python as either

``` python
   obj.callMethod(  java.lang.List//p1, JInt//p2)
```

vs

``` python
   obj.callMethod(  java.lang.List@p1, JInt@p2)
```

The other operators are too common, but these two are low enough frequency not to be confused.

Thrameos:
I did a survey of my coworkers to see what would make the most sense in terms of operators.   Candidates where ``@``, ``%``, ``//``.   All of them preferred ``@`` but that was likely due to it being the least familiar thus the least common.   ``//`` was considered the worst because it looks like a Java comment symbol.   I thought that maybe ``%`` would make sense to them as it denotes string substitution, but it was roundly rejected.

There are a few that are options but have other connotations such as ``<<`` and ``**``.

https://www.geeksforgeeks.org/basic-operators-python/

Thrameos:
@pelson, @michi42, @KOLANICH, @bujuin Do you have an opinion on a Python operator to use for type casting?   We are currently planning on using ``@``.   Does that seem okay for Scala and Kotlin, or are there operator connotations there that I am not aware or?  If there are no objections I plan to merge this and update the documentation accordingly.

Thrameos:
Still haven't gotten any user input.  But I generally feel this would be a positive development in terms of clarity.  I am going to freshen and add the needed documentation.

PR #685, opened 2020-04-15
Rework JPype documentation

### LINKED ISSUE #432: User Guide v2
My coworker feels that our current guide is poorly organized.  It is in random order, doesn't define terms, is *not* compelling, and buries the most important information.

He recommends reordering it something like

- Introduction
  - What and why JPype
  - Use cases examples
  - Limitations
- Basic usage
  - Definition of terms (Class, ClassFactory, wrapper, etc)
  - Importing and Class factories
  - Class, fields, methods, overloading , casting
  - Type conversions
  - Arrays, inner classes
- Advanced usage
  - Threading
  - Proxies
  - Customizers
  - Performance tips

Obviously not something to bite off before 0.7 release, but certainly a goal.

### LINKED ISSUE #433: `__hash__` on Java Objects
Currently ``__hash__`` is defined on a generic Java object type.  As a result any  Java object can be used as a dictionary key even if the object is not immutable.  Not sure how desirable this behavior is.  It appears to be consistent with Java though, as it appears there is no issue storing a list as a Java hash key.   Either way it should be noted in the documentation.

### LINKED ISSUE #455: Document use of autopep8 on JPype
Autopep8 seems to like to reorganize the imports which breaks jpype python files by moving the java imports above the startJVM statement.  This can be avoided by either passing in ``--ignore E402`` or setting the ignore in ``.pep8``.   This should be documented in our userguide someplace.

Example:
```
import jpype
import jpype.imports

jpype.startJVM()

from gov.llnl.math import DoubleArray
```

Result without ignore E402
```
from gov.llnl.math import DoubleArray
import jpype
import jpype.imports

jpype.startJVM()
```

### LINKED ISSUE #463: Debugging tricks

### LINKED ISSUE #518: getClassPath after jvm started
Currently we have a few ways of specifying the classpath.  The older direct method was directly specifying 'java.class.path' to the jvm.  It was augmented with a system based off of Matlab in which the classpath is set up with a series of commands prior to calling startJVM.  Things like globbing and pulling from the environment were only in the Matlab based one.  But in JPype 0.7 we modernized by pulling most of the Matlab type features into a keyword args based system.  This new API does not cover all the cases such as construction something using both the environment based path and user options.  So the old Matlab one is still a necessary system and is still quite useful.

But when we added the keyword method, it became much easier to use than the Matlab based one thus it is seeing a lot more use.  However, the kwargs system and the Matlab based system were never designed to work together after the JVM was started.  This leaves the impression that one can use `getClassPath()` to find out the classpath that was used to start the JVM that was specified with the kwargs.  Instead it gives the classpath that would have been used if you did not manually specify it.

The correct procedure has always been to call ``jpype.java.lang.System.getProperty('java.class.path')`` to find out what the loaded classpath was.

It was suggested that rather than always returning the default path (env based with push user specifications) the call to getClassPath should check to see if the JVM is running and if it is get the classpath that was actually used to start the JVM instead of the default one.  It seems reasonable, but I am worried that this will have interactions in the future especially if we have more than one jvm started.  After all if you start a second JVM should it automatically default to the path of the current running one or not?  Alternatively we could just improve the documentation to clarify this issue.

Any opinions on the correct direction for this would be welcomed.

Thrameos:
Okay no getting around this one; eventually we have to update the documentation.  This was the last big task on the JPype 0.8.0 list.  Sorry, this one is going to be a pain to review.

My son is review for clarity and grammar.  Hopefully completed by tomorrow afternoon.

Fixes #432
Fixes #455 
Fixes #433 
Fixes #518 
Fixes #463


doc/userguide.rst
```diff
@@ -1,259 +1,1509 @@
...
+   :maxdepth: 2
+
+JPype Introduction
+******************
+JPype is a Python module to provide full access to Java from within Python. 
+Unlike Jython, JPype does not achive this by re-implementing Python, but
+instead by interfacing both virtual machines at the native level. This
+shared memory based approach achieves good computing performance, while
+providing the access to the entirety of CPython and Java libraries.
+This approach allows direct memory access between the two machine,
```
marscher, line 15: machines

doc/userguide.rst
```diff
@@ -1,259 +1,1509 @@
...
+implementation of Java interfaces in Python, and even use of Java threading.
+
+JPype Use Cases
+===============
+
+Here are three typical reasonse to use JPype.
```
marscher, line 21: reasons

doc/userguide.rst
```diff
@@ -1,259 +1,1509 @@
...
-Eventually, it should be possible to replace Java with Python in many, though
-not all, situations. JSP, Servlets, RMI servers and IDE plugins are all good
-candidates.
+   # Boiler plate stuff to start the module
+   import jpype
+   import jpype.imports
+   from jpype.types import *
 
+   # Launch the JVM
+   jpype.startJVM(classpath=['jars/database.jar'])
```
marscher, line 88: shouldn't we pass convertStrings here in order to avoid a UserWarning?

marscher: Or keep it as simple as possible?

Thrameos: The guide was written assuming that we were releasing a 0.8 so some of the deprecated behavior was being removed.

Thrameos:
@marscher okay I believe this is ready for review.

PR #687, opened 2020-04-16
Handle nulls in strings

Thrameos:
When working on documentation problems were found with handling null.   This PR deals with those issues.  It also fixes a crash in comparable during richcompare.

Tests are include, doc will be in other PR.

### DIFF
--- a/jpype/_jstring.py
+++ b/jpype/_jstring.py
@@ -56,7 +56,7 @@ class _JStringProto(object):
         return self.contains(other)
 
     def __hash__(self):
-        if self == None:
+        if self == None: # lgtm [py/test-equals-none]
             return hash(None)
         return self.__str__().__hash__()
 
--- a/jpype/_jstring.py
+++ b/jpype/_jstring.py
@@ -40,16 +40,6 @@ class _JStringProto(object):
     def __add__(self, other):
         return self.concat(other)
 
-    def __eq__(self, other):
-        if isinstance(other, JString):
-            return self.equals(other)
-        return str(self) == other
-
-    def __ne__(self, other):
-        if isinstance(other, JString):
-            return not self.equals(other)
-        return str(self) != other
-
     def __len__(self):
         return self.length()
 
@@ -62,22 +52,12 @@ class _JStringProto(object):
             raise IndexError("Array index exceeds length")
         return self.charAt(i)
 
-    def __lt__(self, other):
-        return self.compareTo(other) < 0
-
-    def __le__(self, other):
-        return self.compareTo(other) <= 0
-
-    def __gt__(self, other):
-        return self.compareTo(other) > 0
-
-    def __ge__(self, other):
-        return self.compareTo(other) >= 0
-
     def __contains__(self, other):
         return self.contains(other)
 
     def __hash__(self):
+        if self == None:
+            return hash(None)
         return self.__str__().__hash__()
 
     def __repr__(self):
--- a/native/python/pyjp_object.cpp
+++ b/native/python/pyjp_object.cpp
@@ -158,7 +158,7 @@ static PyObject *PyJPComparable_compare(PyObject *self, PyObject *other, int op)
 		if (!cls2->findJavaConversion(match))
 			JP_RAISE(PyExc_TypeError, "Type is not comparable");
 		obj1 = match.convert().l;
-	} else if (!null1)
+	} else if (!null1 && javaSlot1 != NULL)
 		obj1 = javaSlot1->getValue().l;
 
 	switch (op)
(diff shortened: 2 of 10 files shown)


native/python/pyjp_object.cpp
```diff
@@ -129,7 +129,7 @@ static PyObject *PyJPComparable_compare(PyObject *self, PyObject *other, int op)
 	if (self == Py_None || javaSlot0 == NULL ||
 			(!javaSlot0->getClass()->isPrimitive() && javaSlot0->getValue().l == NULL))
 		null0  = true;
-	if (other == Py_None || ( javaSlot0 != NULL &&
+	if (other == Py_None || (javaSlot1 != NULL &&
```
Thrameos, line 132: There was a copy and paste error.  Corrected.

native/python/pyjp_object.cpp
```diff
@@ -138,8 +138,6 @@ static PyObject *PyJPComparable_compare(PyObject *self, PyObject *other, int op)
 
 	if (!null0)
 		obj0 = javaSlot0->getValue().l;
-	if (!null1)
-		obj1 = javaSlot1->getValue().l;
```
Thrameos, line 142: These lines need to move down.

native/python/pyjp_value.cpp
```diff
@@ -169,6 +169,9 @@ PyObject* PyJPValue_str(PyObject* self)
 	if (cls->isPrimitive())
 		JP_RAISE(PyExc_TypeError, "toString requires a Java object");
 
+	if (value->getValue().l == NULL)
+		return JPPyString::fromStringUTF8("null").keep();
+
```
Thrameos, line 174: Nulls now covert to "null" when calling string rather than value error.

jpype/_jstring.py
```diff
@@ -40,16 +40,6 @@ class _JStringProto(object):
     def __add__(self, other):
         return self.concat(other)
 
-    def __eq__(self, other):
-        if isinstance(other, JString):
```
Thrameos, line 44: These all fall back to comparable.  There may be interactions with hashcode.   This needed to clear CI on all versions.

jpype/_jstring.py
```diff
@@ -62,22 +52,12 @@ def __getitem__(self, i):
...
-        return self.compareTo(other) > 0
-
-    def __ge__(self, other):
-        return self.compareTo(other) >= 0
-
     def __contains__(self, other):
         return self.contains(other)
 
     def __hash__(self):
+        if self == None: # lgtm [py/test-equals-none]
```
marscher, line 59: it complains, because a None check in Python should be written as "is None" (for whatever reason).

native/python/pyjp_value.cpp
marscher: Why would a mapping to an empty string be a bad idea in this case? I mean I'm fine with the "null" string too, but mapping nothing to a real word seems somehow odd. Is it a Java convention to do this mapping?

test/harness/jpype/attr/Test1.java
```diff
@@ -62,11 +62,11 @@ public static String testStaticHolder(Holder h)
 	
 	public String stringValue = "Foo";
 	public char charValue = 'a';
-	public static Object objectValue= new Integer(234);
+	public static Object objectValue= (int)234;
```
marscher, line 65: why do you need to trigger autoboxing? The result is still a static object

Thrameos: For reasons that are beyond me Integer(int) is marked deprecated after Java 9.   Instead they want you use valueOf. But that doesn't exist on the earlier version.  So the best I could do to fit both is autoboxing.   It seems really silly to me.

jpype/_jstring.py
Thrameos: Yes they have it tagged as a performance warning.    True the None, True, and False are singletons, but that doesn't mean "is" is always the right answer.   I didn't feel like making a special function just for checking for null.

native/python/pyjp_value.cpp
Thrameos: The problem is if you don't make it "null" then you get a ValueError when you are printing.

Consider the following Java code
``` java
  {
    Object o = null;
    System.out.println("A " + o);
  }
```

The correct output is "A null".   So to str(o) is "null".

test/harness/jpype/attr/Test1.java
marscher: yes, that's hilarious

native/python/pyjp_value.cpp
marscher: I see.

PR #690, opened 2020-04-17
Support for arrays and JNI names in JClass

Thrameos:
Simple enhancement to JClass to handle arrays and JNI names properly.

This enhancement fixes 
https://stackoverflow.com/questions/60906495/how-do-i-build-my-java-classes-to-not-result-in-noclassdeffoundexception-from-jp/61082786#61082786

### DIFF
--- a/test/jpypetest/test_jclass.py
+++ b/test/jpypetest/test_jclass.py
@@ -204,3 +204,21 @@ class JClassTestCase(common.JPypeTestCase):
         test = JClass("jpype.types.InnerTest")()
         test.test()
 
+    def testLookupGeneric(self):
+        self.assertEqual(JClass('java.util.ArrayList<>'), JClass("java.util.ArrayList"))
+
+    def testLookupJNI(self):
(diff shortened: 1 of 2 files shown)


test/jpypetest/test_jclass.py
```diff
@@ -204,3 +204,21 @@ def testInnerClass(self):
         test = JClass("jpype.types.InnerTest")()
         test.test()
 
+    def testLookupGeneric(self):
```
marscher, line 207: Is a not-found class mapped to None? Or what is the desired outcome?

Thrameos: Not found should raise if I remember correctly.

Thrameos:
We do not use JNI names except when calling bootstrapping with FindClass.   Given some users appear to have attempted to use JNI names in JClass (which isn't really a terrible thing to guess), I figured it costs nothing to support.

There will be a few minor conflicts with the package update which I will resolve in another PR.  This PR was written when we were guessing what is a package.   That won't be necessary once the other PR is in place.

PR #699, opened 2020-04-23
Fixed repr

### LINKED ISSUE #698: Modification of __repr__
Hi, testing jPype and looks rgood!

I am mostly using ipython/jupyter and the __repr__ string is then important as one is very interactive with the objects. Is there a way to change all __repr__ strings for objects, I can do it on a single class with JImplementationFor(classname) but would like to do it for all classes or a set of them? Adding a method like:

def __repr__(self):
        return f"<{self.getClass().getSimpleName()}: {self.toString()}>"

Thrameos:
Deals with unifying repr strings.

Fixes #698

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -3,6 +3,11 @@ Changelog
 
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
+- **Next Release**
+
+  - Updated the repr methods on Java classes.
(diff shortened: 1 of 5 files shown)


Thrameos:
A test case failed in another branch that indicates that this can't handle a certain null input.  I need to track it down before proceeding.

Thrameos:
Checking finds no issues, so it is ready for review.

PR #703, opened 2020-04-24
Fix shutdown sequence

Thrameos:
This addresses two bugs in the shutdown sequence:

- Daemon threads were still running and could execute during resource cleanup
- The shutdown flag was not being respected due to a bug in the parameter list in onShutdown

To improve this 
- Make before break was added to JProxy to prevent a race condition
- JPype attempts to send interrupt to all threads so that they may be serviced.
- disabling all Java calls was moved up in the sequence

This was tested against a series of sleep/call loops.   Python does not respect any sort of interrupt code as signals all divert to the main thread and they didn't plan for interruptions of blocking calls like sleep.  Therefore, it is impossible to make it so that Python code can't stall the process.  Further, there are always going to be edge cases (monitor held during interruption, exception thrown in exception handling).   But those are really user errors for failing to plan an orderly shutdown, and not really the sole responsibility of library.

### DIFF
--- a/native/common/include/jp_context.h
+++ b/native/common/include/jp_context.h
@@ -140,11 +140,6 @@ public:
 		return m_JavaContext.get();
 	}
 
-	bool isShutdown()
-	{
-		return m_IsShutdown;
-	}
-
 	/** Release a global reference checking for shutdown.
 	 *
 	 * This should be used in any calls to release resources from a destructor.
@@ -223,7 +218,7 @@ private:
 	jint(JNICALL * CreateJVM_Method)(JavaVM **pvm, void **penv, void *args);
 	jint(JNICALL * GetCreatedJVMs_Method)(JavaVM **pvm, jsize size, jsize * nVms);
 
-	static JNIEXPORT void JNICALL onShutdown(JNIEnv *env, jlong contextPtr);
+	static JNIEXPORT void JNICALL onShutdown(JNIEnv *env, jobject obj, jlong contextPtr);
 
 private:
 	JPContext(const JPContext& orig);
@@ -262,8 +257,7 @@ private:
 	jmethodID m_Context_OrderID;
 	jmethodID m_Object_GetClassID;
 private:
-	bool m_IsShutdown;
-	bool m_IsInitialized;
+	bool m_Running;
 	bool m_ConvertStrings;
 public:
 	JPGarbageCollection *m_GC;
--- a/native/java/org/jpype/JPypeContext.java
+++ b/native/java/org/jpype/JPypeContext.java
@@ -19,11 +19,13 @@ package org.jpype;
 import java.lang.reflect.Array;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.nio.Buffer;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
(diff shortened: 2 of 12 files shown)


native/common/include/jp_context.h
```diff
@@ -223,7 +218,7 @@ class JPContext
 	jint(JNICALL * CreateJVM_Method)(JavaVM **pvm, void **penv, void *args);
 	jint(JNICALL * GetCreatedJVMs_Method)(JavaVM **pvm, jsize size, jsize * nVms);
 
-	static JNIEXPORT void JNICALL onShutdown(JNIEnv *env, jlong contextPtr);
+	static JNIEXPORT void JNICALL onShutdown(JNIEnv *env, jobject obj, jlong contextPtr);
```
Thrameos, line 221: This was the bug on the JNI call.  We set the flag on the wrong object.

native/common/include/jp_context.h
```diff
@@ -140,11 +140,6 @@ class JPContext
 		return m_JavaContext.get();
 	}
 
-	bool isShutdown()
-	{
-		return m_IsShutdown;
-	}
-
```
Thrameos, line 147: We do not really need 2 flags for one state.

native/java/org/jpype/JPypeContext.java
```diff
@@ -128,75 +130,98 @@ public void run()
...
-   * Stop all JPype resources.
-   *
-   * This code is part of a shutdown hook so it should be bomb proof.
-   *
-   */
-  void shutdown()
-  {
-    // Inform Python no more calls are permitted
-    onShutdown(this.context);
-
```
Thrameos, line 199: We don't need to split the shutdown.  Just confuses things.

native/java/org/jpype/ref/JPypeReferenceQueue.java
```diff
@@ -68,7 +68,7 @@ public void registerRef(Object javaObject, long host, long cleanup)
   public void start()
   {
     isStopped = false;
-    queueThread = new Thread(new Worker());
+    queueThread = new Thread(new Worker(), "Python Reference Queue");
```
Thrameos, line 71: Name the queue so we can find it while debugging

native/java/org/jpype/proxy/JPypeProxy.java
```diff
@@ -68,28 +68,28 @@ public Object newInstance()
...
-    synchronized (typeManager)
+    try
     {
-      returnType = typeManager.findClass(method.getReturnType());
-      Class<?>[] types = method.getParameterTypes();
-      parameterTypes = new long[types.length];
-      for (int i = 0; i < types.length; ++i)
+      context.incrementProxy();
+      if (context.isShutdown())
+        throw new RuntimeException("Proxy called during shutdown");
```
Thrameos, line 75: Make before break was needed here.

native/java/org/jpype/JPypeContext.java
```diff
@@ -128,75 +130,98 @@ public void run()
...
+      Thread t1 = Thread.currentThread();
+      Map<Thread, StackTraceElement[]> threads = Thread.getAllStackTraces();
+
+      for (Thread t : threads.keySet())
       {
-        thread.start();
+        if (t1 == t)
+          continue;
+//      if (t.getState() == Thread.State.RUNNABLE)
+        t.interrupt();
```
Thrameos, line 191: Everyone gets an interrupt so that nothing is left waiting.  This doesn't hit Python unless it was waiting on a blocking Java call.

PR #705, opened 2020-04-25
Exception chaining and other improvements

### LINKED ISSUE #612: Java exception causes do not appear in Python cause list
The improved exception handling while a big improvement does not currently capture the full Java cause list in the Python output forcing the user to open the Java exception to get additional diagnostics.   This can be improved by converting not just the original source exception but also any cause information.

### LINKED ISSUE #697: Imports does not properly report java.lang.ExceptionInInitializerError
Example

```
 File "TypeManager.java", line 166, in org.jpype.manager.TypeManager.findClassByName
  File "Class.java", line 315, in java.lang.Class.forName
  File "Class.java", line -2, in java.lang.Class.forName0
Exception: Java Exception

  File "test2.py", line 6, in <module>
    db= jpype.JClass('org.my.Database')

```

Actual cause was 

```
Caused by: java.nio.file.NoSuchFileException: ./app.properties
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
        at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:219)
        at java.base/java.nio.file.Files.newByteChannel(Files.java:370)
        at java.base/java.nio.file.Files.newByteChannel(Files.java:421)
        at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:420)
        at java.base/java.nio.file.Files.newInputStream(Files.java:155)
        at org.my.getProperties(ApplicationProperties.java:53)
        ... 4 more
```

Thrameos:
This started as a simple bug fix because I was not getting the right exception on the output when I had a bug in code.  Some 6 hours later I believe it is fixed.   To make it work required making exceptions chain themselves properly so that added Java cause is not getting lost.

Fixes #697
Fixes #612


native/common/jp_context.cpp
```diff
@@ -171,14 +199,16 @@ void JPContext::startJVM(const string& vmPath, const StringVector& args,
...
+		m_Throwable_GetCauseID = frame.GetMethodID(cls, "getCause", "()Ljava/lang/Throwable;");
+
 		// After the JVM is created but before the context is started, we need
 		// to set up all the services that the context will need.
 		JP_TRACE("Initialize");
 
 		// We need these first because if anything goes south this is the first
 		// thing that will get hit.
-		jclass cls = frame.FindClass("java/lang/Object");
+		cls = frame.FindClass("java/lang/Object");
```
marscher, line 211: Is this some kind of variable recycling?

native/common/jp_exception.cpp
```diff
@@ -210,7 +209,18 @@ void JPypeException::convertJavaToPython()
...
-	JPPyObject trace(JPPyRef::_call, PyTrace_FromJavaException(frame, th));
+	JPPyObject trace = PyTrace_FromJavaException(frame, th, NULL);
+
+	// Attach Java causes as well.
+	jthrowable jcause = frame.getCause(th);
+	if (jcause != NULL)
+	{
+		jvalue a;
+		a.l = (jobject) jcause;
+		JPPyObject prev = frame.getContext()->_java_lang_Object->convertToPythonObject(frame, a, false);
```
marscher, line 220: can this call raise something?

native/common/jp_context.cpp
Thrameos: Yeah, cls gets used about 20 times here with each class.  I suppose I can rename it new names each time.   I just don't want any of these items to get cycled into globals like we had before.

native/common/jp_exception.cpp
Thrameos: In theory yes, in practice no.   Calling convertToPythonObject can fail if the object type does not exist for some reason.  But exception types should always be wrappable unless there is something very wrong in the Python layer.   Unfortunately recovering from a double exception throw is very hard, so the best I can do here is catch the bad object  and bail from this path so that the original exception gets spat out.   

I will work on a revision.

PR #707, opened 2020-04-26
Java classes are closed

Thrameos:
This PR adds some type markers so that Java classes cannot be extending from Python accidentally.  This should reduce the possibilities of errors.

It also includes a missing autopep8.


Thrameos:
Drat.  Will need to reformulate to work with 3.5   Guess on hold until I can cycle back to it.

Thrameos:
@marscher  The fix was easier than I was thinking (literally just changing an argument to NULL).    So this one is ready for review.

PR #708, opened 2020-04-26
List and Collection as collections.abc

Thrameos:
Minor PR to upgrade List and Collection to better fit Python collections model.  We can't complete the contract on MutableSequence due to a conflict on remove.  I could rename the Java remove to ``remove_`` if we want to complete that one as well.


test/jpypetest/test.py
```diff
@@ -0,0 +1,3 @@
+import platform
```
marscher, line 1: delete me :)

Thrameos: Ouch sorry.  I don't know how those two found their way into the commit.

marscher:
It is good not to break contracts. Would you please autopep8 your python files?

jpype/_jcollection.py
```diff
@@ -54,6 +54,12 @@ def __delitem__(self, i):
         raise TypeError(
             "'%s' does not support item deletion, use remove() method" % type(self).__name__)
 
+    def __contains__(self, i):
+        try:
+            return self.contains(i)
```
marscher, line 59: Why is it a TypError? I would have expected a KeyError, if i isn't contained.

Thrameos: When we call a Java method and it can't be converted then it produces a TypeError.  We have to catch that TypeError and return a false, so that we can meet the Python contract for contains.

Thrameos:
Autopep8 is complete.   So just for clarification....  would you prefer?

- [ ] To rename the Java method ``remove`` to ``remove_`` so that we can complete the Python contract for ``collections.abc.MutableSequence``, but will confuse Java programmers.  Both Python and Java sematics are then possible (with Java programmer forced to call ``remove_`` due to interface conflicts.)
- [ ] Leave ``remove`` as is as making it have Python remove semantics would break code, but will confuse Python programmers.

As it stands the Java contract is basically a silent failure currently.  It doesn't raise a ValueError and expects someone to check the return which isn't a Python idiom.

marscher:
I'm OK with confusing Java programmers. Even though JPype can be now be used bidirectional, I think most users will use it to call Java code from within Python.

Issue #672, opened 2020-04-08
startJVM fails without any message with Python 3.8 32bit and jre1.8.0_241 on Windows 10


Matthias1975:
It seems, that it fails in jp_exception.cpp (line 493):

![image](https://user-images.githubusercontent.com/63334343/78748762-1a6d7a80-796d-11ea-8c01-57ac4384c852.png)

There is no exception caught in the catch-Block at jp_env.cpp (line 184):

![image](https://user-images.githubusercontent.com/63334343/78748810-307b3b00-796d-11ea-8cd3-5d3e9fd0f818.png)

jpype1 was buildt with:
![image](https://user-images.githubusercontent.com/63334343/78749123-e8104d00-796d-11ea-9e45-530d65eb82ab.png)

the result is:
![image](https://user-images.githubusercontent.com/63334343/78749144-f5c5d280-796d-11ea-8a6a-24098868e9b3.png)

What could be the reason?

Matthias1975:
Windows Event Viewer shows:
Name der fehlerhaften Anwendung: python.exe, Version: 3.8.2150.1013, Zeitstempel: 0x5e55a3ed
Name des fehlerhaften Moduls: _jpype.cp38-win32.pyd, Version: 0.0.0.0, Zeitstempel: 0x5e8d6107
**Exception Code: 0xc0000409
Error offset: 0x0001696d**
ID des fehlerhaften Prozesses: 0x5694
Startzeit der fehlerhaften Anwendung: 0x01d60d69a19940cf
Pfad der fehlerhaften Anwendung: c:\DEVTOOLS\Python\38-32\python.exe
Pfad des fehlerhaften Moduls: c:\DEVTOOLS\Python\38-32\lib\site-packages\jpype1-0.7.2-py3.8-win32.egg\_jpype.cp38-win32.pyd
Berichtskennung: e758c246-05c9-477c-9272-a4715b5a616a
Vollständiger Name des fehlerhaften Pakets: 
Anwendungs-ID, die relativ zum fehlerhaften Paket ist: 

Regarding to the link below 0xc0000409 means STATUS_STACK_BUFFER_OVERRUN.
https://stackoverflow.com/questions/23409809/windows-7-exception-code-0xc0000409

Thrameos:
Ouch that is a bad place to throw an error.   At that particular point we don't yet have a working set of resources to handle an exception.  But the fact that the exception isn't caught is a general indication that something happened during the JNI call.  So why would it not be caught?

[...]

PR #709, opened 2020-04-26
Updated docs for problems with Path

### LINKED ISSUE #221: Jpype on Windows 2016 Server
I currently run a couple of tests with Jpype under different versions of Windows. While Jpype works fine on Windows 2012 Server, I cannot start the JVM through Jpype on Windows Server 2016. In both cases, I cloned the git repo and built it from source. JAVA_HOME points to the correct path.
Does anyone have a clue what could have possibly gone wrong?

System setup:
- Windows 2016 Server 64-bit
- JRE 1.8 64-bit
- Visual C++ Built Tools
- Anaconda 64-bit w/ Python 3.6

screen shot 2017-06-03 at 24 31 01

Thrameos:
I found this old issue #221 which may be related to #672.  I am going to document the fix for now.

Fixes #221

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -164,6 +164,18 @@ From the Python developer page:
 When building for windows you must use the Visual Studio developer command
 prompt.
 
+
+Path requirements
+-----------------
+
+On certain systems such as Windows 2016 Server, the JDK will not load properly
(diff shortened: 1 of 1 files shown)


marscher:
Thanks!

PR #715, opened 2020-04-28
Quickfix

### LINKED ISSUE #712: Resource leak on printf
While investigating the crash on Spyder, I ran into an issue in which the sample code was leaking a local reference.   Everything in the local call area has appropriate JPJavaFrame guards so it must be something going in from running in interactive mode.   Definitely needs fixing.

Thrameos:
While tracing through another bug report I found to my horror that there was a problem with local referencing which causes arrays to leak references both when used directly and when used as part of argument handling which is a major problem.   I recommend that we do a 0.7.4 release to deal with this significant issue. 

Fixes #712 

The fix for the issue was to remove two incorrect keep statements.   Finding it was much more challenging.  The resource in this case was being lost during the middle of a call to variable arguments.  The result was the Java frame was now the root frame and pushing all the local references that were after the bad call to global scope.   I am not sure why Java doesn't flag this type of problem directly.  There is no case in which creating local references without a frame is acceptable so ``-Xcheck:jni`` should have caught it.  To identify this hairy issue I had to upgrade the trace utilities to report python, java and trace events and allow them to be selectively controlled.  Once it was clear the source, I developed an internal consistency check which is evaluated whenever instrumentation or tracing is enabled.  Creating a local reference outside of Java frame will now trigger an exception.   The only case where this does not happen is on the exception paths where throwing an exception is likely to be fatal and thus lead to a lack of diagnostics.   This modification ensures that the current test bench will fail if this issue ever reappears.

I backported this from the head so it excludes all of the other changes on master.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,6 +4,7 @@
 
 pr:
   - master
+  - release
   - releases/*
 
 jobs:
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -3,10 +3,12 @@ Changelog
 
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
+- **0.7.4 - 4-28-2020**
+
   - Corrected a resource leak in arrays that affects array initialization, and variable
     argument methods.  
 
-  - Upgraded tracing and JNI checks to prevent future resource leaks.
+  - Upgraded diagnostic tracing and JNI checks to prevent future resource leaks.
 
 - **0.7.3 - 4-17-2020**
 
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -3,6 +3,11 @@ Changelog
 
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
+  - Corrected a resource leak in arrays that affects array initialization, and variable
+    argument methods.  
+
+  - Upgraded tracing and JNI checks to prevent future resource leaks.
+
 - **0.7.3 - 4-17-2020**
 
   - **Replaced type management system**, memory management for internal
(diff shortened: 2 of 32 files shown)


marscher:
Lets release it ASAP. Thanks for quickly fixing this.

On April 28, 2020 4:38:57 AM UTC, Karl Nelson <notifications@github.com> wrote:
>While tracing through another bug report I found to my horror that
>there was a problem with local referencing which causes arrays to leak
>references both when used directly and when used as part of argument
>handling which is a major problem.   I recommend that we do a 0.7.4
>release to deal with this significant issue. 
>
>The fix for the issue was to remove two incorrect keep statements.  
>Finding it was much more challenging.  The resource in this case was
>being lost during the middle of a call to variable arguments.  The
>result was the Java frame was now the root frame and pushing all the
>local references that were after the bad call to global scope.   I am
>not sure why Java doesn&#39;t flag this type of problem directly. 
>There is no case in which creating local references without a frame is
>acceptable so ``-Xcheck:jni`` should have caught it.  To identify this
>hairy issue I had to upgrade the trace utilities to report python, java
>and trace events and allow them to be selectively controlled.  Once it
>was clear the source, I developed an internal consistency check which
>is evaluated whenever instrumentation or tracing is enabled.  Creating
>a local reference outside of Java frame will now trigger an exception. 
>The only case where this does not happen is on the exception paths
>where throwing an exception is likely to be fatal and thus lead to a
>lack of diagnostics.   This modification ensures that the current test
>bench will fail if this issue ever reappears.
>
>I backported this from the head so it excludes all of the other changes
>on master.
>
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/jpype-project/jpype/pull/715
>
>-- Commit Summary --
>
>  * Work on tracer support for debugging
>  * Work on JNI tracing.  Add indent optimization.
>  * Add check to catch broken referencing code
>  * Fix memory issue.
>
>-- File Changes --
>
>    M native/common/include/jp_primitive_accessor.h (3)
>    M native/common/include/jp_tracer.h (47)
>    M native/common/jp_arrayclass.cpp (4)
>    M native/common/jp_classloader.cpp (2)
>    M native/common/jp_javaframe.cpp (425)
>    M native/common/jp_method.cpp (2)
>    M native/common/jp_tracer.cpp (113)
>    M native/python/pyjp_array.cpp (5)
>    M native/python/pyjp_module.cpp (24)
>    M native/python/pyjp_value.cpp (1)
>
>-- Patch Links --
>
>https://github.com/jpype-project/jpype/pull/715.patch
>https://github.com/jpype-project/jpype/pull/715.diff
>
>-- 
>https://github.com/jpype-project/jpype/pull/715

-- 

marscher:
You could do the release preparations (tags, changelog release date etc.) in this PR, right?

Thrameos:
Yeah.  I will set it up for release.

PR #716, opened 2020-04-28
Handling for Control-C

### LINKED ISSUE #556: Ctrl+C Segfault on Python/IPython
As indicated in this video https://youtu.be/6_9fqXGT5z8
```
from jpype import *
test_varaible = 2
for i in range(10000):
    java.lang.System.out.printf("*")
```

After start running the code and that I pressed Ctrl+C, segfault happens and the variable `test_variable` could no longer be accessed.

Thrameos:
This PR fixes a bug when Java handles Control-C.   The standard handler for Java is to call shutdown.   As this was not solicited by the user, it effectively yanks the rug out from under Python in the middle of a call.   If it does survive this then the JVM will be shutdown and useless.  Though the most likely outcome is for everything to crash.

This PR reroutes the INTR signal to a method which sends interrupt to the main thread.  This will break out the code assuming Java code is doing something with wait, read, or sleep.  If not then it will do nothing until control is returned to Java.  When it does then we will issue a ``InterruptedException`` and clear the interrupt.

Java really doesn't want this interface to get used.  The warnings from the compiler are obnoxiously bad and they can't be suppressed.  So we will use reflection to call it and skip the warnings.  

Fixes #556


native/java/org/jpype/JPypeSignal.java
```diff
@@ -0,0 +1,47 @@
+package org.jpype;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+
+/**
+ * Java wants to make this action nearly impossible.
```
marscher, line 9: Did you start reading OpenJDK source code? :+1: 
Or hell loads of reverse engineering?

native/java/org/jpype/JPypeSignal.java
```diff
@@ -0,0 +1,47 @@
...
+{
+
+  static Thread main;
+
+  static void installHandlers()
+  {
+    try
+    {
+      Class Signal = Class.forName("sun.misc.Signal");
+      Class SignalHandler = Class.forName("sun.misc.SignalHandler");
```
marscher, line 24: Is this signal handler also abundant in other JVMs?

marscher: Why isnt the usual shutdown hook mechanism sufficient for SIGINT?

native/java/org/jpype/JPypeSignal.java
Thrameos: It was necessary to read OpenJDK source code to pull this one off.   I started reading to try to solve local reference counting.  But it was do compelling that I had to stick around for the section on signals.  Absolutely riveting, I just can't wait to find out what will happen to little oops next.

native/java/org/jpype/JPypeSignal.java
Thrameos: That is a tough one to answer.  I haven't got much vision into others.  But I suspect that ``sun.misc.Signal`` and ``sun.misc.Unsafe`` are both prevalent in many versions.   I guess I should misname this misc2 and make sure we will still run even if the hooks are missing.

Thrameos: The problem here is that people don't want to shutdown here.   If someone is running a procedure under interactive Python and they hit ^C they are expecting a break out of the loop or whatever is running and a return to the interactive console.  But Java is catching the ^C because it has taken over all the signal handlers.   So it takes the ^C as a signal to kill everything (which would be fine if it was in charge), and starts a shutdown sequence.  Python doesn't know anything about it and all of sudden the stack frames in Java start collapsing.  Thus a crash is likely to happen.   The correct action is for Java to take the ^C as a hint to and let Python decide if it wants to shutdown (after all it is in charge here).   If Python wants to shutdown then shutdownJVM happens and everything is normal.   

I suppose we can add a keyword argument to startJVM ``interactive=False`` which will allow ^C to proceed with the default behavior.   But 90% of the time Java shutting down without telling Python about it first is likely to just go straight to a crash.

Thrameos:
I found some relevant information on this topic.

https://stackoverflow.com/questions/1050370/java-error-java-lang-illegalargumentexception-signal-already-used-by-vm-int

IBM does have this though some switches may be needed to install the hook.   So we have a choice. 

(more than one may apply)

- [ ] Add an option not to install the signal handler to startJVM() so we can skip this if needed.
- [ ] Raise a RuntimeException if we fail to install the hook.  This should be reflected in a proper startJVM failed message.
- [ ] Install the hook and silently fail if we don't install the hook properly.  User will get the crash on ^C on jvms that lack the hook and shutdown on ^C.

I vote for 1 and 2 as the best compromise.  Install the hook if we can (which is the right thing most of the time), turn off if the user has other plans for the signal handlers.  The last one is just a recipe for bug reports.

native/java/org/jpype/JPypeSignal.java
marscher: Thats one of the greatest benefits of open implementations. I hope you will not discover more of these nasty oops any soon...

native/java/org/jpype/JPypeSignal.java
marscher: Would it also be possible to unregister all signal handlers, which the JVM registered during startup, and let only Python handle signals? This way we would not need to take care about different JVM implementations and have full control of the process.
Probably I'm currently unaware of all possible consequences, but it might be worth a consideration. What do you think?

Thrameos: I am not sure that really works.   The JVM is pretty darn insistent about its signal handlers.  Several of them a vital for operation.  We could repoint just this one back to C++ by capturing and restoring this one handler. 

But to do that we would have to understand how Python and Java are directing signals (when mixed with threads).   My understanding is that Python is redirecting all signals to the main thread and then calling the signal routines.   Java on the other hand spins a thread up for each single.

It is also not clear how the signals were layered in the first place.  When we hit control C in Python it is clear that both Python and Java got the signal.  I assume to came to Java who chained it to Python, but I don't have technical details.

Do you want me to try to manually control the signals for this PR?   If so I may have to deal with OS specific behaviors as windows and unix have different concepts of signal handling.

marscher: Thanks for pointing out, that my idea is infeasible to implement. I was just thinking about per process signals, and forgot about the multithreaded path.

PR #717, opened 2020-04-28
Fix instability in coverage

Thrameos:
We are getting a lot of failed coverage test because Java is getting different results each time that coverage is run.   The problem is there is a race condition with the coverage tool during shutdown.  This PR may fix that issue by forcing a dump after we have completed all of our tasks.


Thrameos:
I suppose I should compile the hard coded call to Jacoco in to the test harness and then just place an install hook.  That would be cleaner.   I am going to place a hold on this one for now.

Thrameos:
Merging this one so that coverage on PRs is more stable.

PR #718, opened 2020-04-28
Release 0.7.4

Thrameos:
This is a PR to pull the quick fix back to master.


Thrameos:
@marscher I put out a draft copy of the release with the wheels, but I don't have time this morning to verify the individual wheels.  I still lack a script that can create a virtual env, run the tests, and then give me a thumbs up or down on each.   If you have time to spot check one then you can hit the release button on publish the draft.   Or I will do it late tonight when I have time to spot check.

docker/build_windows.sh
```diff
@@ -0,0 +1,4 @@
+~/env/python-3.5-conda/Scripts/python.exe -m pip wheel JPype1-0.7.4.tar.gz -w wheelhouse
```
marscher, line 1: Do we need to change this script for every new release? Eventually we can make it version agnostic

Thrameos: I am hoping to get all the release scripts moved to an azure pipeline release pipeline.   That would also allow me to build the missing OSX builds.   For now I have just checked in my build and test wheel scripts so that when covid finally arrives there is some record of what I did (and given that i have burned lungs from an industrial accident I am not looking forward to it).

marscher:
That would be a good improvement. It will shorten release cycles, if there is only minimal effort to trigger a shipping.

I really wish, that you and your family will be fine. Unfortunately the people in charge of protecting everyone are not competent.

marscher:
A tool for that purpose is tox. It has a minimal configuration file to create a venv and run commands in it.

On April 28, 2020 3:36:12 PM UTC, Karl Nelson <notifications@github.com> wrote:
>@marscher I put out a draft copy of the release with the wheels, but I
>don't have time this morning to verify the individual wheels.  I still
>lack a script that can create a virtual env, run the tests, and then
>give me a thumbs up or down on each.   If you have time to spot check
>one then you can hit the release button on publish the draft.   Or I
>will do it late tonight when I have time to spot check.
>
>-- 
>https://github.com/jpype-project/jpype/pull/718#issuecomment-620682492

-- 

Thrameos:
Yeah I considered tox (and nox).  It wasn't clear to me how to make tox chose the different versions of Python I have installed.  I  think I want to try for the long term solution using azure first and failing that going to tox.

PR #722, opened 2020-04-30
Fix for conversion loading.

### LINKED ISSUE #721: automatic conversion of python lists in parameters does not work until java.util.Collection(?) is imported
Assume the following Java class:
```java
package jptest;
import java.util.Collection;

public class Test {
    public static void testMethod(Collection<String> input) {
        System.out.println(input);
    }
}
```
invoked by the following python code:
```python
import jpype

jpype.startJVM(
    jpype.getDefaultJVMPath(),
    classpath='/path/to/built/class,
    convertStrings=True
)

import jpype.imports
jpype.imports.registerDomain("jptest")

# from java.util import Collection

from jptest import Test

Test.testMethod(["a", "b"])
```

Expected result:
```
[a, b]
```

Actual result:
```
  File "D:/jpype-playground/test.py", line 16, in <module>
    Test.testMethod(["a", "b"])
TypeError: No matching overloads found for jptest.Test.testMethod(list), options are:
	public static void jptest.Test.testMethod(java.util.Collection)
```

Once `java.util.Collection`, `java.util.ArrayList` (or probably any other collection?) is imported, e.g. in the commented-out line, conversion works as expected.

JPype 0.7.3, Java 1.8 (x64) on Windows 10.

Thrameos:
This refreshes the Python type cache before matching arguments.  Without this user defined conversions may be missed.

Fixes #721

### DIFF
--- a/native/common/jp_classhints.cpp
+++ b/native/common/jp_classhints.cpp
@@ -244,11 +244,13 @@ public:
 	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls)
 	{
 		PyJPClassHints *pyhints = (PyJPClassHints*) cls->getHints();
+		// GCOVR_EXCL_START
 		if (pyhints == NULL)
 			PyJPClass_create(*match.frame, cls);
 		pyhints = (PyJPClassHints*) cls->getHints();
-		if (pyhints == NULL) // GCOVR_EXCL_LINE
-			return match.type = JPMatch::_none;  // GCOVR_EXCL_LINE
+		if (pyhints == NULL)
+			return match.type = JPMatch::_none;
+		// GCOVR_EXCL_STOP
 		JPClassHints *hints = pyhints->m_Hints;
 		hints->getConversion(match, cls);
 		return match.type;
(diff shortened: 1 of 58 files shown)


native/common/jp_classhints.cpp
```diff
@@ -244,8 +244,13 @@ class JPHintsConversion : public JPConversion
 	virtual JPMatch::Type matches(JPMatch &match, JPClass *cls)
 	{
 		PyJPClassHints *pyhints = (PyJPClassHints*) cls->getHints();
+		// GCOVR_EXCL_START
+		if (pyhints == NULL)
+			PyJPClass_create(*match.frame, cls);
+		pyhints = (PyJPClassHints*) cls->getHints();
```
marscher, line 250: this should go into the pyhints == NULL block, right?

test/harness/jpype/collection/CollectionTest.java
```diff
@@ -0,0 +1,17 @@
+package jpype.collection;
```
marscher, line 1: is this file used elsewhere?

Thrameos: This was a left over from the user report.  I think I should make a test that passes a Python list and make sure I get an array list out the other side.  Same with map

native/common/jp_classhints.cpp
Thrameos: Yeah it is not necessary to look it up twice.  Will fix.

PR #723, opened 2020-05-01
Stubs

Thrameos:
This adds support for the proposed stubbing API.  The stub info is available in ``_hints``.  Switched from attribute to typing.Protocol for some conversions (though may have a performance hit for some cases). 

Fixes #741


michi42:
Thanks a lot already!

One suggestion for backwards compatibility, at least for Python >= 3.6 you can use typing_extensions (https://github.com/python/typing/tree/master/typing_extensions) to provide the PEP 560 / Python 3.8 Protocol functionality. This would spare handling 3.6 and 3.7 in a special way.
For 3.5 I'm afraid there is no nice solution.

jpype/protocol.py
```diff
@@ -0,0 +1,92 @@
...
+def _JFileConvert(jcls, obj):
+    return jcls(obj.__fspath__())
+
+
+@_jcustomizer.JConversion("java.util.Collection", instanceof=Sequence)
+def _JSequenceConvert(jcls, obj):
+    return _jclass.JClass('java.util.Arrays').asList(obj)
+
+
+@_jcustomizer.JConversion("java.util.Collection", none=str)
```
michi42, line 66: This decorator looks out of place here ...

jpype/_jcustomizer.py
```diff
@@ -17,7 +17,7 @@
 __all__ = ['JImplementationFor', 'JConversion']
 
 
-def JConversion(cls, exact=None, instanceof=None, attribute=None):
+def JConversion(cls, exact=None, instanceof=None, attribute=None, none=None):
```
michi42, line 20: Cosmetics - I think I would call the attribute to deny certain types "exclude" instead of "none", as in python "none" is potentially confusing with the "None" keyword and type.

jpype/protocol.py
Thrameos: Moved back  to resolve.

jpype/_jcustomizer.py
Thrameos: Does the same thing apply to the user guide 

https://jpype.readthedocs.io/en/latest/userguide.html#type-matching
https://jpype.readthedocs.io/en/latest/userguide.html#java-conversions

Thrameos:
I think that adding a compatibility library for 3.6 and 3.7 and still having to patch around it for 3.5 is a lot of effort for very little gain.  The kludge that I used to backport works for all versions with the minor  difficulty that you need to use jpype.protocol to get the aliases that work for all versions.  Do you see any issues with that solution?

Thrameos:
@marscher this likely should go in 0.8 so that we can have the stub support that @michi42 is working on.  

Basically I dupped all the findJavaConversion methods into getInfo methods.  This forced the long planned cleanup of the JPConversions to push all of the conversion logic out of findJavaConversion.   The end result is a bit cleaner.   There still may be bugs in the tables as it was a lot of bookkeeping to make sure each conversion reports its values accurately.   I attached all of the conversion information on to the existing hints structure so that it will be cached.   I pushed all of the conversion logic into ``jpype.protocol`` which holds types and conversions for the stub generator.   As it will need to be an exposed interface I gave it a public name though it currently has no exposed symbols. 

The big challenge on this one is getting it to work across versions.  We currently span 3.5 to 3.8 but protocols were not developed until 3.6 and not fully realized until 3.8.  So I used some logic to create fake protocol types which can be used across all versions.  Unfortunately, even in 3.8 Protocol does not mean isinstance checks properly.  ``isinstance("your dog", typing.Protocol) == True`` despite the fact that "your dog" is clearly not something that you can use in ``isinstance(obj, "your dog")``.     So instead I used ``__instancecheck__`` to see if it quacks.

After this I will hit one of the following.
- [ ] Finish up API to make Java doc available in JPype
- [ ] Finish up review of Collections and String API to see how many Python contracts I can satisfy.
- [ ] Consider it there is any good way to make JChar derive from str so that we complete the last broken return contract.
- [ ] Try to solicit users to give an opinion on the cast operator so that we can get that in (or reject it) so the documentation can be final.

jpype/protocol.py
```diff
@@ -0,0 +1,92 @@
...
+    from collections.abc import Sequence, Mapping
+
+    class _AttributeMeta(type):
+        def __instancecheck__(self, obj):
+            return hasattr(obj, self._attrib)
+    SupportsPath = _AttributeMeta("SupportsPath", (object,), {
+                                  "_attrib": "__fspath__"})
+    SupportsIndex = _AttributeMeta("SupportsIndex", (object,), {
+                                   "_attrib": "__index__"})
+    SupportsFloat = _AttributeMeta("SupportsFlaot", (object,), {
```
michi42, line 32: Typo in first constructor argument (Fl**ao**t). Also I think SupportsFloat was there even before protocols, as of Python 3.5 (unlike SupportsIndex, that was introduced in 3.8).

michi42:
> I think that adding a compatibility library for 3.6 and 3.7 and still having to patch around it for 3.5 is a lot of effort for very little gain. The kludge that I used to backport works for all versions with the minor difficulty that you need to use jpype.protocol to get the aliases that work for all versions. Do you see any issues with that solution?
I will check - my concern is that static type checkers like MyPy or pytype and IDEs might well know about typing_extensions but they won't run custom code to carry out type checks. I will have  to do a few tests with various python versions to confirm.

Thrameos:
Okay is that any better?

jpype/protocol.py
```diff
@@ -17,35 +17,29 @@
...
+    class SupportsFloat(Protocol):
+        def __float__(self) -> str: ...
 
-    # Types we need
     @runtime_checkable
-    class SupportsPath(Protocol):
-        """An ABC with one abstract method __fspath__."""
-        __slots__ = ()
+    class SupportsIndex(Protocol):
+        def __index__(self) -> str: ...
```
michi42, line 30: Return type should be `int`

jpype/protocol.py
```diff
@@ -17,35 +17,29 @@
...
-    SupportsIndex = _AttributeMeta("SupportsIndex", (object,), {
-                                   "_attrib": "__index__"})
-    SupportsFloat = _AttributeMeta("SupportsFlaot", (object,), {
-                                   "_attrib": "__float__"})
-else:
-    from typing import Protocol, runtime_checkable, abstractmethod
-    from typing import SupportsIndex, SupportsFloat
-    from typing import Sequence, Mapping
+    @runtime_checkable
+    class SupportsFloat(Protocol):
```
michi42, line 25: I think this is not needed, Python >= 3.5 should have it in the standard library `typing` module

michi42:
Looks good to me (with the 2 little comments) - thanks a lot.

Thrameos:
Yeah, we could call ``_canConvertToJava`` on all of the types that have a default constructor.  That would verify that each conversion actually works.

PR #725, opened 2020-05-05
Mutable list

Thrameos:
This completes the MutableSequence contract on ``java.util.List``. 

In addition, it fixes the following bugs:
- sticky methods were being applied twice causing endless recursion in some cases.
- sticky methods removed ``__jclass_init__`` on a base class.
- delitem on negative indices failed.

### DIFF
--- a/jpype/_jcollection.py
+++ b/jpype/_jcollection.py
@@ -18,7 +18,7 @@ import _jpype
 from . import _jclass
 from . import types as _jtypes
 from . import _jcustomizer
-from collections.abc import Mapping, Sequence
+from collections.abc import Mapping, Sequence, MutableSequence
 
 JOverride = _jclass.JOverride
 
@@ -83,6 +83,7 @@ class _JList(object):
 
     def __jclass_init__(self):
         Sequence.register(self)
+        MutableSequence.register(self)
 
     def __getitem__(self, ndx):
         if isinstance(ndx, slice):
@@ -90,7 +91,7 @@ class _JList(object):
             return self.subList(ndx.start, ndx.stop)
         else:
             if ndx < 0:
-                ndx = self.size() + ndx
+                ndx += self.size()
             return self.get(ndx)
 
     def __setitem__(self, ndx, v):
@@ -100,7 +101,7 @@ class _JList(object):
             self.addAll(ndx.start, v)
         else:
             if ndx < 0:
-                ndx = self.size() + ndx
+                ndx += self.size()
             self.set(ndx, v)
 
     def __delitem__(self, ndx):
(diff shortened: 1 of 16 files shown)


jpype/_jcollection.py
```diff
@@ -100,20 +101,22 @@ def __setitem__(self, ndx, v):
...
+                ndx += self.size()
             self.set(ndx, v)
 
     def __delitem__(self, ndx):
         if isinstance(ndx, slice):
             ndx = _sliceAdjust(ndx, self.size())
             self[ndx.start:ndx.stop].clear()
         elif hasattr(ndx, '__index__'):
-            return self.remove(_jtypes.JInt(ndx))
+            if ndx<0:
```
marscher, line 112: https://www.google.com/search?q=autopep8+git+precommit+hook :+1:

jpype/_jcollection.py
```diff
@@ -134,6 +137,45 @@ def count(self, obj):
...
+    def insert(self, idx, obj):
+        if idx<0:
+            idx += self.size()
+        return self.add(idx, obj)
+
+    def append(self, obj):
+        return self.add(obj)
+
+    def reverse(self):
+        _jpype.JClass("java.util.Collections").reverse(self)
```
marscher, line 149: This impl is an inplace operation, right? I thought reverse() just returns an reversed iterator, but leaves the original sequence untouched.

jpype/_jcustomizer.py
```diff
@@ -121,10 +124,11 @@ def _applyStickyMethods(cls, sticky):
...
         attr = getattr(method, '__joverride__')
         rename = attr.get('rename', None)
+        name = method.__name__
         if rename:
-            type.__setattr__(
-                cls, rename, type.__getattribute__(cls, method.__name__))
-        type.__setattr__(cls, method.__name__, method)
+            orig = type.__getattribute__(cls, name)
+            type.__setattr__(cls, rename, orig)
+        type.__setattr__(cls, name, method)
```
marscher, line 131: Is it intended to have the original name and the rename pointing to the same method? It would be an alias then, right?

jpype/_jcollection.py
marscher: It is inplace. Sorry for the noise

jpype/_jcustomizer.py
Thrameos: Not sure what you mean here...

```
 orig = type.__getattribute__(cls, name)    # orig <= cls[name]
 type.__setattr__(cls, rename, orig)            # cls[rename] <= orig
 type.__setattr__(cls, name, method)         # cls[name]<=method
```

So the original is renamed and then the method from the prototype put in place.

PR #733, opened 2020-05-09
Updates to documentation and Threading API

### LINKED ISSUE #732: Deal with missing AttachThreadAsDaemon
The call the attaching threads is currently missing from the API.  The isAttached and attachToJVM are both not functioning as excepted either.  This needs to be cleaned up for the 0.8 release.

Thrameos:
This PR moves the threading entry points to ``java.lang.Thread`` and restores the missing add thread as daemon method.   Updates the documentation and adds a fake java module under doc so that we produce much better documentation.

This also fixes two contracts in Map so that it won't unnecessarily produce TypeErrors.

Fixes #732


jpype/_jthread.py
```diff
@@ -0,0 +1,68 @@
...
+        that it can be disconnected prior to thread termination to prevent
+        leaks.
+
+        Returns:
+          True if the thread is attached to the JVM, False if the thread is
+          not attached or the JVM is not running.
+        """
+        return _jpype.isThreadAttachedToJVM()
+
+    @classmethod
```
marscher, line 31: This should be staticmethod, if you do not need the class to be passed. But this is only a minor detail.

Thrameos: I am not sure if the customizer takes staticmethod customizations properly.   I will have to verify.   I remember that when I was doing the ``__javainit__`` methods I ran into a wierd corner case that transferring certain annotated methods caused odd behaviors, but I don't recall which it was.

Thrameos:
At some point we need to deprecate the original attachThreadToJVM.   Long ago it was necessary to attach threads to prevent segfaults so many codes still do so, but because our shutdown logic was missing that hid a bug that most of the time the user does not want to attach a thread in such a way that the JVM can't be shutdown which is what attach implies.  

So this leaves us with an issue.   Every call to attachThreadToJVM needs to be examined to avoid hanging on shutdown and 98% of the time calling attachThreadToJVM is unnecessary.  In the 2% of times where it was necessary it should have been attachThreadAsDaemon.  Then if the user really wants a shutdown to be controlled by the thread than use ``Thread.attach()``.

The leaves the old call in a very bad spot.  It does exactly what it says it does, but that behavior is likely to cause bugs and other difficult to understand behaviors.   This favors given it a hard deprecation in which it warns "remove me" and is made into a stub that does nothing.   Then point the user to documentation which lays out how to fix it.

Is that reasonable?

PR #734, opened 2020-05-09
Update userguide for swing

### LINKED ISSUE #730: Call a class that extends javax.swing
My question about JPackage and JFrame using jpype.
Is it able to support swing or awt on current jpype version ?
For example,
```
<test.java>
package com.gui;

import java.awt.*;
import javax.swing.*;

public class foo extends javax.swing.JFrame {
   ....
}
```

```
<test.py>
from jpype import *

def test
   startJVM()
   pkg = JPackage('com').gui
   testclass = pkg.foo
```

My testing above some code results nothing happen and without any output log.
So, I think that current jpype version is not support the javax.swing.JFrame package.
Any some opinion about jpype using javax.swing or JFrame ?

Thanks.

Thrameos:
Update the documentation regarding swing.

Fixes #730

### DIFF
--- a/doc/userguide.rst
+++ b/doc/userguide.rst
@@ -2430,7 +2430,7 @@ it is removed, it is free to switch identities every time it is garbage
 collected.
 
 AWT/Swing
-*****
+*********
 
 Java GUI elements can be used from Python.  To use Swing
 elements the event loop in java must be started from a user thread. 
--- a/doc/userguide.rst
+++ b/doc/userguide.rst
@@ -2433,7 +2433,7 @@ AWT/Swing
 *********
 
 Java GUI elements can be used from Python.  To use Swing
-elements the event loop in java must be started from a user thread. 
+elements the event loop in Java must be started from a user thread. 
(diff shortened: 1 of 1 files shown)


doc/userguide.rst
```diff
@@ -2429,6 +2429,43 @@ it remains in the Java map, it will maintain the same identify.  But once
 it is removed, it is free to switch identities every time it is garbage
 collected.
 
+Swing
```
jeesang7, line 2432: How about that ? 

```suggestion
AWT/Swing
```

doc/userguide.rst
```diff
@@ -2429,6 +2429,43 @@ it remains in the Java map, it will maintain the same identify.  But once
 it is removed, it is free to switch identities every time it is garbage
 collected.
 
+AWT/Swing
+*********
+
+Java GUI elements can be used from Python.  To use Swing
+elements the event loop in java must be started from a user thread.
```
marscher, line 2436: s/java/Java/g

PR #739, opened 2020-05-10
Ibm

Thrameos:
Debugging help for AIX user.


marscher:
Do we really need this in master? I see new tracing sections in the JVM startup and some debugging files. I guess it would be sufficient, if these live on some debug branch. That way the aix user can compile the code from there (via pip install git+https://github.com...) and we do not drag code into master we would only need for this specific case. If you think however that this will be useful in future debugging efforts, please merge accordingly.

Thrameos:
The main reason for putting this in is so that project/debug/unix is available.  One of the most challenging piece is when someone has a site specific problem with the JVM.   Going back I see we have about 6 or so of these types of requests (counting threads and "me too" attached to them.)  The problem being that they are not really JPype issues but rather problems on the user's side.   

Though I don't think we need to ship this sort of documentation and test scripts, having them around in master so that we can point them to it helps.   The rest of the PR is just minor improvements to the logging when a problem like that occurs.  I will check it over and remove anything that looks out of place.

Thrameos:
Reviewing this I still think it is a good thing.  The failure the user was getting is unfortunately on a path we can't test from our test suite and therefore there were bugs in which a portion of the diagnositic information was truncated so at to make it hard to debug.  The outcome was going to be the same either way as JPype was going to bail, but bailing with a reason is better then just aborting.  It took serious work to actually test them so that I could see what the user was getting.  (Making copies of the whole JVM with a byte editor to cause specific failures.)  I certainly don't want to repeat that if adding some otherwise useless diagnostics to the master will help the next time.

The other half is the site diagnostic routines.  We could leave them off on a debugging branch, but it would end up hopelessly out of date which likely to just confuse the issue.  90% of the time these problems happen to new users who are developers trying the module for the first time.   (who also are like me and have machines with many copies of the JVM some of which are not in working states).   And most of the time the user decides that JPype is too much work and just moves on to another solution.  Having a defined way to resolve those issues before we make a bad impression is needed.

I wish there was a way to actually get diagnostics from within the JPype module, but the shared library call that we use is really minimalist and most of the time says "success" or "fail" rather than something useful like "tried to load libjvm.so but library libc.42.so is 32bit and libjvm.so is 64bit".   Thus we are trying to solve problems which are really outside of our scope.

PR #742, opened 2020-05-11
Release

### LINKED ISSUE #731: Project with JPype master hangs, with v0.7.4 and v0.7.3 it does not - problematic commit is 9b0cfe01
[edited to reflect the insight that v0.7.4 works but master does not and that conda 0.7.4 does not work and that conda 0.7.4 is maybe not equal to tag v0.7.4]

In my project https://github.com/hexhex/hexlite I use jpype and since changing to 0.7.4 in conda I nearly always get hangs in my automatic tests. reverting to 0.7.3 solved the issue. Tag v0.7.4 does work, but master branch does not work.

Symptoms: when my program hangs and even CTRL+C cannot interrupt it. Killing is possible with CTRL+Z  followed by `jobs -p |xargs kill -9`

I have bisected the problem in the jpype repository and the problematic commit is 9b0cfe01

Is there any way I can help to find out why it hangs?

How you can reproduce it:
Cloning the master of https://github.com/hexhex/hexlite and do in a new conda environment the following:

install jpype with 
```
$ python setup.py sdist
$ pip install dist/*
```

install prerequisites of hexlite
```
$ conda install -c potassco clingo
```

build the java part of hexlite
```
mvn clean compile package install
```

install hexlite in develop mode and run the test script that demonstrates some of the hanging tests
```
$ python setup.py develop
$ ./tests.sh
```
the last step might succeed in a few seconds
or it hangs

Thrameos:
Revised release system.  Ported some work from master to avoid extra work.  (docs, mock, setupext)

- [x] Start from the release branch
      ``git checkout release``
- [x] Make a new branch for the release cycle
      ``git checkout -b releases/{version}``
- [x] Merge the current master with the release
      ``git pull origin master``
- [x] Edit doc/CHANGELOG.rst
- [x] Start a release
      ``bumpversion release``
- [x] Send the release to be evaluated
      ``git push``
- [x] Verify CI on azure
- [x] Manually trigger a ``jpype.release`` on azure
      If successful, download the artifacts for publication.
- [x] Advance the release pointer
      ``git checkout release``
      ``git pull releases/<version>``
- [x] Publish the release
  - Add draft release on github
  - Attach the artifacts to the release.
- [ ] Start master on a new cycle
  - Use a PR to pull release back to master
  - ``git checkout master``
  - ``git checkout -b cycle``
  - ``git pull master``
  - ``bumpversion patch``
  - Use PR to insert the cycle in master

Fixes #731


.azure/scripts/osx-python.sh
```diff
@@ -0,0 +1,34 @@
...
+
+sudo installer -pkg $INSTALLER_NAME -target /
+
+sudo rm /usr/local/bin/python
+sudo ln -s /usr/local/bin/python$PYTHON_VERSION /usr/local/bin/python
+
+which python
+python --version
+python -m ensurepip
+python -m pip install setuptools wheel
```
marscher, line 34: have you considered using pyenv for setting the actual python version? https://github.com/pyenv/pyenv It should make our lives easier switching between interpreters, Symlinking seems a bit like brute force (and what would happen if /usr/bin/ etc. takes a higher priority in PATH?)

.bumpversion.cfg
```diff
@@ -1,16 +1,30 @@
 [bumpversion]
-current_version = 0.7.2
+current_version = 0.7.6_dev0
```
marscher, line 2: I like that you reincarnated bumpversion, I also forgot about it in the meantime...

jpype/__init__.py
```diff
@@ -47,8 +47,7 @@
 __all__.extend(_jcustomizer.__all__)
 __all__.extend(_gui.__all__)
 
-__version_info__ = (0, 7, 4)
-__version__ = ".".join(str(i) for i in __version_info__)
+__version__ = "0.7.6_dev0"
```
marscher, line 50: I would not delete __version_info__ without deprecation. We can easily set it like this:

__version_info__ = tuple(__version__.split('.'))

setupext/platform.py
```diff
@@ -49,22 +49,15 @@
     platform_specific['extra_link_args'] = ['/DEBUG']
```
marscher, line 49: So Windows will have debug symbols, while OSX and Linux don't?

.azure/scripts/documentation.yml
```diff
@@ -0,0 +1,24 @@
...
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '3.8'
+
+- script: |
+    pip install  --exists-action=w --no-cache-dir -r test-requirements.txt -r doc/doc-requirements.txt
+  displayName: 'Install requirements'
+
+- script: |
+    python -m sphinx -T -b readthedocs -W -d _build/doctrees-readthedocs -D language=en doc build/html
```
marscher, line 12: I have added -W in the past to turn warnings into errors, as most of time these are mistakes in the markup. Now I propose to add the flag --keep-going as it seems to collect all warnings first and then exit with status -1.

from the sphinx documentation:
```

-W

    Turn warnings into errors. This means that the build stops at the first warning and sphinx-build exits with exit status 1.

--keep-going

    With -W option, keep going processing when getting warnings to the end of build, and sphinx-build exits with exit status 1.

    New in version 1.8.
```

.azure/scripts/coverage.yml
```diff
@@ -0,0 +1,48 @@
...
+- template: jdk.yml
+  parameters:
+    version: 11
+
+- script: mvn -q -f project/coverage package
+- script: |
+    python setup.py test_java
+    pip install gcovr pytest-cov jedi
+    pip install -r test-requirements.txt
+    pip install numpy typing_extensions
```
marscher, line 16: I think it would be faster to pass all requirements in one pip invocation, but this is just a minor detail.

.azure/scripts/build-wheels.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Bundle external shared libraries into the wheels
+for whl in wheelhouse/$package_name-*.whl; do
+    auditwheel repair --plat $PLAT "$whl" -w /io/wheelhouse/
+done
+
+# Install packages and test
+for PYBIN in "${pys[@]}"; do
+    "${PYBIN}/python" -m pip install $package_name --no-index -f /io/wheelhouse
+    # Manylinux does not have a JVM so there is no way to test the wheel in the docker
```
marscher, line 26: Have we had the discussion already, if we should derive from Manylinux for this very purpose?

marscher:
What happend to v0.7.5? Why would we omit that minor version?
I thought I borked the source dist of 0.7.3/4

Thrameos:
v0.7.5 is waiting on the branch for your approval.  This PR is the post release which the final state after release.  The master will be 0.7.6pre1  so that we no longer get confusion between the master and the release branches.

setupext/platform.py
Thrameos: This is an odd one.  For some reason (I can't explain) the linux binary is bloated to 10 times all the other ones.  I didn't want to have to strip because stripping will get rid of way to much.   I think that -g3 was an accident on my part many versions back (I think I did the edit because I needed the symbols).  You can see from setupext that we deliberately override the -g0 to -g3 for tracing and coverage builds.   

So what is the correct compiler flags for a release?

jpype/__init__.py
Thrameos: Okay I will update the rerelease with this change.  The hard part is the fact that tuples elements are numbers.   Not sure what to do about it  But I am fine with ("0","7","6_dev0").

.azure/scripts/osx-python.sh
Thrameos: The recipe for this was from the official azure blog on wheel releases.  I just copied.  There may be better methods.   The blog was also pretty old so I had to up versions to get it to work.

.azure/scripts/coverage.yml
Thrameos: Yeah I was debating this myself.  I was going to put it in doc-requirements.txt, but when I read the logs on lgtm I could see that it also is adding a bunch of stuff then pulling doc-requirements.  If I put this in doc-requirements I may foul up some of the stuff they preinstall.  So we end up needing rtd-requirements.txt (used only on azure) with the stuff they put in and doc-requirements for all the stuff we are adding.   Is that a workable solution?

.azure/scripts/build-wheels.sh
Thrameos: I think you may have mentioned it.  Though I am not sure I can pull it off.  My reading of manylinux is that it is some ancient version of redhat.   It would be hard to get a JVM on that version.  A better approach would be to add a second docker task or use azure itself to test the install.  Unfortunately I ran out of gas after 10 hours of working on the release and just decided to punt.   The full pipeline of linux is pretty slow so to develop this we would need to have test pipeline that pulls the artifacts from the earlier build, works out all the patterns then adds it to the release.

Thrameos:
So lets run down the remaining tasks on the v0.7.5 release

- [ ] Checkout releases/0.7.5 and add the missing version_info using split.  Commit.
- [ ] Manually trigger the release pipeline to create a fresh set of binaries.
- [ ] Download all of the different artifacts from https://dev.azure.com/jpype-project/jpype/_build/results?buildId=378&view=results  (or rather the next run)
- [ ] Go to releases page and draft the release notes from the CHANGELOG
- [ ] On the release page select ask it create a tag on releases/0.7.5  (not master or we will have to repeat the process or move the tag (which is bad because some processes trigger on it)
- [ } Copy all the wheels onto the release (including the sdist which I forgot until 2 days later last time)
- [ ] Merge v0.7.5 into this PR so that we have the correct head for future work.
- [ ] Merge this PR

Okay I think that covers it.   Assuming that you have approved everything, I can finish the cycle late tonight.  (9PM PST 5/12)  I believe that there was only the one requested change.  But poke me between now and then if I missed something.

.azure/scripts/documentation.yml
Thrameos: Okay that is on a CI task, so we can do that in master and I will let the release go without updating as this file is not in the release itself.

jpype/__init__.py
marscher: I would not consider it a fact, but a recommendation. The splitting could be better though.

setupext/platform.py
marscher: I could also be a leftover from ancient times, when I actively fiddled around with the code. If we omit debug flags completely, we are not able to retrieve proper stacktraces from our users. I saw how you instructed a user to get a nice tracing; pinning down the issue. So I guess we can safely omit the symbols, and instruct people to reproduce the found crashes with a debug build.

.azure/scripts/coverage.yml
marscher: When there are conflicts with RTD, we should avoid installing them solely on RTD and just install them on cI. How did this fire up on lgtm? I dont really understand.

.azure/scripts/build-wheels.sh
marscher: Many thanks for this massive effort!
It used to be derived from redhat, now it is debian stretch (oldstable). So in principle we should be able to get a jdk. But I guess you're right, we should test it outside of the manylinux container. We can postpone it for later efforts on the __great__ release pipeline

.azure/scripts/coverage.yml
Thrameos: Sorry I meant to say RTD not LGTM.   In RTD they first pull the repo then install a set of requirments.  Then it pulls doc-requirements .   I don't want to bust the requirements.   So I need to have another text file with the list of the stuff RTD puts in place.

setupext/platform.py
Thrameos: Okay let me make it ``-g`` so we have some symbols.   The size reduction didn't work anyway so there is likely some other bug that I will fix later.

marscher: It seems like ggdb is added by distutils (it derives the CFLAGS from the ones used to compile python itself). This flag overrides g0 and the binary is huge after all.

marscher:
I cannot login to Azure via Github... sorry. I really wanted to perform all the steps necessary. In the meantime I will fix the source release for 0.7.4

Thrameos:
Okay you should be on the access list.  If not just poke me on my email.

marscher:
I don't believe it is an permission problem, since I cannot login via github at all

Thrameos:
@marscher are you going to kick of PyPi distributions soon?

marscher:
Sorry for the delay. The packages are online on PyPI now.

PR #743, opened 2020-05-13
Fix exception bugs

Thrameos:
This PR fixes a number of bugs that were introduced due to interactions between the JPackage upgrade and exception updates.

-  JPackage take over of JImports removed the error handling routines so they needed to be replaced.
- The hidden constructor token was leaking into the args list for exceptions.
- Using certain exceptions as base classes was triggering internal fatal errors.
- The wrong constructor was called in some cases.

To fixed this the exception arguments are now generated internally.  And JPackage has a special method ``_handler`` so that JPackage can redirect exceptions to ImportError as needed.

(This will be checked once the release is merged)


jpype/_jexception.py
```diff
@@ -29,10 +29,6 @@ class JException(_jpype._JException, internal=True):
     ``java.lang.Throwable``.
 
     """
-    def __new__(cls, *args, **kwargs):
-        self = _jpype._JException.__new__(cls, *args)
-        _jpype._JException.__init__(self, *args)
-        return self
```
Thrameos, line 35: The old new method is now complete removed.   As we deprecated the last of the use of JException as factory this is just a required stub for catching exceptions.

jpype/_jexception.py
```diff
@@ -58,18 +54,9 @@ def stacktrace(self):
...
-        return self._jargs()
-#    args = property(lambda self: self._jargs(), None)
-#    """ Test doc string on property"""
-
-    def _jargs(self):
-        cause = self.getCause()
-        if cause is None:
-            return (str(self.getMessage()),)
-        return (str(self.getMessage()), cause,)
+        return self._args
```
Thrameos, line 59: All of the code for handling args is now created when the exception is thrown.

jpype/imports.py
```diff
@@ -76,74 +76,25 @@ def _keywordWrap(name):
...
+        raise ImportError("Unable to import '%s' due to initializer error" % javaname) from ex
+    if exname == "java.lang.UnsupportedClassVersionError":
+        raise ImportError("Unable to import '%s' due to incorrect Java version" % javaname) from ex
+    if exname == "java.lang.NoClassDefFoundError":
+        missing = str(ex).replace('/', '.')
+        raise ImportError("Unable to import '%s' due to missing dependency '%s'" % (
+            javaname, missing)) from ex
+    raise ImportException("Unable to import '%s'" % javaname) from ex
+
+
```
Thrameos, line 97: There was an unexpected interaction between JPackage and exception handling.   When the new JPackage changes were added, it bypassed the checks as the import is now taking place in python's ``importlib/_builtin.py``.  To resolve this I had to add a hook to JPackage so that it could redirect error handling to a new location.   The behavior is the same, but the method is rather different.

native/common/jp_class.cpp
```diff
@@ -321,10 +322,29 @@ JPPyObject JPClass::convertToPythonObject(JPJavaFrame& frame, jvalue value, bool
...
+				tuple0.setItem(0, JPPyString::fromStringUTF8(
+						frame.toStringUTF8(m)).get());
+			} else
+			{
+				tuple0.setItem(0, JPPyString::fromStringUTF8(
+						frame.toString(value.l)).get());
+			}
+			tuple1.setItem(1, tuple0.get());
+		}
 		// Exceptions need new and init
```
Thrameos, line 346: We get a new block of code that works out what the proper arguments to an exception would have been had it be thrown from Python.  Thus Python and Java thrown exceptions are now identical.

native/python/pyjp_object.cpp
```diff
@@ -239,14 +231,52 @@ static PyType_Spec objectSpec = {
 	objectSlots
 };
 
+static PyObject *PyJPException_new(PyTypeObject *type, PyObject *pyargs, PyObject *kwargs)
```
Thrameos, line 234: Special constructors are needed.  There was a flaw in the old design in which the wrong new method was called.  We zero the memory so it wasn't causing an issue today, but I picked it up as an error when I extended as type in the sql work and got really unexpected results.  It should now be resolved.

native/python/pyjp_object.cpp
```diff
@@ -256,17 +286,32 @@ static PyObject* PyJPException_expandStacktrace(PyObject* self)
...
 static PyMethodDef exceptionMethods[] = {
 	{"_expandStacktrace", (PyCFunction) PyJPException_expandStacktrace, METH_NOARGS, ""},
 	{NULL},
 };
 
+static PyGetSetDef exceptionGetSets[] = {
+	{"_args", (getter) PyJPException_args, NULL, ""},
+	{0}
+};
+
```
Thrameos, line 306: We can't implement args directly as it has a conflict with the descriptor in Pythons chain, but we can do the work here an just redirect back to the Python wrapper layer.

native/python/pyjp_package.cpp
```diff
@@ -120,7 +124,30 @@ static PyObject *PyJPPackage_getattro(PyJPPackage *self, PyObject *attr)
...
+				ex.toPython();
+				JPPyErrFrame err;
+				err.normalize();
+				err.clear();
+				JPPyTuple tuple0 = JPPyTuple::newTuple(3);
+				tuple0.setItem(0, (PyObject*) self);
+				tuple0.setItem(1, attr);
+				tuple0.setItem(2, err.exceptionValue.get());
+				PyObject *rc = PyObject_Call(self->m_Handler, tuple0.get(), NULL);
+				if (rc == 0)
```
Thrameos, line 145: Error handler redirect is here.

test/jpypetest/test_exc.py
```diff
@@ -123,6 +123,11 @@ def testExcCtor(self):
         WE = jpype.JClass("jpype.exc.WierdException")
         with self.assertRaises(WE):
             WE.testThrow()
+        try:
+            WE.testThrow()
+        except Exception as ex:
+            ex1 = ex
+        self.assertEqual(ex1.args, ("Got it",))
```
Thrameos, line 130: And we can at last check that the message is indeed now the first argument.

Thrameos:
@marscher I am going to have to push this one in without a review.  The bugs in the master need to be fixed.   Two of the PRs had an odd interaction that slipped through while was missed when azure was not functioning.   I need this one to continue development.  Sorry.

I did a review that outlined the changes.   If you see something retroactively we can back this out or use another PR to clean it up.

PR #744, opened 2020-05-15
DBAPI2

Thrameos:
Okay bare with me on this one.  We currently have three different user bases.

- Scientific and Engineering codes
- A set of chat bots
- JayDeBeAPI users

We have picked up a lot in category 1 in the last year.  But the other two are still present at some level.  Unfortunately, JayDeBeAPI never made the transition from JPype 0.6.3 to 0.7.  Thus, it is the number one reason that JPype is pinned at an old version in a Python package.  This creates problems in many ways.  First, JayDeBeAPI hides JPype and launches it internally, thus we end up as a dependent package with zero visibility.  Next, the wrapper is very incomplete leaving a huge pile of holes that generate questions.  Last, it is pretty much abandoned leaving users to wander in here and ask questions about code I don't know anything about.  Anyone using it is basically stuck with buggy code and can't benefit from the huge improvements we have made.

When 687 lines of code is causing me headaches the nature response is rewrite it.  So this is my ham fisted attempt to deal with it.  I took a bunch of different testbenches that were license compatible (JayDeBeAPI is not), and began writing code to satisfy them.  Unfortunately, dbapi2 is really incomplete in Python compared to JDBC so there is a lot of holes.  Many of the DBAPI tests could never pass as one version expects one set of behavior and another version expects the opposite.  This left a lot of ambiguity so I read though a bunch of other DBAPI compatible wrappers and came up with a set of features the meets the needs when you have 3 different type systems playing together.   It is close to JayDeBeAPI where I could make it, draws pretty heavy on sqlite and also mysql and others.  Nothing quite covered the needs exactly so in some places I decided it was time to be creative.
 
Thus, my best solution is just build a better version of it and include it in our tool so that we don't lose that section of the community as we plow ahead.   Thoughts?  Objections?

**This PR is ready for user input... it is time to look of the API, suggest name changes, steer stuff that is wrong headed in a new direction, and try to shake a few bugs out.**

**If anyone has a test which can use callproc() to produce an OUT argument and an IN/OUT argument is would be great.**

**If anyone has a test which can exercise nextset with a Java compatible memory only database it would help**

Once the bugs are gone we can proceed to code review for inclusion in 0.8.


Thrameos:
Items that need to be done before this is releasable.

- [ ] Collect input from users (and reviews) regarding how JDBC types are presented.  The getters/setters/converters/adapters was invented to try to make it usable, but it is not clear if it meets the needs yet.
- [ ] Collect input on how SQLExceptions are transferred back to the user.   Python requires much finer grain exceptions than JDBC so splitting the exceptions to the right type is a chore.
- [ ] Add in common adapters that users are likely to use.  This is complicated by the fact that there are so many databases.  One group may want VARCHAR to accept memoryview and another not.  We don't want default behavior that causes problems.  So we may have to make a way to bulk import a set of adapters when the user wants to configure use.
- [ ] Finish the default converters (date/time/bigint).
- [ ] Add a real test bench preferably running more than one Java DB solution (anything that supports memory DB is fair game)
- [ ] Callable procedures
- [ ] Isolation levels
- [ ] Documentation

Thrameos:
In addition there is one unresolved technical question with fairly strong implications.   The code was modeled on sqlite driver which does not support Connection or Cursors being used cross thread so the same restriction was built into this implementation.  However, JDBC has a different concept of how threading works.   I can drop the restriction easily, but we need tests and also flags that properly report the threading model used by JDBC.

Does anyone know how JDBC threading support matches with Python dbapi2 model?

Thrameos:
Making good progress, but need to reconsider setters.  Many databases don't provide good information on the incoming type with the exception of inserts.  We will need to make the setter depend on the incoming type rather than the type requested by the database which may be very vague.

Thrameos:
@cgivre I think it may be ready for some user testing/input.    The first hack at some documentation is located at...

https://dev.azure.com/jpype-project/jpype/_build/results?buildId=409&view=artifacts&type=publishedArtifacts

cgivre:
@Thrameos 
Do you have any preference as to how I test this?  My main use case was with JayDeBeApi, so should I try that?

Thrameos:
@cgivre I think the first step would be to see how well it functions as a drop in replacement in some code that uses JayDeBeApi.   If you have a project which was using it lets try a switch.  There should be two modifications minimum.

- [ ] JayDeBeApi launched JPype internally, but that option was always a maintenance problem as the options to start the JVM from JPype never got shadowed.  So you will need to start JPype manually before you can use the database.

```
import jpype
jpype.startJVM(<other JVM args here>, classpath='<your class path here>')
```

- [ ] Try to pretend that jpype.dbapi2 is jaydebeapi
```
import jpype.dbapi2 as jaydebeapi
```

You can see from the PR that I have tested with "h2", "hsqldb", and "sqlite" (yes I know that python already has that driver).  I could not get derby running so I am interested if your project has any luck.

Thrameos:
@Szczepanov I know you are actively using JayDeBeAPI.  Does this PR look like a suitable long term replacement?

Thrameos:
@baztian I see that you have come out of hibernation.   I am working to eat your lunch as your project is lagging behind (nothing personal, but I just can't keep having my package pinned).  I had to start fresh as this is Apache 2 and your library in LGPL so I couldn't just fork.  If you would like to review this as a potential replacement it would help.  

Key differences between this and current.
- support for ``with`` statement.
- support for curproc and other more advance DBAPI2 features.
- support for pulling result set items with something other then getObject
- support for setting parameters with specified types.
- automatic translation to Java on to and from in most cases.
- mode switching from Java and Python types.

cgivre:
I'll just weigh in here re: licensing.  FWIW, it is much more helpful if this is licensed under the Apache license (or other license that is Apache compatible) than LGPL.

baztian:
@Thrameos I didn't look into your code yet. What is your preferred option? Do you like to develop your approach further or do you want me to resume development on jaydebeapi? Or do you like to contribute to jaydebeapi? 

Some features you mention sound nice.
For me personally I would love to keep jython support available which you probably won't implement.

In either way the most challenging part is keeping it compatible with various jdbc drivers.Therefore I find it very important to have a solid test approach. As the CI is currently somewhat broken I'm a bit restrictive in accepting pull request and am more or less blocked until I find the time fixing the CI issues. Also most of the PRs I received so far did not provide proper test or only worked with a limited number of databases. 

I don't get why lgpl is a problem for you. But I'm open on changing it to apache.

KaiTyrusNelson:
@baztian Long term I am just trying to give users an option so that people don't have to pin my library.  You are free to borrow as needed for your own subject to license compatibility.  So if some feature in mine looks interesting just steal from the Public Domain testbench and you can provide the same feature.    I did have to add some hooks in JPype to make it easier to do some things easier such as a way to check if a call would succeed to perform type conversions prior to calling and such.  I could wait for an exception to be thrown and then made the second attempt with additional cost.  But the advantage of better integration is that I can place in needed hooks.

I have no particular issue with LGPL.  I have contributed to many LGPL licensed projects in the past.  But LGPL (and GPL) is sort of a one way street so LGPL licensed code is not usable in JPype.   Thus since JPype is Apache 2 I can't simply borrow code from an LGPL project, but my understand is that may not be the same in reverse.  The JPype maintainer is pretty strongly against mixed licenses in this project..

Typical discussions on mixing Apache and LGPL.

https://opensource.stackexchange.com/questions/5664/linking-from-lgpl-2-1-software-to-apache-2-0-library

Having a clean rewrite should be beneficial to the community as I implemented a lot of features due to how I had to work from a clean start.   I started with all of the publicly available test benches for dbapi2 and then tried to see how many I could complete.  I then constructed an API that fit them all including whatever JDBC stuff that I could.   Then I rewrote the test benches to be coherent (many disagreed on what to raise, should it raise or return None, when should a behavior happen when etc).   I then put that whole test bench into Public Domain so that other authors like you could use it rather than being force to rewrite.

As far as the long term direction assuming that my version looks useful is that this driver will eventually be implemented in Java just using JPype as the bridge code rather than an actual Python module.  My long term plan is to make is so that JPype can implement Java classes from within Python and Python classes from within Java.  So for something like this were most of the code that is needed to talk with is on the Java side it will end up as a Java code with annotations make it automatically create a Python wrapper.  (JNI is fastest when few things cross the boundary so that means that stuff always needs to push the dominate side.  I am not sure how compatible that model will be with Jython.  Unfortunately, their code base wasn't useful to me when I started with JPype and it remains so today.   Looking at the Python object mode and the Java object model it would appear to be a tall order to get decent performance under the JVM.

There are unfortunately some Py3 only features that were used in my version, so it isn't clear if my API would translate to the current Jython.

Thrameos:
**Sorry my son's login appears to have been cached on the my browser.  Resending from the proper account.  **

@baztian Long term I am just trying to give users an option so that people don't have to pin my library. You are free to borrow as needed for your own subject to license compatibility. So if some feature in mine looks interesting just steal from the Public Domain testbench and you can provide the same feature. I did have to add some hooks in JPype to make it easier to do some things easier such as a way to check if a call would succeed to perform type conversions prior to calling and such. I could wait for an exception to be thrown and then made the second attempt with additional cost. But the advantage of better integration is that I can place in needed hooks.

I have no particular issue with LGPL. I have contributed to many LGPL licensed projects in the past. But LGPL (and GPL) is sort of a one way street so LGPL licensed code is not usable in JPype. Thus since JPype is Apache 2 I can't simply borrow code from an LGPL project, but my understand is that may not be the same in reverse. The JPype maintainer is pretty strongly against mixed licenses in this project..

Typical discussions on mixing Apache and LGPL.

https://opensource.stackexchange.com/questions/5664/linking-from-lgpl-2-1-software-to-apache-2-0-library

Having a clean rewrite should be beneficial to the community as I implemented a lot of features due to how I had to work from a clean start. I started with all of the publicly available test benches for dbapi2 and then tried to see how many I could complete. I then constructed an API that fit them all including whatever JDBC stuff that I could. Then I rewrote the test benches to be coherent (many disagreed on what to raise, should it raise or return None, when should a behavior happen when etc). I then put that whole test bench into Public Domain so that other authors like you could use it rather than being force to rewrite.

As far as the long term direction assuming that my version looks useful is that this driver will eventually be implemented in Java just using JPype as the bridge code rather than an actual Python module. My long term plan is to make is so that JPype can implement Java classes from within Python and Python classes from within Java. So for something like this were most of the code that is needed to talk with is on the Java side it will end up as a Java code with annotations make it automatically create a Python wrapper. (JNI is fastest when few things cross the boundary so that means that stuff always needs to push the dominate side. I am not sure how compatible that model will be with Jython. Unfortunately, their code base wasn't useful to me when I started with JPype and it remains so today. Looking at the Python object mode and the Java object model it would appear to be a tall order to get decent performance under the JVM.

There are unfortunately some Py3 only features that were used in my version, so it isn't clear if my API would translate to the current Jython.

Thrameos:
@cgivre Any luck on the testing?

cgivre:
@Thrameos 
Thanks for your patience... I'm having some issues getting this to connect to the JVM and find the driver.  

This is the code I'm using:
```
 jpype.startJVM(
                "/Library/Java/JavaVirtualMachines/jdk1.8.0_211.jdk/Contents/Home/jre/lib/server/libjvm.dylib",
                "-ea",
                "-Djava.class.path=" + self.jdbc_driver_path + self.jdbc_jar_name)
```

which is producing the error below.  I've verified that the driver jar is in that path.  Any suggestions?

```
-Djava.class.path=/Users/cgivre/github/drill-dev/drill/distribution/target/apache-drill-1.18.0-SNAPSHOT/apache-drill-1.18.0-SNAPSHOT/jars/jdbc-driver/drill-jdbc-all-1.18.0-SNAPSHOT.jar
  File "org.jpype.JPypeContext.java", line -1, in org.jpype.JPypeContext.callMethod
  File "Method.java", line 498, in java.lang.reflect.Method.invoke
  File "DelegatingMethodAccessorImpl.java", line 43, in sun.reflect.DelegatingMethodAccessorImpl.invoke
  File "NativeMethodAccessorImpl.java", line 62, in sun.reflect.NativeMethodAccessorImpl.invoke
  File "NativeMethodAccessorImpl.java", line -2, in sun.reflect.NativeMethodAccessorImpl.invoke0
  File "Class.java", line 264, in java.lang.Class.forName
  File "Class.java", line -2, in java.lang.Class.forName0
  File "ClassLoader.java", line 357, in java.lang.ClassLoader.loadClass
  File "ClassLoader.java", line 424, in java.lang.ClassLoader.loadClass
  File "org.jpype.classloader.JPypeClassLoader.java", line -1, in org.jpype.classloader.JPypeClassLoader.findClass
  File "ClassLoader.java", line 530, in java.lang.ClassLoader.findClass
Exception: Java Exception

  File "test.py", line 34, in <module>
    with jdbc_engine.connect() as con:
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/engine/base.py", line 2272, in connect
    return self._connection_cls(self, **kwargs)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/engine/base.py", line 104, in __init__
    else engine.raw_connection()
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/engine/base.py", line 2379, in raw_connection
    self.pool.unique_connection, _connection
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/engine/base.py", line 2345, in _wrap_pool_connect
    return fn()
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/base.py", line 304, in unique_connection
    return _ConnectionFairy._checkout(self)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/base.py", line 778, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/base.py", line 495, in checkout
    rec = pool._do_get()
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/impl.py", line 345, in _do_get
    c = self._create_connection()
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/base.py", line 309, in _create_connection
    return _ConnectionRecord(self)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/base.py", line 440, in __init__
    self.__connect(first_connect_check=True)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/base.py", line 661, in __connect
    pool.logger.debug("Error on connect(): %s", e)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/util/langhelpers.py", line 69, in __exit__
    exc_value, with_traceback=exc_tb,
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/util/compat.py", line 178, in raise_
    raise exception
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/pool/base.py", line 656, in __connect
    connection = pool._invoke_creator(self)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/engine/strategies.py", line 114, in connect
    return dialect.connect(*cargs, **cparams)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/SQLAlchemy-1.3.17-py3.7-macosx-10.9-x86_64.egg/sqlalchemy/engine/default.py", line 490, in connect
    return self.dbapi.connect(*cargs, **cparams)
  File "/Users/cgivre/github/jdbc_test/env/lib/python3.7/site-packages/JPype1-0.7.6.dev0-py3.7-macosx-10.9-x86_64.egg/jpype/dbapi2.py", line 404, in connect
    _jpype.JClass('java.lang.Class').forName(driver).newInstance()
jpype.dbapi2.ClassNotFoundException: java.lang.ClassNotFoundException: jdbc:drill:drillbit=localhost:31010
```

baztian:
You're probably specifying parameters to connect method in the wrong order. A sample code could help

Thrameos:
The failed line appears to passing the driver name to the url.  The driver is supposed to be   "org.apache.drill.jdbc.Driver" and the url should be 'jdbc:drill:drillbit=localhost:31010'.   It appears that jaydebeapi is requiring the driver name first.  'def connect(jclassname, url, driver_args=None, jars=None, libs=None):'  which would be a bad practice as using a specific driver to load has not been required since 1.6.

https://mkyong.com/jdbc/jdbc-class-forname-is-no-longer-required/

My implementation did the following.
```
def connect(url, driver=None, driver_args=None,
            adapters=_default, converters=_default,
            getters=GETTERS_BY_TYPE, setters=SETTERS_BY_TYPE, **kwargs):
```

But jaydebeAPI did

```
def connect(jclassname, url, driver_args=None, jars=None, libs=None):
```

I would expect a proper start up to look like ...
```
import jpype
import jpype.dbapi2 as dbapi2
    
jpype.startJVM('-ea', classpath=self.jdbc_driver_path + self.jdbc_jar_name)
dbapi2.connect('jdbc:drill:drillbit=localhost:31010', driver="org.apache.drill.jdbc.Driver")
```

As per the apache documentation, the driver name is 'org.apache.drill.jdbc.Driver':
https://drill.apache.org/docs/using-the-jdbc-driver/#example-of-connecting-to-drill-programmatically

Things to note:
- The style of startJVM with everything specified explicitly and requiring '-Djava.class.path' was deprecated after 0.6.3 because it was prone to typos (hence a constant question generator).  The keyword form is safer.   If you need more than one item on the classpath use a list.   That way you don't need to worry about the system separator.
- The driver name is likely not required.  In all of my testing with sqlite, h2, hsqldb, and others, the driver name was not needed.   Only when drivers fail to register themselves properly with meta data does a driver name become required.

I will add a more explicit catch statement to the driver so it produces better diagnostics.  

Looking through the pep it appears that both implementations are incorrect.  The correct arguments are called out in footnote 1.  I will revise to match the standard.

https://www.python.org/dev/peps/pep-0249/#id1

Unfortunately it will be tough to match jaydebeapi as it doesn't follow the PEP in this regard.   I suppose the safest thing I can do is force keyword only after the dsn which will make the breakage apparent.

The revised error message would be...
```
  File "test.py", line 5, in <module>
    dbapi2.connect('jdbc:drill:drillbit=localhost:31010', "org.apache.drill.jdbc.Driver")
TypeError: connect() takes 1 positional argument but 2 were given
```

cgivre:
HI @Thrameos 
I hate to ask you to debug my code, but here's the module I'm using: [1] and here is a link to the line where it opens the connection: [2]

I'm still getting the class not found exception.   Here is my test code: [3].  Would you mind taking a look?  I really want to get this to work as SQLAlchemy is used by a lot of tools and doesn't have a generic JDBC dialect for it.  My gut is saying that there is some argument is not being passed down the chain correctly.  

[1]: https://github.com/cgivre/sqlalchemy-drill/tree/jpype2
[2]: https://github.com/cgivre/sqlalchemy-drill/blob/bcfe14f0bd8328e38c16fbe7378fa661c7c65814/sqlalchemy_drill/jdbc.py#L82
[3]: https://gist.github.com/cgivre/4f51fb8affa308413a7648a2e23117a2

Thrameos:
Sure I can try to give it a run tonight after work.  I will need some step by step instructions on how to replicate the issue.

Thus far I cloned the repo, called ``python setup.py build`` and then ``python setup.py test``. Neither command fails (though I think it likely should as I haven't pointed it to the jpype install with dbapi2 yet.

Thrameos:
Unfortunately it doesn't look like we are going to get enough feedback on this one for it to make the next release.  So I am going to put a hold on it for now.

cgivre:
Hi Karl, 
I had some issues getting this to work, then we took a family vacation.  If you can hold off until Monday, I'll get you feedback over the weekend.

Thrameos:
Sounds good.  We should getting close to our release candidate this weekend assuming no issues pop up.

Thrameos:
@cgivre haven't heard back from you regarding the review of this work.  For this to be included I will need to locate another reviewer.    

I know that some other people have tested this branch using just the dbapi2 strict portion, but  thus far noone has looked over the extensions, documentation, and general  presentation.  If someone has interest it would help to add a review.

pelson:
> Unfortunately it doesn't look like we are going to get enough feedback on this one for it to make the next release. So I am going to put a hold on it for now.

*Taken from the perspective of a completely naive interested party in JPype who doesn't use JayDeBeAPI, but who cares about the long-term sustainability of JPype.*

Looking at the changes, I see a bunch of python code which looks like it could live equally well outside of JPype core in its own package (even under the jpype-project GitHub organisation). Is there a reason why is would be a bad thing to have them as separate projects with their own lifecycle? The positives include:

 * a reduced codebase & test suite in core JPype
 * a warm feeling that it is entirely possible to implement a powerful DB interface built on JPype as a third-party package (and somebody else could implement/maintain something similar without special JPype hooks)
 * the potential to maintain the Jython interface (mentioned above, not sure if that is something that is plausible/desirable in this API)

cgivre:
@pelson @Thrameos 
I apologize for not getting back with you.  I had some health issues last month and hence haven't had a chance to dig into this.  If you could give me another week or so, I can take a look and give some proper feedback from the Drill perspective.

Thrameos:
@cgivre Sorry to hear that you had heath problems.   I hope everything is okay.  I accidentally put a pick ax into my foot while working in the yard and was too cowardly to go into have it treated due to covid 19 concerns given lung scaring from an industrial accident working a summer job in high school.   Sliced straight to the bone on my heel.  Fortunately I have little feeling there due to prior nerve damage and I am finally back on my feet after several weeks of nursing it.

pelson:
> My main motivation for writing a JDBC driver is that the current project for supporting JDBC in Python completely buried the use of JPype.

Thanks for the context. I had understood most of this from your prior comments, and am 100% supportive of your motivation. It is clear that the interface is already useful (and used!), and it is also clear that anything that helps the community move away from v0.6 is a positive step.

> Adding additional steps to require triggering the extra packages to test each time someone submits a pull request to JPype would make supporting the interface less desirable.

Agreed. My proposal is much more than just "move the code into its own repository" though, it is about keeping their *lifecycles* independent (even if the dependent one (JDBC driver) binds to a specific JPype version). There are other projects which have a dependency on JPype which follow this approach (I maintain one of them :smile:), and this should be seen as a strength of JPype (along with the fact that you've been able to implement a powerful JDBC API without special JPype hooks) IMO.

> At least part of the code required hooks into JPype because the front end of JPype did not support conversions that were required for speed.

Are these the customizers you've added, or are they not part of this PR? If the former, then they can happily live as a standalone JDBC driver library as much as then can in ``jpype.protocol``, no? If the latter, it seems like you've added some stuff that can benefit any potential third-party library which uses the JDBC.

> Once the reverse bridge is completed JPype will actually compete with Jython more than it does today as we will occupy the same space. As JPype will support all CPython modules we may well leapfrog them.

Indeed, I think the recent improvements to JPype are already starting to pay dividends - I can have a full CPython interpreter and still access the full power of Java. From a "Scientific and Engineering codes" perspective it is clear that being able to use a standard scipy stack is a HUGE advantage for JPype (after all, the data science movement is almost exclusively the reason for Python's explosion in popularity). This isn't really a topic for this PR though, so I'll stop there :smile:.

Just to be completely clear, I personally don't mind if this PR goes in to JPype core, I just want to make sure that giving domain specific tools an independent life-cycle is considered in order to keep JPype as focused as possible and to allow the possibility of interested parties to be able to contribute/collaborate more easily (i.e. not be daunted by collaborating on JPype-core if you only care about some Python library that interfaces to JDBC through JPype).

doc/dbapi2.rst
```diff
@@ -0,0 +1,401 @@
...
+
+`threadsafety`_
+    The threadsafety level is 2 meaning "Threads may share the module and
+    connections".  But the actual threading level depends on the driver
+    implementation that JDBC is connected to.  Connections for many databases
+    are synchronized so they can be shared, but threads must execute statement
+    in series.  Connections in the module are implemented in Python and 
+    have per object resources that cannot be shared.  Attempting to use a
+    connection with a thread other than the thread that created it will
+    raise an ``Error``.
```
marscher, line 56: Do you mean RuntimeError or what kind of exception can users expect here?

marscher: I should have read further :)

marscher:
I strongly doubt, that creating new packages will help us in any way. I was maintaining a modular system and the cross-dependency testing between them turned out to be a nightmare, introducing lots of overhead with almost no real benefit. It had been mostly a political decision to split things up, without taking into account what additional burdens the development process will suffer. After my time it was decided to create a monolith (to lower the barrier for new developers to contribute, as reading into fragmented modules is considered more painful).

If the new protocol API does not interfere with existing projects wrapping JDBC I'm totally fine with merging now. By interference I mean that JPype claims to be the only implementation by registering wrappers for core Java API. If jpype.protocol does not get automatically imported, this should be sufficient I guess.

jpype/dbapi2.py
```diff
@@ -0,0 +1,1444 @@
...
+
+
+def _nop(x):
+    return x
+
+###############################################################################
+# Types
+
+
+class JDBCType:
```
marscher, line 53: I'd favor new-style objects (e.g. derive from object).

jpype/dbapi2.py
```diff
@@ -0,0 +1,1444 @@
...
+    def __repr__(self):
+        if self._name is None:
+            return ""
+        return self._name
+
+    def __eq__(self, other):
+        return other in self._values
+
+    def __hash__(self):
+        return hash(self._name)
```
marscher, line 132: are names meant to be unique?

jpype/dbapi2.py
```diff
@@ -0,0 +1,1444 @@
...
+# Setters take (connection, meta, col, type) -> JDBCTYPE
+def SETTERS_BY_META(cx, meta, col, ptype):
+    """ Option for setters to use the metadata of the parameters.
+
+    On some databases this option is useless as they do not track parameter
+    types.  This method can be cached for faster performance when lots of
+    parameters.  Usually types can only be determined accurately on inserts
+    into defined columns.
+    """
+    return _default_map[_registry[meta.getParameterType(i + 1)]]
```
marscher, line 251: where is i defined? Seems like broken/unused code to me

jpype/dbapi2.py
```diff
@@ -0,0 +1,1444 @@
...
+    Error = Error
+    Warning = Warning
+    InterfaceError = InterfaceError
+    DatabaseError = DatabaseError
+    InternalError = InternalError
+    OperationalError = OperationalError
+    ProgrammingError = ProgrammingError
+    IntegrityError = IntegrityError
+    DataError = DataError
+    NotSupportedError = NotSupportedError
```
marscher, line 449: are these static aliases just for convenience or is there a deeper logic behind this? E.g. downstream you refer to the outer defined exceptions

jpype/dbapi2.py
```diff
@@ -0,0 +1,1444 @@
...
+                    # find the column converter by type
+                    converter = cx._converters.get(type(value), _nop)
+                    row.append(converter(value))
+            return row
+        except TypeError as ex:
+            raise _UnsupportedTypeError(str(ex)) from ex
+
+    def _validate(self):
+        """ Called before any method that requires the statement to be open. """
+        if self._closed or self._jcx.isClosed() or threading.get_ident() != self._thread:
```
marscher, line 803: Please raise a different error in the mismatching thread case.

project/test/test_sql_regression.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def setUp(self):
+        common.JPypeTestCase.setUp(self)
+        self.con = dbapi2.connect(getConnection())
+
+    def tearDown(self):
+        self.con.close()
+
+    def testPragmaUserVersion(self):
+        # This used to crash pysqlite because this pragma command returns NULL for the column name
```
marscher, line 58: It is not clear to me, why we want to test against regressions in pysqlite.

marscher: But if this unit contains lots of useful cases, it might be worth checking them. I guess you have deleted unrelated cases already.

jpype/dbapi2.py
Thrameos: Fixed.   It got renamed to col.

jpype/dbapi2.py
Thrameos: This is a formality of the dbapi2 specification.  I am not sure why it is done but it appears in JayDeBeAPI, MySQL, and Sqlite.   So to conform with other packages I did the same.   My best guess is so that the user can import "Connection" from dbapi2 and then call the errors from the aliases.   But I would have to defer to dbapi2 user.

jpype/dbapi2.py
Thrameos: Will do.  Errors are very poorly defined in the the dbapi2 spec.   It seems like to make dbapi2 generic to whatever the underlying implementation that just left all the errors as "an Error will be raised" but at the same time defined a whole tree of error types.   This leaves it hard to decide what error type applies to what situation.   I split the thread error to an Error with the message indicating it is a thread problem.  But which specific error it should be is poorly defined.  I tried to read through other dbapi2 wrappers to try to puzzle it out, but found no consistency at all.

marscher (approved):

project/test/test_sql_regression.py
marscher: Thanks!

jpype/dbapi2.py
marscher: Thanks for the explanation.

jpype/dbapi2.py
marscher: Sorry for the confusion, I'm just saying that the error message should be specific to the issue (threading).

marscher: Ah seems like the threading issue is already firing up.

marscher:
Thanks a lot for this major improvement!

PR #749, opened 2020-05-20
Javadoc

Thrameos:
This is a PR for the javadoc extractor.

- [x] Add a directory search function that searches the zip (if registered) and classpath for javadoc.
- [x] Convert HTML to Restructured Text for each method, ctor, field, class.
- [x] Add caching method so that Classes will lookup the java doc only once per class.
- [x] Wire up the results to class, method and fields?

Javadoc can be pulled either from the jdk zip file or from individual jars using classloader.


Thrameos:
@petrushy  This is ready for your input, testing, etc.

@marscher this one is ready for review.    Though it is a giant pile of code fundamentally it is a very small change to JPype.   I added a HTML parser and Restructured Text renderer to the Java internal module.  The only place that actually touches JPype is on the class and method generator which now look to the internal module first before generating the usual minimalist documentation.   I through I would need additional API support but as it turns out you can place a zip file in the classpath and the classloader can unpack it and pull files like a file system.  So this mean the impact on the Python side was really small. 

The choice of going through the html rather than the source code is pretty much a no brainer.  Source distributions are not available regularly nor for the base Java documentation, and even if we parse the source we are still going to have to deal with html tags and also the javadoc tags.  Better to let the javadoc program deal with it so that we only have one set of annotations to work through.

The renderer and parser are not the best as it is challenging to replicate a full browser with small amounts of code.  I tried the Java solution to HTML parsing, but concluded it would not work as Javadoc is filled with HTML errors (double closed tags, unclosed tags, illegally unclosed tags) and thus I would need a parser that was rather forgiving.  Even then sometimes it fails and I just say punt, eat all the exceptions and send it the default doc generator.  I did make use of the DOM in Java rather than writing my own though that presents some challenges as DOM has its own issues especially if you try to transform it.   If we can find an industrious user that is willing to make the transform into an XSTL rather than a big blob of code it would help, but XSLT is not in my programming arsenal.

native/java/org/jpype/javadoc/JavadocRenderer.java
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Render a node as ReStructured Text.
+ *
+ * @author nelson85
+ */
+public class JavadocRenderer
+{
+
+  static final int WIDTH = 80;
```
marscher, line 17: I guess 120 width is also fine nowadays. For instance jupyter notebooks render it with that, if I recall correctly.

project/utility/testDoc.py
```diff
@@ -0,0 +1,70 @@
+import _jpype
```
marscher, line 1: development stub?

marscher, line 1: devel stub?

Thrameos: Used on line 54.   (I can fix it once JPackage isinstance logic goes in)

native/java/org/jpype/javadoc/JavadocRenderer.java
Thrameos: Change to 120 in final integration PR.

PR #755, opened 2020-05-26
Jchar

Thrameos:
At long last I have completed all of the return contracts in JPype.  Every type with the exception of Boolean and null (which must return singletons) now correctly returns the expected Java type.  This one was a huge amount of work because Char needed to act both as being able to promote to integers and at the same time be a string.

With this we are now ready to feature freeze for JPype 0.8 once all of the PRs clear.  We will then need to see if we can resolve the current shutdown issues and make one last pass to increase coverage to see if there are additional bugs requiring work.

Bugs fixed here:

- Leaks of Python objects created when the ctor produces an exception were dealt with.
- richcompare needs to use not implement rather than throw so that it goes to the other type to see if the operator is supported there.  (Discovered by reading Python source code)


Thrameos:
@marscher If you can get some of the PRs reviewed this week I can start some integration testing next weekend.  If you don't have time for it, I will refresh the devel branch and work on the integration testing there.

marscher:
I will look into it this week. Thank you

On May 26, 2020 2:36:00 PM UTC, Karl Nelson <notifications@github.com> wrote:
>@marscher If you can get some of the PRs reviewed this week I can start
>some integration testing next weekend.  If you don't have time for it,
>I will refresh the devel branch and work on the integration testing
>there.  
>
>-- 
>https://github.com/jpype-project/jpype/pull/755#issuecomment-634065100

-- 

test/jpypetest/test_boxed.py
```diff
@@ -169,8 +169,9 @@ def testPrivitiveToBoxed(self):
         java.lang.Double(JDouble(0))
 
     def testBooleanBad(self):
-        with self.assertRaises(TypeError):
-            java.lang.Boolean(tuple())
+        # with self.assertRaises(TypeError):
+        # This test was wrong.   'bool(tuple()) == False'
+        self.assertFalse(java.lang.Boolean(tuple()))
```
marscher, line 174: The same thing applies to empty lists, dicts and sets, doesn't it?

test/jpypetest/test_conversion.py
```diff
@@ -107,9 +107,10 @@ def testCharConversion(self):
         self.assertEqual(JChar._canConvertToJava(bytes([1])), "implicit")
         self.assertEqual(JChar._canConvertToJava(bytes([1, 1])), "none")
 
-    def testCharOverflow(self):
-        with self.assertRaises(OverflowError):
-            JChar(1000000)
+     # This test is wrong 'char q = (char) 1000000;' works in Java
```
marscher, line 110: If a java char is 16 bit long, the cast would overflow. So by "works" you mean there is no exception thrown?

test/jpypetest/test_jchar.py
```diff
@@ -56,9 +54,10 @@ def testJPJavaFrameCharArray(self):
         _jpype.fault("JPJavaFrame::ReleaseCharArrayElements")
         with self.assertRaisesRegex(SystemError, "fault"):
             ja[0:3] = bytes([1, 2, 3])
-        _jpype.fault("JPJavaFrame::ReleaseCharArrayElements")
-        with self.assertRaisesRegex(SystemError, "fault"):
-            jpype.JObject(ja[::2], jpype.JObject)
+        # Not sure why this one changed.
```
marscher, line 57: So it does not trigger the releaseCharArrayElements any more? Which wrapper is used instead?

test/jpypetest/test_conversion.py
Thrameos: Yes.  The test was looking for an exception but by the Java standard there should not be one.   Our implementation is that explicit cases should not throw, but implicit conversions should.   This test was therefore testing for an incorrect behavior.

test/jpypetest/test_boxed.py
Thrameos: Yes.  I am not sure I like this behavior too much as it makes JBool and java.lang.Boolean very permissive.  But since that technically is Python rules regarding apply bool to complex object , it the behavior that the user expected on an explicit cast.  I guess I should test the implicit case as well.

test/jpypetest/test_jchar.py
Thrameos: Hmmm.  Good question.   I will track this one down and make sure  there is not a new leak here.    Nothing I was touching in this PR should have touched the array paths.  But a closer look is warrented.

marscher:
I'm a bit confused by the statement, that numerical conversions are still possible, but in the testsuite there are lots of exceptions being thrown in numerical operations. What am I missing here?
Anyways very good job on the fixing this. It looks like as it was a tough task to fulfill.

Thrameos:
I was trying to break as little code a possible, but also be able to have it follow java rules.   Our previous implementation had two properties that were vital.   First whenever the user got one back it must of type ``str`` and be able to be used with ``str`` operations.   But implementation details on the casting were backed by ``int`` so we ended up with a really horrible contract.  It was completely a string on a return, and completely a number on the parameter side.

So the question is what does Java do....   so I wrote some programs to find out.

`` 
   char c0, c1, c2;
   c0 = c1+1; // fail as c1 was promoted to int
   c0 = c1 + c2; // fail as c1 and c2 promoted to int
   c0 = c1*2; // fail as c1 is promoted to int
   c0 = (char)(c1+1); // pass so you can cast an int back to a char after doing math.
```

So my conclusion is any numerical operation you do to a JChar should make it turn into an int but otherwise it should be like a length 1 string.  So I made that happen by first making a class that derived from string.  This was rather challenging is there is no methods in Python that actually let you construct with type so I had to replicate it all and as you can see there were three different rules need for the ctor (all well documented).  That just left how to make it also automatically promote to int.  That was more challenging.  I could have left it as `` int(c1)+1`` but that would imply an explicit casts is required which does not match Java.  So instead I had to implement each of the operators that applies.  

I also turned up bugs in the comparison operators when I went to mix floats and chars.  It was not very well documented in Python  but there were rules that had to be obeyed to make that happen.   

As for the potential to break code I think it is very low.   Anyone using JChar as documented (calling it to cast to hit an overload or getting it back as a field or return value) will get exactly the same behavior.   If they were making a JChar and doing math with it, it would also act the same by acting like an int.  The edge cases are math overloads that conflict with string.   These are JChar('a')*10 and JChar('a')+1 ... here Java and Python rules are in conflict.   Java has it cast to int first and do a numerical operation, while Python would have it replicate 'a' ten times and cast 1 to str first.   This is really in the corner case domain so I said is Java type so should follow Java rules.

marscher:
Thanks for the detailed explanation. It has shed light on the thing now and I see the major improvement for the type wrapper. Please merge this accordingly, when the conflicts are resolved.

test/jpypetest/test_jchar.py
Thrameos: This was a transient problem (happens when I jump branches, compile, then jump back.)  Re-enabled the test and all is good.

Thrameos:
I know it seemed like a trivial detail to complete the last wrapper and char really seems like like a huge waste of time, but as a hard core completionist  I can't let it go without finishing the mission even if it was an escort mission.

PR #762, opened 2020-05-28
Named

### LINKED ISSUE #747: JProxy reference leak
While running the test bench with stdout turned on, this exception warning turned up.   Likely something in the exception path on proxies is leaking.

```
test/jpypetest/test_proxy_multithreaded.py::ProxyMultiThreadedTestCase::testProxyWithSingleThreadExecution java.lang.RuntimeException: Local reference outside of frame

        at org.jpype.proxy.JPypeProxy.hostInvoke(Native Method)
        at org.jpype.proxy.JPypeProxy.invoke(Unknown Source)
        at com.sun.proxy.$Proxy25.testMethodInt(Unknown Source)
        at jpype.proxy.ProxyExecutor$ProxyCaller.call(ProxyExecutor.java:108)
        at jpype.proxy.ProxyExecutor$ProxyCaller.call(ProxyExecutor.java:94)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

### LINKED ISSUE #752: JPackage isinstance fails checks on _JPackage instances
JPackage instancecheck is not implement to properly report ``_JPackage`` instances as True.  This is a minor bug affected meta programming (walking the type tree for documentation and stub extractors.)

### LINKED ISSUE #756: Classes created more than once in multithreading
While debugging a stray exception in the multithread proxy test I turned up a race condition which is allowing multiple instances of a class to be created.  This needs to be resolved for the release.

```
  File "native/python/pyjp_package.cpp", line 194, in PyJPPackage_getattro
New class  jpype.proxy.ReturnObject
New class  java.util.Arrays
New class  java.util.Arrays
  File "native/python/pyjp_package.cpp", line 194, in PyJPPackage_getattro
    JP_PY_CATCH(NULL);
New class  java.util.Arrays
  File "native/python/pyjp_class.cpp", line 1023, in PyJPClass_create
New class  java.util.Arrays
New class  java.util.Arrays
New class  java.util.RandomAccess
  File "native/python/pyjp_package.cpp", line 194, in PyJPPackage_getattro
```

### LINKED ISSUE #757: Reference leak in casting operator
While tracing an unexplained exception, I identified an instance in which an object inappropriately called ``keep()`` on a frame allowing the object to leak into the global scope.   To prevent this from happening the JPJavaFrames need to be segregated into three types.  External frame which are born in Java, outer frames that are not allowed to leak any resources, and inner frames which are used to control resources within outer and external frames.   This should prevent further leaks of this type as all  Java actions must be done in a frame and if those frames have rules that prevent leaking it should not be possible to leak a resource.

Thrameos:
This is a PR dealing with a few of the bugs in the issues list.  It will pull in other PRs are we work towards an integration test.  This PR also includes a coverage pass and corrections to APIs discovered during the coverage pass.

Fixes #747
Fixes #756
Fixes #757
Fixes #752


jpype/_jcollection.py
```diff
@@ -176,14 +176,6 @@ def remove(self, obj):
...
 
 
-def isPythonMapping(v):
-    if isinstance(v, Mapping):
-        if not hasattr(v.__class__, '__metaclass__') or \
-           v.__class__.__metaclass__ is _jclass._JavaClass:
-            return True
-    return False
-
-
```
Thrameos, line 186: Dead code removed

jpype/_jpackage.py
```diff
@@ -19,7 +19,15 @@
...
-class JPackage(_jpype._JPackage):
+class _JPackageMeta(type):
+    def __instancecheck__(self, other):
+        return isinstance(other, _jpype._JPackage)
+
+    def __subclasscheck__(self, other):
+        return issubclass(other, _jpype._JPackage)
+
+
+class JPackage(_jpype._JPackage, metaclass=_JPackageMeta):
```
Thrameos, line 30: This is needed so that imported packages test as being JPackage instances.

jpype/_jexception.py
```diff
@@ -30,9 +30,6 @@ class JException(_jpype._JException, internal=True):
 
     """
 
-    def __str__(self):
-        return str(self.toString())
-
```
Thrameos, line 35: For reasons that are not quite clear removing this line seems to change the output message slightly even through I don't see a reason.   I removed it as part of a general clean up of unnecessary methods.

native/common/include/jp_exception.h
```diff
@@ -67,9 +67,6 @@ extern int _method_not_found;
 #define JP_STACKINFO() JPStackInfo(__FUNCTION_NAME__, __FILE__, __LINE__)
 
 
-// Macro to all after executing a Python command that can result in
-// a failure to convert it to an exception.
-#define JP_PY_CHECK() { if (JPPyErr::occurred()) JP_RAISE_PYTHON();  } // GCOVR_EXCL_LINE
```
Thrameos, line 72: There were a lot of legecy things left over from the previous Python wrapper when this was to support ruby.  As they serve no purpose other than obscuring the function at this point I have removed them.

native/common/include/jpype.h
```diff
@@ -208,11 +209,4 @@ typedef _object PyObject;
...
 #include "jp_typemanager.h"
 
-// Accessors
-#include "jp_array.h"
-#include "jp_buffer.h"
-
-// Services
-#include "jp_proxy.h"
-
-#endif // _JPYPE_H_
```
Thrameos, line 218: There was  a lot of global includes.  This PR finished a lot of that clean up work by pushing the last few specializazed wrappers out of the global space.

native/common/jp_exception.cpp
```diff
@@ -432,6 +425,25 @@ void JPypeException::toJava(JPContext *context)
...
+		if (m_Type == JPError::_python_exc)
+		{
+			JPPyCallAcquire callback;
+			// All others are Python errors
+			JP_TRACE(Py_TYPE(m_Error.l)->tp_name);
+			PyErr_SetString((PyObject*) m_Error.l, mesg.c_str());
+			convertPythonToJava(context);
+			return;
+		}
+
```
Thrameos, line 446: This section is to correct a bug in which Python type information was lost when returning to Python for one specific path.

native/python/pyjp_monitor.cpp
```diff
@@ -35,27 +35,37 @@ static int PyJPMonitor_init(PyJPMonitor *self, PyObject *args)
...
+		PyErr_SetString(PyExc_TypeError, "Java strings cannot be used to synchronize.");
+		return -1;
+	}
 
 	if ((v1->getClass())->isPrimitive())
-		JP_RAISE(PyExc_TypeError, "Java primitives cannot be used to synchronize.");
+	{
+		PyErr_SetString(PyExc_TypeError, "Java primitives cannot be used to synchronize.");
+		return -1;
+	}
```
Thrameos, line 62: There were lots of spots where the C++ exception path (slow) was being taken just to dump a Python exception back.   This is unnecessary so all these were removed.

native/python/pyjp_method.cpp
```diff
@@ -117,25 +128,21 @@ static PyObject *PyJPMethod_str(PyJPMethod *self)
...
-		ss << "<java method `";
-	else
-		ss << "<java bound method `";
-	ss << self->m_Method->getName() << "' of '" <<
-			self->m_Method->getClass()->getCanonicalName() << "'>";
-	return JPPyString::fromStringUTF8(ss.str()).keep();
+	return PyUnicode_FromFormat("<java %smethod '%s' of '%s'>",
+			(self->m_Instance != NULL) ? "bound " : "",
+			self->m_Method->getName().c_str(),
+			self->m_Method->getClass()->getCanonicalName().c_str());
```
Thrameos, line 145: Indirectly going through C++ stringstream when there is already a Python format to unicode is unnecessary and has been removed.

native/python/jp_pythontypes.cpp
```diff
@@ -21,38 +21,72 @@ static void assertValid(PyObject *obj)
...
+ * Will throw an exception an error occurs.
+ */
+JPPyObject JPPyObject::call(PyObject* obj)
+{
+	JP_TRACE_PY("pyref new(call)", obj);
+	JP_PY_CHECK();
+	ASSERT_NOT_NULL(obj);
+	assertValid(obj);
+	return JPPyObject(obj, 3);
+}
```
Thrameos, line 85: These specific cases were being mapped to a big tree of options not all of which were used.  I have replaced it with named constructors which makes it much more clear what is going on at any point and easier for the coverage tool to analyze as rather than a pile of branches which 90% were not used in any single case it instead is simple straight forward code at each usage.

native/common/include/jp_javaframe.h
```diff
@@ -65,15 +69,39 @@ class JPJavaFrame
 	 * @throws JPypeException if the jpype cannot
 	 * acquire an env handle to work with jvm.
 	 */
-	JPJavaFrame(JPContext* context, int size = LOCAL_FRAME_DEFAULT);
+	static JPJavaFrame outer(JPContext* context, int size = LOCAL_FRAME_DEFAULT)
+	{
+		return JPJavaFrame(context, NULL, size, true);
+	}
```
Thrameos, line 75: In order to make sure we can't leak Java objects we need to segregate frames which are inner (controlling scope) and outer (meaning no java objects can survive.   I have switched these to named constructors so that it is clear which is which,

marscher: That's a very meaningful change. Thanks!

native/python/include/jp_pythontypes.h
```diff
@@ -44,36 +55,33 @@ typedef _object PyObject;
...
+ *
  */
-namespace JPPyRef
+class JPPyObject
 {
+	/** Create a new reference to a Python object.
+	 *
+	 * @param obj is the python object.
+	 */
+	JPPyObject(PyObject* obj, int i);
```
marscher, line 66: what is the i parameter used for? If you document the obj parameter, I'd recommend to do the same for i.

Thrameos: The i parameter was placed simply to make sure that signature is not hit except by the static method callers.  I could just use explicit but I needed to make sure it busted in every place the only method was used.  I can remove the extra parameter now if you would like me to.

native/python/include/jp_pythontypes.h
```diff
@@ -243,61 +203,39 @@ class JPPyString : public JPPyObject
...
-
 /** Wrapper for a Python sequence.
  *
  * In most cases, we will not use this directly, but rather convert to
  * a JPPyObjectVector for easy access.
  */
-class JPPySequence : public JPPyObject
+class JPPySequence
 {
+	JPPyObject obj_;
```
marscher, line 213: hopefully not too pendantic, but you started to use m_ prefix for members all around the place (which I think is good practice) - so why not here?

native/python/include/jp_pythontypes.h
```diff
@@ -243,61 +203,39 @@ class JPPyString : public JPPyObject
...
 	{
+		return JPPySequence(obj);
 	}
 
 	JPPyObject operator[](jlong i);
 
 	jlong size();
+
+private:
+	JPPySequence& operator= (const JPPySequence& ) ;
```
marscher, line 237: you allow copy construction, but forbid assignment *head scratching*

native/python/jp_pythontypes.cpp
```diff
@@ -21,38 +21,72 @@ static void assertValid(PyObject *obj)
...
+
+	return JPPyObject(obj, 1);
+}
+
+/**
+ * This policy is used when we are given a new reference that we must
+ * destroy.  This will steal a reference.
+ *
+ * Assert not null, claim reference, and decremented when done.
+ * Will throw an exception in the object is null.
```
marscher, line 61: s/exception in/exception if"

native/python/include/jp_pythontypes.h
Thrameos: I think I tried to remove both and one didn't work so I just put it back.  I will reexamine.

native/python/include/jp_pythontypes.h
Thrameos: Fixed.

native/python/include/jp_pythontypes.h
Thrameos: This is a odd edge case of using named constructors.  In order to use a named constructor the class must have a copy constructor even through that copy constructor will not be called due to optimizations.   On the other hand assignment operator is not required.    

I am not sure why C++ has this formulation (and I have tested and found that the copy constructor is optimized out), but it is still required.

project/jpype_java/test/org/jpype/manager/TypeFactoryHarness.java
```diff
@@ -389,6 +389,12 @@ public void failFindMethod(ClassDescriptor desc, Method requestedMethod)
 //</editor-fold>
 //<editor-fold desc="inner" defaultstate="collapsed">
 
+  @Override
+  public void newWrapper(long context, long cls)
+  {
+    throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
marscher, line 395: pls delete this comment :)

test/jpypetest/test_collection.py
```diff
@@ -4,86 +4,84 @@
 import collections.abc
 
 
-class CollectionTestCase(common.JPypeTestCase):
+class CollectionCase(common.JPypeTestCase):
 
     def setUp(self):
-        super(CollectionTestCase, self).setUp()
+        common.JPypeTestCase.setUp(self)
```
marscher, line 10: whats wrong with super?

Thrameos: When I was copying and pasting around the test code and changing the method names the ``super`` ended up tripping me because it forced the name of the test case to match.  Thus I just switched to to match all of the the others.

If you would like me to use a script to make everything consistently use super instead I can also make that happen.

Thrameos:
Thanks every so much for the quick review.  Hopefully we can prep for the release this week.   I will try to put in some more coverage checks so that we can finally define the API well enough to detect any changes in the future releases.

PR #763, opened 2020-05-31
Alpine

### LINKED ISSUE #759: Issue installing on Alpine Linux 0.7.3 -> current (0.7.5)
**High level description:**
JPype1 (0.7.3 -> current) is unable to install correctly on Alpine Linux

**Steps to reproduce:**
1. docker run --rm -it --entrypoint=sh alpine:latest
2. apk add python3-dev gcc build-base
3. pip3 install JPype1

**Expected Result**
- Installation Success 

**Actual Result**
- Install fails due to alpine's musl libc not including mallinfo (which is used in your jp_gc.cpp). mallinfo is apparently a GNU glibc addition and thus not in alpine's implementation.

Thrameos:
Change memory get routine to use /proc/self on linux rather than glibc mallinfo.  This is a draft until the user tests.

Fixes #759


native/common/jp_gc.cpp
```diff
@@ -63,6 +87,11 @@ void JPGarbageCollection::triggered()
 
 JPGarbageCollection::JPGarbageCollection(JPContext *context)
 {
+#if defined(USE_PROC_INFO)
+	statm_fd = open("/proc/self/statm", O_RDONLY);
```
marscher, line 91: you should also close this fh again.

Thrameos: Done.  I also set the default for linux back to GLIBC when available.   The /proc/self/statm figures does not shrink when memory is releases so it is hard to properly set the lower bounds.

Thrameos:
@marscher Is this one now good to go?   We are just two PRs from having a release candidate version ready.  Then we have to figure out how to get it tested.

dm-ion:
Hey @marscher testing passed.

You can see the thread: https://github.com/jpype-project/jpype/issues/759

marscher:
Great thanks for your efforts!

PR #776, opened 2020-06-19
Review release

Thrameos:
This draft release contains changes for review by DOE.   May require additional changes as identified by DOE.


Thrameos:
@marscher I moved all of the copyright notices and other boiler plate statements to the NOTICES file so that I could apply mass licence application tools and so the my organization doesn't require their full notice attached to every file.  I believe we discussed this previously and you approved.   Can you verify this is correct?  There was only one file I have located which was not listed as authored by Steve or myself.

Thrameos:
@marscher The review is complete and a release number was issued for the current code.  I removed the file test/sample as it was identified as having a different license/copyright holder.  I also fixed two minor errors.  The review instructions recommend converting the license header to https://spdx.dev/, but it was not a requirement and we can do so after the release.  I will merge this once everything is clear in CI and we can start with the release process this weekend.

ShaheedHaque:
I don't see 1.0.0 on PyPI yet...is that expected?

Thrameos:
I got everything through the release system late last night (there was a problem with the manylinux pattern which was challenging to replicate locally).  Assuming nothing appears to be grossly broken in this first big release @marscher, the keeper of the keys, can push it through to PyPI and Anaconda when when he is next available.

PR #794, opened 2020-07-17
Optimization for speed

Thrameos:
Reworked paths so that repeated calls to same overload bypass the normal path for speed.

### DIFF
--- a/native/common/jp_classhints.cpp
+++ b/native/common/jp_classhints.cpp
@@ -64,13 +64,20 @@ JPMethodMatch::JPMethodMatch(JPJavaFrame &frame, JPPyObjectVector& args, bool ca
 	m_Overload = 0;
 	m_Offset = 0;
 	m_Skip = 0;
-	m_Cache = callInstance ? 0 : 1000;
+	m_Hash = callInstance ? 0 : 1000;
 	for (size_t i = 0; i < args.size(); ++i)
 	{
 		PyObject *arg = args[i];
 		m_Arguments[i] = JPMatch(&frame, arg);
-		m_Cache *= 0x10523C01;
-		m_Cache += (long) (Py_TYPE(arg));
+
+		// This is an LCG used to compute a hash code for the incoming
+		// arguments using (A*X+A_i) mod2^64 where A_i is the address of each
+		// type the argument list.  The hash will be checked to avoid needing
+		// to resolve the method if the same overload is called twice. There
+		// is only a speed cost if there is a collision, so we don't need to
+		// prove this is a perfect hash function.
+		m_Hash *= 0x10523C01;
+		m_Hash += (long) (Py_TYPE(arg));
 	}
 }
 
--- a/native/common/jp_classhints.cpp
+++ b/native/common/jp_classhints.cpp
@@ -56,17 +56,21 @@ jvalue JPMatch::convert()
 	return conversion->convert(*this);
 }
 
-JPMethodMatch::JPMethodMatch(JPJavaFrame &frame, JPPyObjectVector& args)
-: argument(args.size())
+JPMethodMatch::JPMethodMatch(JPJavaFrame &frame, JPPyObjectVector& args, bool callInstance)
+: m_Arguments(args.size())
 {
-	type = JPMatch::_none;
-	isVarIndirect = false;
-	overload = 0;
-	offset = 0;
-	skip = 0;
+	m_Type = JPMatch::_none;
+	m_IsVarIndirect = false;
+	m_Overload = 0;
+	m_Offset = 0;
+	m_Skip = 0;
+	m_Cache = callInstance ? 0 : 1000;
 	for (size_t i = 0; i < args.size(); ++i)
 	{
-		argument[i] = JPMatch(&frame, args[i]);
+		PyObject *arg = args[i];
+		m_Arguments[i] = JPMatch(&frame, arg);
+		m_Cache *= 0x10523C01;
+		m_Cache += (long) (Py_TYPE(arg));
 	}
 }
 
--- a/native/common/jp_methoddispatch.cpp
+++ b/native/common/jp_methoddispatch.cpp
@@ -27,7 +27,7 @@ JPMethodDispatch::JPMethodDispatch(JPClass* clazz,
 	m_Class = clazz;
 	m_Overloads = overloads;
 	m_Modifiers = modifiers;
-	m_LastCache.m_Cache = -1;
+	m_LastCache.m_Hash = -1;
 }
 
 JPMethodDispatch::~JPMethodDispatch()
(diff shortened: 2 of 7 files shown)


Thrameos:
After discussion with some users it appears that speed is some peoples primary concern as opposed to stability, ease of use, etc.   Though I personally believe that speed is great but if you give up safety to do so it is to your detriment.  However in this case if we can have both speed, ease and safety it is all the better.   So I added some caches so that repeated calls to the same overload will bypass the normal argument resolution if they have the same types.  This doesn't work for variadic parameters as it it can lead to a bad decision if the arguments are packed in a list.  To make is safe against potential sheer between threads it simply validates the arguments rather the running a full match.  If the validation fails it then proceeds to the standard method so a cache miss just falls to the usual cost.  In most cases in a typical program the same overloads are hit routinely this should increase speed on average by a fair bit. 

There are still a few places where we can improve speed on bulk transfers, but method calling is the lowest hanging fruit.   We already had pretty darn optimized paths for the method after I converted it to the match and convert once, but takes another 50% out of the path in many cases where there were overloads.

native/common/jp_classhints.cpp
```diff
@@ -56,17 +56,21 @@ jvalue JPMatch::convert()
...
+	m_Overload = 0;
+	m_Offset = 0;
+	m_Skip = 0;
+	m_Cache = callInstance ? 0 : 1000;
 	for (size_t i = 0; i < args.size(); ++i)
 	{
-		argument[i] = JPMatch(&frame, args[i]);
+		PyObject *arg = args[i];
+		m_Arguments[i] = JPMatch(&frame, arg);
+		m_Cache *= 0x10523C01;
```
marscher, line 72: What's that magic number arising from? When does the cache overflow and is that handled properly? I think a function a can have up to 255 arguments. Eventually I'm missing something here.

native/common/jp_methoddispatch.cpp
```diff
@@ -45,46 +46,57 @@ bool JPMethodDispatch::findOverload(JPJavaFrame& frame, JPMethodMatch &bestMatch
 	JP_TRACE("Checking overload", m_Name);
 	JP_TRACE("Got overloads to check", m_Overloads.size());
 	JPMethodList ambiguous;
-	JPMethodMatch match(frame, arg);
+	JPMethodMatch match = bestMatch;
```
marscher, line 49: can't this be a reference (avoid a copy)?

native/common/jp_classhints.cpp
```diff
@@ -56,17 +56,21 @@ jvalue JPMatch::convert()
...
+		m_Cache += (long) (Py_TYPE(arg));
```
marscher, line 73: This assumes that all casts of the pointer ob_type are smaller than the magic number to achieve proper separation of all types, am I wrong? I like this idea, but can you demonstrate that this holds in most or all cases?

native/common/jp_methoddispatch.cpp
Thrameos: Unfortunately no.  We need to two copies here.  One is the used to hold the best match we have found and one is used to test the next match (which destroys the state of the match).   We are doing a copy here because we happened to have run the same command out one layer.   There was no reason to rerun the command when we could just copy the results.

marscher: I see. Thanks!

native/common/jp_classhints.cpp
Thrameos: In this case the m_Cache is just a psuedo random number generator that is computing ((((X0*A+X1)*A+X2)*A+X3)*A+X4%2^n using a LCG (https://en.wikipedia.org/wiki/Linear_congruential_generator) where the incoming addresses are providing the randomness in the form of c to the classic formula (A*X+c) mod 2^n.   

The magic of the number is just to spread the bits picked in this case by randomly bashing keys on the keyboard so that we are less likely to have collisions because memory addressing  will have very similar address keys.    The number is expected to just keep overflowing due to 2^n mod math.  The number using for spreading doesn't need to be prime in the field, it just has to have sufficient entropy that we are unlikely to get a sequence where A*X1+X2 = A*X2+X1.  

The good news is about this particular algorithm is collisions don't actually matter.  Say we have 3 overloads which because of the way the user is hitting them with arguments are forming 5 different hits in the table.  That will form 5 hash codes in our m_Cache.  Each time one is called and it doesn't match the previous call, we cache the method and random code assigned.  If the hash in m_Cache does match we test the cached method to see if it is a match.   If two of the methods have a collision of their argument cache, we pay a cache miss on the test of the next method call (just loss of speed), but if we have a cache hit we get to realize the last resolution address skipping deciding which of the 3 overloads to use.  So we get cost=1 on a hit, cost=3 on a miss (we find no match on the hash), and cost=4 on a collision (one extra test).  Assuming  typical programming with 80% hits, our speed is .8*1+0.2*3.  If one of the 5 has collision with 10% usage it would be 0.7*1+0.2*3+0.1*4.  So given that there is no cause for alarm if we get a collision, and the chance of getting a collision given the entropy of the inputs is staggeringly low this is a pretty safe algorithm for deciding is the user hitting the same overload again and again.

I will add the reference to this being a LCG to make it more clear.  (I could have gone and written a program to help chose the optimal magic number given a set of addresses, but that would have been way more work that randomly generating a number by blindly typing on the keyboard)   Maybe I should have named it m_Hash?

native/common/jp_classhints.cpp
Thrameos: See comment above.   If you are worried I can actually generate a real LCG magic number (assuming c is zero) by creating a number which is prime in a field of 2^64.  It is just more effort.

native/common/jp_classhints.cpp
marscher: Agreed upon the naming (it is in fact a hash) and thanks for the detailed explanation. Did you have the chance to measure the impact of this? Not that I'm requiring it, but just being curious :+1:

Thrameos: Yeah the numbers are in a thread about 1/3 of the way down for ArrayList access https://github.com/kivy/pyjnius/issues/551.   It really depending on the number of overloads in the method and the usable pattern.   Obviously it does best when you try to do a benchmark as you get nothing but hits, but in general people tend to feel this most when they execute an iterator or list comprehension where it is going hit the same pattern hard.  

Before...
Pack arraylist 2.768904e-06
Iterate arraylist 5.208071e-06
Access(get) arraylist 4.037985e-06
Access([]) arraylist 4.690264e-06

After...
Pack arraylist 2.226081e-06
Iterate arraylist 4.082152e-06
Access(get) arraylist 2.962606e-06
Access([]) arraylist 3.644642e-06

It was more significant than I expected because the bypass path actually cut out some time even when there was no overloads because it simplified the logic and reduced the memory setup for overload resolution.

Thrameos: Actually I bet I can move the copy command after the cache check to see if we can short cut this before the list argument copy.   The method overload check is destructive so it really won't matter if the cache check results are copied in on the first check.

Thrameos: @marscher I updated the documentation and did reordered it to cut off one more list copy.  Should be good to go.

marscher: Thank you very much!

PR #795, opened 2020-07-17
fix typo: ImportException -> ImportError

michi42:
In the java exception handler of imports.py, there is a little typo - if none of the cases match, it tries to create `ImportException` (which does not exist) instead of `ImportError`.


Thrameos:
Drat, I missed that one.   Thanks for the PR.   I will release it with the next patch release.

PR #799, opened 2020-07-18
Port to android

Thrameos:
This is the port of JPype to Android for Kivy.  It requires the corresponding recipe file for pythonforandroid to use.   This PR must be included after backport PR as it is constructed on the features of backport.

The porting process consists of 
- Disabling the private module calls to unsafe operations.
- Defining an entry point for connecting to the DVM and a start up hook in `jpype/__init__.py`
- Documenting the differences in the DVM

The port currently has the isPackage functionality disabled but it otherwise operations.  As conditional compilation is not supported in Python and Java, the recipe file makes the remainder to the changes in the form of a patch file.   

Bugs fixes
-  JPype was calling the wrong method for longValue() on boxed types.  Though this was accepted by Oracle JVM it was in error and has been corrected.  This required moving the entry points back to the boxed type C++ wrappers.
- Many warnings generated by clang were addressed.  Added virtual and overloaded as required to C++ declarations.
- A name conflict with overloaded methods with Array constructors was resolved.
- Setup script updates to allow the ``--android`` to the build_ext

There are no user level viewable changes in this PR.


Thrameos:
``buildozer android clean`` definitely cleans way too much stuff.  Is there a way to clean just the jpype recipe stage?

Thrameos:
Thoughts on how to get this into CI when all this is complete.

- The full buildozer process is too long to run in the azure pipeline as part of CI so we are going to have to split it.
- We can run a separate pipeline that runs buildozer once, cleans the stage for JPype, and pushes the result into an artifact.
- When we run the CI pattern we will need to pull the build artifact, install buildozer, and then run just the JPype stage.
- We will need to find some way of running the test inside of an emulator on azure.

That should give us confidence that we have a working system.  Though I may be getting ahead of myself.

Thrameos:
So tasks remaining

- [X] Set up the build system (SDK and NDK)
- [X] Set up a recipe
- [X] Get a basic build pattern to launch from the command line
- [X] Figure out how to build in place rather than pulling from the remote archive.
- [X] Fix all the warnings
- [X] Figure out a symbol for conditional compiling (ANDROID).
- [X] Pass ``--android`` to setup.py for build and install.
- [ ] Figure out how to unpack an apk to see what stuff has been included.
- [ ] Backport the thunk free method of loading from epypj.
- [ ] Replace the start JVM and shutdown using a patch.
- [ ] Replace the thunk loader.
- [ ] Set up an emulator so we can test code.
- [ ] Set up a basic start script that launches and interactive shell.
- [ ] Validate that private module ``_jpype`` works.
- [ ] Validate main entry points for ``jpype``.
- [ ] Test functionality and correct issues with ``org.jpype`` for differences between Oracle and Android JVM.
- [ ] Figure out how to mount the test directory from the emulator.
- [ ] Get the JPype test bench to run completely.
- [ ] Define a CI mechanism so we can test this in CI and detect future breakage.
- [ ] Document the required process to use.
- [ ] Get kivy folks to contribute to the review process.

tshirtman:
apks are just a zip file, you can extract them as one, and inside there is a `private.mp3` file which is not actually an mp3, but also an archive (tar.gz iirc), that contains the python distribution and application, this mp3 shenanigans is just a way to work around some arbitrary limitations of the android sdk (that may have been lifted since, but we didn't try doing away with it, as it works).

Thrameos:
@marscher What is your opinion on starting to back out some of the C++ code in common?   Currently we use C++ in three ways.  We redirect all exceptions into C++, we used C++ style functions and classes, and we use portions of C++ std library.  

The exceptions is likely a bad idea because C++ unwinding of exceptions is pretty slow. I having clocked it in a while but it was 100 times slower than a function call last I checked.   C++ exceptions are designed to be exactly that "exceptional", but Python exceptions are intended to be routine.   Unfortunately this design choice is hard to remove.

The C++ functions and classes do not present a problem.

C++ std library is somewhat of problem because it in not available routinely on android, so we are adding another dependency just to get stringstream, string, vector, and list.   Though I know reuse is best over reimplementing, switching those components would save us.  I could make the switch conditional so we only use the substitution on android, but then it won't get routine testing.

As it is, I went through the current usage and found that most of it was old or cruft.   We called stringstream where we could have called PyErr_ functions, including map and other headers we no longer used, and  our use of cout and cerr was just to push out simple strings and not do anything that we couldn't to as well and just as fast using libc.

marscher:
I don't think that we should drop using c++ standard library. It also supported for recent Android versions (lollipop) according to https://developer.android.com/ndk/guides/cpp-support
Why should we spent lots of effort to support something old? 
I agree about the (mis-)usage of cpp exceptions, but if it is hard to remove, I'd go for more low hanging fruits.

Thrameos:
Sounds good to me.   I finished the port of JPype on to Android last night after I got C++ linkage working.   Some of the more advanced features are not workable so I will have to figure out how to split the JPype core up a bit.  A package option to setup.py will need to exclude things that lack support on the android platform.  I think for the most part I can do this by moving functions in the module into specific files that we then delete or convert those files to stubs in packaging stage.   Things like starting and shutting down the JVM, the import hooks, pickle, and some of the doc hooks will need to be cut out.

Thrameos:
@tshirtman Okay this code is ready to start review.   Please look over the changes carefully and ask questions about anything that looks out of the ordinary (bad formatting, stray debug print, undocumented changes to the user api in jpype, incomplete or incorrect documentation).  1200+ line PRs take a while to review so anything you catch will speed up how quickly it gets included.

setupext/platform.py
```diff
@@ -20,79 +20,102 @@
...
-    platform_specific['include_dirs'] += [os.path.join(java_home, 'include')]
-
-    # check if jni.h can be found
-    for d in platform_specific['include_dirs']:
-        if os.path.exists(os.path.join(d, 'jni.h')):
-            print("Found native jni.h at %s" % d)
-            found_jni = True
-            break
+
+def Platform(include_dirs=[], sources=[]):
```
marscher, line 29: I vote against monkey-patching sys.platform and add an additional argument to this function, so we can pass the platform name from setup.py and just change it case the android argv is present. Sorry for being pedantic about this, but I've experienced horrible side-effects in the past, when changing such things (especially with distutils).

Thrameos: Okay fine with me.   Not sure why the cross compiler doesn't have any symbols that tell you which you are compiling but I didn't find any.

marscher (approved):

tshirtman:
Hey, sorry for going dark, i had started a review, but got quite busy with work, so couldn't get back to it (and i must say i had a hard time building an understanding of the change without understanding the whole project), anyway glad this got merged.

Now i wonder how you built jpype on p4a? it seems adding the master.zip from github as a requirement is not enough, as it fails to compile for me, complaining about missing <map> header, i tried to activate c++_shared in Android.mk as it seem to be the normal solution for missing stl problems on android, but that didn't get ride of the error.

I see there is the --android flag but i'm not sure how you pass it through buildozer/python-for-android and from the diff, i don't think it changes anything about the usage of <map> which would indeed be quite surprising), or about included headers.

Thanks again.

Thrameos:
The completed demo with the recipes is located at https://github.com/Thrameos/kivy-remote-shell 

I sent some instructions on discord regarding which branch and setup was required.   But those have been lost to time.  I believe is was pretty obvious if you look as the revision history.

I don't recall any issues with the C++ headers when I did the compiling.   The specific build that I did was "android" from jpype using the kivy-remote-shell, and the p4a modifications on Thrameos.  I have not testing with master on jpype because we it wasn't merged, that means there may be some drift that needs to be resolved.

Some of the modifications are in the jpype project but there were additional files that needed patching.   I tried to keep all the files on the main jpype line, but certain files need to be removed due to JVM differences that required additional patching.   I figured once you had a chance to review the changes and pull the recipe into the main line of p4a then it would be easier to maintain. 

My development machine is still in transition, but if you need some assistance I am be able to find the old setup.   I had a backup on my gaming machine that I believe still runs.  If you have a specific time that you would like to work on it please poke me so that I have some time to get the setup running again.

PR #803, opened 2020-07-19
fix typo in import system: "jname" -> "name"

michi42:
In the import system, if a TLD is not registered (and therefore the check in line 163 fails), when trying to create the parent module in it is tried to access an undefined variable "jname" instead of using "name".


Thrameos:
Thanks for the contribution.

PR #805, opened 2020-07-20
Improvements to speed for primitive arrays

### LINKED ISSUE #804: Feature request optimize numpy array transfer
The benchmarking of transfer of large arrays shows that there is still room for improvement for transfer of numpy arrays.   It should be going through a direct memory buffer transfer, but it may be that calling a conversion function each time is slowing things down.   

Recommendation:
- [x] Identify what precise path numpy arrays are taking in the code.
- [x] Analyze the source of the slowdown.
- [ ] If it is the converter functions, then add a start1, start2, len parameter to converters so they can be used to do bulk transfer rather than just one off transfers.

Thrameos:
Identified one of the old slow paths in argument passing that we have not yet addressed and transfers it to the optimized paths.

Fixes #804

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -6,6 +6,12 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 Latest Changes:
 - **1.0.2_dev0 - 2020-07-16**
 
+  - Improved speed on transfer of lists, tuples, buffers to arrays of Java
+    primitives by a factor of 4 to 100 depending on the data type.  The
+    conversion uses optimized path for memory buffers, rather than the 
+    Sequence API.  When a Python buffer is encountered only the
+    first element is checked for conversion as Python buffers are homogeneous. 
+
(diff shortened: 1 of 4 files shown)


Thrameos:
@marscher so do shall we cut a 1.0.2 at this point?   Thus far I haven't seen anything that would break the API, but we do have a few whoops and some actual improvements that can go out the door.

marscher:
Sounds good to me. Please go ahead.

On July 21, 2020 3:08:04 PM UTC, Karl Nelson <notifications@github.com> wrote:
>@marscher so do shall we cut a 1.0.2 at this point?   Thus far I
>haven't seen anything that would break the API, but we do have a few
>whoops and some actual improvements that can go out the door.
>
>-- 
>https://github.com/jpype-project/jpype/pull/805#issuecomment-661918492

-- 

fleiber:
Great news! We had been fighting with some small calls with int[] argument which should have been almost free but were quite slow, so this is a great improvement for us, thanks :)

PR #809, opened 2020-07-22
Support of switchable interactive behavior.

### LINKED ISSUE #778: ^C on non interactive sessions
We added a signal handler so that we don't crash out of interactive sessions in Python from the Java handler, but when running a non-interactive session that may be the proper behavior.  Often when running a big script a ^C means stop the process and thus letting Java ignore it so that we can fall back to an interactive shell doesn't make sense.   Should we add an option to startJVM() which enables or disables signal behavior.   If not set it would autodetect based on isatty() and ``sys.ps1``.   

Something like 
- jpype.startJVM( interrupt=True) - ^C stops process
- jpype.startJVM( interrupt=False) - ^C breaks Java if possible and returns control to Python
- jpype.startJVM() - ^C selected interrupt if not interactive shell.

Does this make sense?

Thrameos:
This will need to wait until 1.1 as it introduces a new switch.

Fixes #778


Thrameos:
This one still has the problem that on rare occasions a ^C can come at a bad time such the JVM shuts down will Python is making a call resulting in a segmentation fault.   But given the other case is having no way to get out of a long computation loop I am not sure what the alternative should be.

marscher:
Agreed, that this solution is better than not handling the interactive situation at all.

Thrameos:
This one is going to take a while to rebase as there is a conflict with the arguments to the Context start.

marscher:
Warning, treated as error:
/home/vsts/work/1/s/doc/CHANGELOG.rst:8:Unexpected indentation.

marscher:
native/java/org/jpype/JPypeContext.java:117: error: package instance does not exist
    instance.typeManager.init();
            ^
native/java/org/jpype/JPypeContext.java:118: error: package instance does not exist
    instance.referenceQueue.start();

Thrameos:
@marscher I am trying to get some of these pull requests through so I can release 1.1.   My development machine is still in scrambles because my work load picked up and thus did not have time to continue bugging support finish.   

For now I have the old lap top running without a battery available and cloned the files onto my new machine in the mean time.   They wanted the old machine back within a week of sending my new machine, but as they haven't finished the new one and have forgotten about reclaiming the old, I have the old one available as a back up.   Unfortunately, if they can't fix my new laptop they plan to reformat it which means that my advanced work on the reverse bridge code is still stalled.  

Somehow the new laptop also triggered a security review for some of the software I use in development which unfortunately given the glacial rate at which my organization works means that they are still in the process of working on the paperwork and until I get permission I am precluded from having the code install and connecting to my office network which means a lot of installing, doing work, uninstalling, and then connecting. (rinse, lather, repeat)  I know that this doesn't seem like a reasonable amount of time to resolve issues, but unfortunately it is the nature of my work where security can trump productivity.

doc/CHANGELOG.rst
```diff
@@ -4,9 +4,14 @@ Changelog
 This changelog *only* contains changes from the *first* pypi release (0.5.4.3) onwards.
 
 Latest Changes:
-
 - **1.0.3_dev0 - unreleased**
 
+  - ^C behavior is switchable with interrupt flag to startJVM.
```
marscher, line 9: I'd write "strg+c" as it is more search engine friendly :)

marscher (approved):

marscher:
Thanks a lot!

Issue #714, opened 2020-04-27
automatic type stub (pyi) generation for java classes


michi42:
I'm working on making a rather big, domain specific java API available through JPype. One of our main issues is the complete lack of type information during static analysis - and therefore no linting or IDE auto-completion.

PEP 484 and 561 support providing type information in separate .pyi "stub" files. The mypy project provides stubgen, a tool to automatically create these from python or C modules.
Since Java is strongly typed, all the information necessary to generate these kind of stub files is in principle available in the class files and can be obtained e.g. through reflection.

I implemented a first version: https://gist.github.com/michi42/2110a8615a0e917a13ec6748c6168735
The module is supposed to be called from a simple script that sets up the JVM, e.g.
```python
import jpype
import stubgenj

# fire up the JVM with the necessary JARs

stubgenj.generate_java_stubs(["java", "cern.lsa"], 'pyi')
```
This is still very limited, just a starting point that handles the "most common" cases. In particular, it does not support Generics (should be translated to python type vars), Constructors (should be translated to __init__ methods) yet, and the types of constants and enum constants are not yet detected.

Thrameos:
Interesting.  I look forward to some additional information.   

Can this be connected to the existing annotation API that we are using?  We currently support Jedi auto-completion which allows linting of returns types if the overload is not mixed.   I haven't tried with Kite.

Here is my recommendation as an implementation.   Just like Java docs we can pick up Java information though the resource API.   Therefore, if you run a preprocessor that converts the Java jar file (``org-mystuff.jar``) into a second jar with the needed annotation ``org-mystuff-py.jar``.  Then when I load a class using ``org.mystuff.MyClass.class``  I can also fetch the ``org.mystuff.MyClass.annot``  file which can then provide stubs.

michi42:
Hi,

Thanks a lot the quick response!

Indeed, it goes into the same direction as the annotation API (and __docs__) that JPype is already populating at run-time (which is great for interactive applications like IPython or Jupyter), however it provides static information that is pre-generated in once and can then be used without executing any of the code (as it is typically the case for IDEs like PyCharm).

Also, this supports overloads, even when the signatures are completely different and return types depend on the signatures, through the overload decorator: https://docs.python.org/3/library/typing.html#typing.overload - also it should be possible to represent generics and type arguments, although my first version does not handle that. I frankly don't know if the annotation API supports all of these.

[...]

PR #815, opened 2020-07-26
Correct bugs in hints system

Thrameos:
Deals with bugs reported in #714.  Throwable had the wrong wrapper type in the class factor.  Adds automatic conversion for List and Iterable, though they may be costly.   This is probably safe for 1.0.2 but if you want me to push it to 1.1 then just set the milestone on the review.

### DIFF
--- a/jpype/protocol.py
+++ b/jpype/protocol.py
@@ -67,8 +67,9 @@ def _JFileConvert(jcls, obj):
     return jcls(obj.__fspath__())
 
 
-@_jcustomizer.JConversion("java.util.Collection", instanceof=Sequence,
-                          excludes=str)
+@_jcustomizer.JConversion("java.util.Iterable", instanceof=Sequence, excludes=str)
+@_jcustomizer.JConversion("java.util.List", instanceof=Sequence, excludes=str)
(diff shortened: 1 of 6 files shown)


Thrameos:
@marcher I removed the one problematic converter (though long term we may want to add it back).   The rest seems small enough to be a bug fix.   As this is a minor API with no user impact, I will merge and push in the 1.0.2 if you don't have time to review.

marscher:
Thanks for quickly fixing this.
23h hours to review? Was it that urgent/had security implications?
At least a change log should contain a notice about the fixed bug (which had user implications as far as I noticed).

Thrameos:
Sorry I was going to cut 1.0.2 and wanted to get it included.  It was unfortunately a serious error as although there are not many interfaces that take Throwable it would have given some very strange results as it would have produced an Object rather than a Throwable.  I will be sure to add a note in the log.

Thrameos:
Again sorry if it appeared that I was rushing you on the review.  I was behind in making the release so I was trying to get it going during my window during the morning group check in call, unfortunately I was running late.  I will need to do another patch release in another two weeks if more issues turn up.

PR #820, opened 2020-07-27
Cycle

Thrameos:
PR to pull the release number changes back into the master.


Thrameos:
@marscher I didn't hear back from you on the release, but I am on the road all day so I wanted to get this out the door so it doesn't bug me on my vacation day.  I removed the one API addition as it didn't fit with a patch release.  I guess the next items should be pushed to a 1.1 release.

The one API change that we did hit is on the policy regarding conversions.   As seen in #821 some people are expecting conversions to concrete types which I would consider a mistake.  JPype should convert (or better yet provide a Java front end for) interface types, but to actually hit a specific concrete type they should explicitly create a wrapper.   Example

- ``void dosomething(List<String> s)`` should take a Python List.
- ``void dosomething(ArrayList<String> s)`` should require ``dosomething(ArrayList(list))`` because we do not guarantee the type of the conversion to be any specific implementation beyond implementing list.

We currently have one test that fails this in that it expected String[] to match better than List<>.   It seems that it is not clear which is a better match given our current system.  There are not likely to  be many cases where this sort of overload hit is a problem, but I can see ``do(Object ... a)`` and ``do(List a)`` as being reasonable overloads.  I would guess that I would favor it matching the later rather than the former, but we do not currently have a preference to concrete parameter lists over variodic ones.

Second, as discussed on the hints API, our current system does matching without regard to the generic type, because our type system has erasure.  Is it better to force a match of items when deciding the overload or when doing the conversion?  For speed reasons I think conversion is best.  In other words don't try to force items into List<String> as String until after we have decided the overload (as the cost of traversing is high), but when we convert give an error before we make the call.  Currently we are converting to objects and then Java has to catch the fail.

PR #823, opened 2020-08-01
Rework JPackage so that packages can be Python modules

Thrameos:
There are times were it may be required to add functionality to a JPackage.  This enhancement makes Java packages into formal Python modules so that that are no long distinguishable from Python.   A followup PR will add support for loading Python code sections from a Jar file.   

Though this does not directly depend on anything it should likely be added after cycle and backport.  It targets 1.1 release.


native/python/pyjp_package.cpp
```diff
@@ -131,134 +136,116 @@ static PyObject *PyJPPackage_getattro(PyJPPackage *self, PyObject *attr)
...
-		{
-			JPPyObject u = JPPyObject::call(PyUnicode_FromFormat("%s.%U",
-					self->m_Package->m_Name.c_str(), attr));
-			JPPyObject args = JPPyObject::call(PyTuple_Pack(1, u.get()));
-			out = JPPyObject::call(PyObject_Call((PyObject*) PyJPPackage_Type, args.get(), NULL));
-		} else
+		PyErr_Format(PyExc_RuntimeError,
+				"Unable to import '%s.%U' without JVM",
+				PyModule_GetName(self), attr);
+		return 0;
```
marscher, line 144: Should it return a null pointer or Py_NONE? I'm not sure what happens to the callee of this function, if it retrieves a null pointer or will the set Python exception hit first?
At least we should return NULL and not mix it with plain integers, IMO.

native/python/pyjp_package.cpp
```diff
@@ -131,134 +136,116 @@ static PyObject *PyJPPackage_getattro(PyJPPackage *self, PyObject *attr)
...
+	}
+	JPJavaFrame frame = JPJavaFrame::outer(context);
+	jobject pkg = getPackage(frame, self);
+	if (pkg == NULL)
+		return NULL;
```
marscher, line 149: meaning mixing NULL and 0 (it would be easier to refactor later e.g. to nullptr).

marscher (approved):
I think this is a much better way of handling packages. Thanks a lot!

native/python/pyjp_package.cpp
Thrameos: I should comment this section.  If the pkg is null then an exception has already been set.

native/python/pyjp_package.cpp
Thrameos: Quite right it should have returned NULL.   Python expects a null pointer to be returned if ever the exception is set.   It least in a few places if the exception is set Python may fail to clean up the reference.   It is considered a SystemError to return anything but NULL when returning an exception.   

I will get that bad return addressed.

PR #824, opened 2020-08-01
Added explicit version warning

Thrameos:
This PR makes it so the use of Python 2 will produce a warning (and hopefully end those "why does this not compile with Python 2" error reports).

This should be safe for a patch release.


marscher (approved):
LGTM. I felt free to change the order of imports, so we only need to guard the version once.

setup.py
```diff
@@ -18,12 +18,17 @@
...
 # *****************************************************************************
 import sys
-import setupext
 from pathlib import Path
 from setuptools import setup
 from setuptools import Extension
 import glob
 
+
+if sys.version_info[0] < 3 and sys.version_info[1] < 5:
```
usmonster, line 27:
```suggestion
if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and sys.version_info[1] < 5):
```

PR #826, opened 2020-08-01
Fix referencing bug

### LINKED ISSUE #822: Python3.8-dbg negative reference
Testing with Python 3.8 debug version indicates that at least one test is hitting a negative reference count.  Though there are no failures shown in the normal run, this should be run to ground and fixed.   Guess this is another patch release fix.

Thrameos:
I identified the issue.  The reference counter was not being incremented on the stack trace creation.   Should be safe for a patch release.

Fixes #822


marscher (approved):
LGTM. Thank you!

PR #828, opened 2020-08-03
Pyc

Thrameos:
This pull request is build on backport and pkg.  With this python packages can be compiled into Java jars and imported in Python.   This will allow the customizers to be stored with the Java classes that need to be customized.  This allows for a scaling from our current model where all the customizers need to be defined before the class is encountered.

### DIFF
--- a/native/common/include/jp_class.h
+++ b/native/common/include/jp_class.h
@@ -54,14 +54,13 @@ public:
 			JPMethodDispatchList& methods,
 			JPFieldList& fields);
 
-	string toString() const;
-
 	string getCanonicalName() const
 	{
 		return m_CanonicalName;
 	}
 
-	string getName() const;
+	string getName(JPJavaFrame& frame) const;
+	jstring getPackageName(JPJavaFrame &frame) const;
 
 	bool isAbstract() const
 	{
--- a/native/common/include/jp_context.h
+++ b/native/common/include/jp_context.h
@@ -226,6 +226,7 @@ private:
 	jmethodID m_Object_HashCodeID;
 	jmethodID m_CallMethodID;
 	jmethodID m_Class_GetNameID;
+	jmethodID m_Class_GetPackageNameID;
 	jmethodID m_Context_collectRectangularID;
 	jmethodID m_Context_assembleID;
 	jmethodID m_String_ToCharArrayID;
@@ -249,6 +250,7 @@ private:
 	jmethodID m_Context_GetPackageID;
 	jmethodID m_Package_GetObjectID;
 	jmethodID m_Package_GetContentsID;
+	jmethodID m_Package_GetImplementationID;
 	jmethodID m_Context_NewWrapperID;
 public:
 	jmethodID m_Context_GetStackFrameID;
--- a/native/common/include/jp_javaframe.h
+++ b/native/common/include/jp_javaframe.h
@@ -396,7 +396,7 @@ public:
 	jobject getPackage(const string& str);
 	jobject getPackageObject(jobject pkg, const string& str);
 	jarray getPackageContents(jobject pkg);
-	jbyteArray getPackageImplementation(jobject pkg);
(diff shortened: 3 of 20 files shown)


marscher:
@Thrameos this one needs to be rebased on master. Thanks!

marscher:
Thanks for resolving the conflicts. There is one test case failing
```
self = <test_reflect.ReflectCase testMethod=testAnnotation>

    def testAnnotation(self):
        method = self.Reflect.class_.getMethod('annotatedMethod')
>       annotation = method.getAnnotation(self.Annotation)
E       AttributeError: Java package 'com.sun.proxy' is not valid
```
Does this mean that the installed JDK11 doesn't have this Oracle/Sun specific package anymore?

Thrameos:
This error is not clear.   I will look into it.

Thrameos:
I investigated the issue.  There are two types of errors introduced both as side effects of the change.  Part of this PR made the Java packages much more similar to Python modules in terms of reported attributes.   This is confusing third party code into treating Java packages in ways that we have not exercised before.   Java classes now report Python module names and paths even if the jpype.imports are not included.   This is causing codes which have no error handling for the case that the package name is not available to fail.   This causes jedi and annotation code to fail on the com.sun.proxy stuff.   I would have to get into it further to see which exactly is triggering this.   

Not clear what the correct solution is.   I could try to disable the package name code which would make JPype modules less compatible with Python modules (which was the prior state) or I can press forward and try to make them more compatible.   Unfortunately, as this is third party error checking issues, that means I would most likely have to actually force Java modules into the Python module system entirely which would break jpype.imports as an optional package.    

Still more work to determine if it is the path or the name piece that is making code have issues.   But as I can replicate it, I should be able to figure out with a few experiments.

marscher:
This one is still milestoned for 1.1, do you think we should try to include it for 1.2 now @Thrameos?

Thrameos:
The basic issue with this is that Python does not have a stable pyc format.   As I discovered debugging the Python 3.9 windows issue, the compiled format of Python is completely non-portable between versions.   It was not even possible to unmarshal from either going forward or backwards between versions.

I would really like to be able to include this sort of capability in our distribution as it would greatly simplify distribution of customizers.   But it seems like there needs to be some fundamental improvements in the Python infrastructure to make this happen.   Java solved the distribution of compiled files that move forward in time long ago and even through there were many warts in their initial implementation (like long/double being double slots, const tables that can't store certain necessary types, and failure to place sizes so you can skip forward), it is still a pretty decent standard to move forward on.   

The funny thing is when I disassembled code from multiple Python versions the contents looked identical so this is mostly just the fact that the marshalling protocol is unstable rather than the need for compiling Python files differently.  There are sometime new "fast" calls or other enhancements but those sort of things still run as byte codes don't appear to be removed.  

It really seems like Python is failing on this aspect of virtual machine design.  As it stands I would need to use a mechanism like multi release jar so that one could include many Python code implementation in one jar which is why this one has been festering.  Thus I am going to have to say we should kick this down the road unless we can come up with a version free way of marshalling Python code elements.

marscher:
Thanks for your detailed explanation on this. I do not see, why shipping a Jar and a Python module (e.g. within one archive) would be so painful anyways.
If you really like to invest more time in this feature, I'd suggest to give ASTs (and their serialization) a try.

Thrameos:
The issue that I would ultimately like to solve is making it so that Python can understand a Jar (or tar) as a file system (without necessarily having the JVM running).   It would be very nice for my customers to just ship them the jar only with the customizers embedded in the jar rather than sending them the jar file and then another Python module to make use of it. 

Currently as JPype is just supporting the Python to Java bridge, it really isn't much of an issue as you use Pypi or conda to distribute Python packages and we can just ship the jar with the package.  But once we have a Java to Python bridge we also need to consider Maven artifact style distributions.   Of course that will open a whole other can of worms.   For pure implementations like Jython this isn't an issue, but mixed systems like we are doing is going to be much more of a pain.

PR #830, opened 2020-08-04
Improved error message for static

### LINKED ISSUE #829: Improve error message for static mismatch
One common error when using JPype is when a member method is called statically.  The error message that this gives is somewhat decieving.

```
class Foo{
   public bar(int i);
}
```

When called as ``Foo.bar(1)`` will give

```
TypeError: No matching overloads found for Foo.bar(int), options are:
        public Foo(int)
```

The message is correct.  You didn't call bar with a object and the only one that matches is not static.   To improve the error message I recommend we check the types of the first argument of the method and the one provided to make the output read as 

```
TypeError: No matching overloads found for *static* Foo.bar(int), options are:
        public Foo(int)
```

It can still be fooled by methods that have this and the first argument the same but it should cut down on the number of confused users a bunch.

Thrameos:
This PR attempts to clue the user into why a method called as a static does not fit to overloaded methods.  This often seems to be a point of confusion.

Fixes #829

```
No matching overloads found for *static* java.lang.String.substring(int), options are:

No matching overloads found for java.lang.String.substring(float), options are:
```


marscher (approved):
LGTM! Thanks for making the error message more clearer.

PR #833, opened 2020-08-04
Support slice operator in JString to match Python str

### LINKED ISSUE #831: JString slice operator
I did a deep scan of the github forums.   It appears the most common problem with convert strings (other than we gave a deprecation warning for a year) is the lack of support for slicing on JString.   It isn't too much to add and will save some of those older libraries that have not yet switched to setting a conversion policy.

Thrameos:
This addresses a common incompatiblity between Java string and Python str types by adding a slice functionality to ``__getitem__``.  
  
Fixes #831


marscher (approved):
Very well. That was missing for a long time. Thanks for adding it.

PR #840, opened 2020-08-10
Exposed the classloader object used by JPype into python.

KOLANICH:
```python
In [1]: %paste
import _jpype
import jpype
jpype.startJVM("-ea", jvmpath=jpype.getDefaultJVMPath(), classpath=[], convertStrings=True, ignoreUnrecognized=False)

jpype._classpath.addClassPath("./test/classes/test.jar")
NumericTest = jpype.JClass("jpype.numeric.NumericTest")
## -- End pasted text --

In [2]: NumericTest
Out[2]: <java class 'jpype.numeric.NumericTest'>
```

### DIFF
--- a/jpype/_classpath.py
+++ b/jpype/_classpath.py
@@ -22,6 +22,8 @@ __all__ = ['addClassPath', 'getClassPath']
 _CLASSPATHS = []
 _SEP = _os.path.pathsep
 
+_ourClassLoader = None
+File = None
 
 def addClassPath(path1):
     """ Add a path to the Java class path
@@ -49,6 +51,8 @@ def addClassPath(path1):
         path1 = path2.joinpath(path1)
 
     _CLASSPATHS.append(path1)
+    if _ourClassLoader:
+        _ourClassLoader.addFile(File(str(path1)).toPath())
 
 
 def getClassPath(env=True):
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -312,6 +312,9 @@ def initializeResources():
     _jpype.JClass('org.jpype.JPypeKeywords').setKeywords(
         list(_pykeywords._KEYWORDS))
 
+    _classpath._ourClassLoader = _jpype._getClassLoader()
+    _classpath.File = _jpype.JClass("java.io.File")
+
     # Everything successed so started is now true.
     _JVM_started = True
 
@@ -327,6 +330,7 @@ def shutdownJVM():
     if threading.current_thread() is not threading.main_thread():
         raise RuntimeError("Shutdown must be called from main thread")
     _jpype.shutdown()
+    _classpath._ourClassLoader = None
 
 
 # In order to shutdown cleanly we need the reference queue stopped
--- a/native/common/jp_typemanager.cpp
+++ b/native/common/jp_typemanager.cpp
@@ -23,7 +23,7 @@ JPTypeManager::JPTypeManager(JPJavaFrame& frame)
 
 	jclass cls = m_Context->getClassLoader()->findClass(frame, "org.jpype.manager.TypeManager");
 	m_FindClass = frame.GetMethodID(cls, "findClass", "(Ljava/lang/Class;)J");
-	m_FindClassByName = frame.GetMethodID(cls, "findClassByName", "(Ljava/lang/String;)J");
+	m_FindClassByName = frame.GetMethodID(cls, "findClassByName", "(Ljava/lang/String;Ljava/lang/ClassLoader;)J");
 	m_FindClassForObject = frame.GetMethodID(cls, "findClassForObject", "(Ljava/lang/Object;)J");
 	m_PopulateMethod = frame.GetMethodID(cls, "populateMethod", "(JLjava/lang/reflect/Executable;)V");
 	m_PopulateMembers = frame.GetMethodID(cls, "populateMembers", "(Ljava/lang/Class;)V");
@@ -50,9 +50,10 @@ JPClass* JPTypeManager::findClassByName(const string& name)
 {
 	JP_TRACE_IN("JPTypeManager::findClassByName");
 	JPJavaFrame frame = JPJavaFrame::outer(m_Context);
-	jvalue val;
-	val.l = (jobject) frame.fromStringUTF8(name);
-	JPClass* out = (JPClass*) (frame.CallLongMethodA(m_JavaTypeManager.get(), m_FindClassByName, &val));
+	jvalue val[2];
+	val[0].l = (jobject) frame.fromStringUTF8(name);
+	val[1].l = (jobject) m_Context->getClassLoader()->getBootLoader();
+	JPClass* out = (JPClass*) (frame.CallLongMethodA(m_JavaTypeManager.get(), m_FindClassByName, val));
 	if (out == NULL)
 	{
 		stringstream err;
--- a/native/java/org/jpype/manager/TypeManager.java
+++ b/native/java/org/jpype/manager/TypeManager.java
@@ -170,15 +170,15 @@ public class TypeManager
    * @param name is the class name.
    * @return the C++ portion.
    */
-  public long findClassByName(String name)
+  public long findClassByName(String name, ClassLoader loader)
   {
-    Class<?> cls = lookupByName(name);
+    Class<?> cls = lookupByName(name, loader);
     if (cls == null)
       return 0;
     return this.findClass(cls);
   }
 
-  public Class<?> lookupByName(String name)
+  public Class<?> lookupByName(String name, ClassLoader loader)
   {
     // Handle arrays
     if (name.endsWith("[]"))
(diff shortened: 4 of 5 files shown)


Thrameos:
Hmm.  Will need to do more testing then.  The long term plan on this is to make it so you can call addClassPath after the JVM is started (assuming it works reliably) and it will add it to the DynamicClassLoader.

KOLANICH:
It seems the stuff isloaded by different loaders and the one I expose to python is not the one used:

```io.kaitai.struct.precompile.ErrorInInput: io.kaitai.struct.precompile.ErrorInInput: nt_mdt: /meta/imports/0: class scala.collection.convert.Wrappers$IterableWrapper cannot be cast to class java.lang.String (scala.collection.convert.Wrappers$IterableWrapper is in unnamed module of loader org.jpype.classloader.DynamicClassLoader @5ca881b5; java.lang.String is in module java.base of loader 'bootstrap')```

KOLANICH:
Tried to insert `Thread.currentThread().setContextClassLoader(this.classLoader);` into various places.

0. tried to execute from python - no positive effect.
1. In JPypeContext ctor - the same
2. Inserted into `JPContext::initializeResources` -> very weird error about string out of index access in an unknown location. The error doesn't depend on if I insert these calls close to the beginning of that method or near its end.

KOLANICH:
Implemented very dumb mode of runtime loading (without possibility of unloading) but haven't solved the issues with loaders conflicts yet :(

Thrameos:
This one still needs some work, so I am going to push it past the 1.1 release.

Thrameos:
@KOLANICH I think you may be over engineering this one.   Please look over #888 and see if it does what you want.   

My general goal is to minimize the number of things in the C layer if I can.   I think this one can be done simply by talking directly the Java context and thus skipping the need for additional C layer hooks.  

Other options we can do if it turns out this still leads to permission issues on the loader is we can add the property "java.system.class.loader" when we start the JVM.  This would ensure everything is forced to use the common classloader (though we would need to create a default loader underneath it to act as a back stop.

Thrameos:
I should point out that unloading does not appear to be possible for oracle and open JDK versions.   Unfortunately they add a hidden link to all classes that are loaded.  I have tried a few times to make it so that I can access that list and clear it but it appears to be superglued in place.  

If you do need to be able to "unload" a class you have to do the following. 
 - Create a URLClassLoader with all the jar that are needed.
 - Use Class.forname or JClass with the classloader option to manually load all classes.
 - When you are done dispose of the JClass instances and URLClassLoader

Note this only gives the possibility of a class getting garbage collected, because the JIT and other units in Java may have held the class.   I am not sure why Java has this restriction.  It appears to be a laziness in the JVM to working through the issues of unloading a class.

Please note that if any Jar files have security constraints and were loaded previously from the underlaying System classloader then the URLClassLoader may fail or the class will generate a security exception when passed (or a class cast exception).   So you must load every class with same classloader or it will treat the others as foreign.

KOLANICH:
I have tested #888. It works fine for the test in the top post, and also with antlr4 + and mdktCompiler. The issue with conflict of classloaders for when used with Kaitai Struct is still unsolved, but it should probably a part of another PR. Thank you.

Thrameos:
Can you post a short test that exercises the Kaitai classloader?   The one likely solution that I have for that is to set the system class loader by properties.    The other option would be to add a module exception to the start flags for the JVM.   

I am really confused by the error message that you are showing.  String is a final class in Java so there is no way ``scala.collection.convert.Wrappers$IterableWrapper`` could ever pass that cast unless someone did some hack magic on the backend which would certainly violate certain many assumptions in the JIT compiler (final methods are final and should have no virtual table entry as you can look them up directly.)

KOLANICH:
>If you do need to be able to "unload" a class

I guess possibility to unload a package is not so important compared to classloaders conflicts, since classloaders conflicts make apps and libs unusable, but inability to unload just causes a bit larger memory consumption. Though being able to reload a package, I mean to load a new version of it, maybe even compiled in runtime, is important.

>Other options we can do if it turns out this still leads to permission issues on the loader is we can add the property "java.system.class.loader" when we start the JVM.

Gonna try tomorow, but I am not sure that that will work, since the error message mentions that `java.lang.String` is loaded with 'bootstrap'.

Thrameos:
The URLLoader trick can at times be used to load more than one version of a library, but it is not a very stable solution because any libraries from different classloaders will not recognize each other nor coexist nicely was they appear to have the same names leading to conflicts.

The one way the I have dealt with having many versions of the same library in memory is though the use of ASM.  One can use jarjar (which is unfortunately falling out of support) to rename all the symbols in a library while the jar is on disk.   I often use this in debugging when I have to do comparisons of how to major revisions of a Jar react to the same inputs.

You can do the same thing using ASM combined with classloader to rename a Jar on the fly.   "org.mypkg." => "org.mypkg.v7."  but it does take some skill with ASM visitor patterns to traverse the jar file, class file, fields and methods and rename all the symbols as you go.   We could add it as an option to the Dynamic loader though it will take a bit of work on your part to make it happen.   This is also how Jacoco and other instrumentation libraries work.

I am doing a lot of that magic already in the extension and epypj where I need to make new classes on the fly and load them.  If and when I want try to bite off the cycle issue with Java and Python I will likely need to create a Java loader agent that adds a "traverse" method to all Java classes (including object) so that Python can traverse through Java to find Python linked objects for GC.

KOLANICH:
>Can you post a short test that exercises the Kaitai classloader?

Gonna try to craft it tomorrow.

>The other option would be to add a module exception to the start flags for the JVM.

Making assumptions about certain packages when starting a JVM defeats the purpose of dynamic loading. Since we cannot restart a JVM (and even if we could, it can have disruptive side-effects) and probably cannot change the most of settings controled by flags in runtime, one of the goals of dynamic loading is to allow the first python package using a jar to start JVM, and then allow other python packages using jars to be used.

Thrameos:
I certainly agree with the desire to start once with the same options.  Though I will point of that in Matlab they start with a different system class loader by default for this exact reason.   You can addpath after the JVM is started under their custom classloader.   If we need to make startJVM automatically add the java.system.class.loader option if it is not specified that is certainly an option we can consider.  Though we will probably need to change DynamicClassLoader a bit so it could potentially work as the system class loader rather than a dependent loader like it is now.   If you look at the mechanics you would see it is actually just delegating to URLClassLoaders to fetch resources and then loading those classes into itself so that everything appears to be coming from one classloader.  Unfortunately, the system classloader is only used for user classes as the bootstrap loader is used for all java.* domain prior to initializing the system loader.

The whole issue of how to get multiple Python modules to use JVM is complicated.   Each module has their own way they have tried tackling starting issues and the lack of a standard is certainly hurting interoperability.   I am not sure that the PyJNIus system where you have a config module that you load first before loading the Java module is a good approach.   The startJVM, shutdownJVM approach we have here certainly has a lot of downsides (especially on the shutdown where resources are yanked from underneath us).

As for as the module security flags, I would recommend testing with those simply to understand the issue.  We are not the only package that is having issues with module signing and permission issues introduced by the module system.   There are still warnings about unsafe access being issued by popular Java applications such as Maven.  I haven't messed with modules much but it is certainly a possibility that we place jpype in a module with controls if that is required.  I suppose there is some process we need in our build process to create modules and sign them with a key.

KOLANICH:
I wonder if we can cheat anyhow, by modifying already loaded classes (such as everything loaded by `bootstrap` loader) in an "unsafe" way (by messing with JVM internals from C++ side) to make JVM think that they were loaded by the same loader all the classes from dynamically appended packages are loaded with, and what will happen jn this case.

Thrameos:
I tried messing with Java internals in the past.  For example I managed to load a second copy of java.lang.String, but the module protection system prevents an instance from being respected.  It was actually to create a new copy of a ClassLoader where I could access the private fields so I can load and unload the Python wrapper classes that I played with it.

You would have to completely patch the Security system right down to the caller sensitive methods to work around it as far as I can tell.   Better to just disable module security prior to starting the JVM than try to fight the security system.   A lot of this comes from the days when Java was used for applets in browsers.   There were so many ways to worm down and breach security that they concluded that super gluing as much of this was the only way they would keep market share.   But that ultimately failed and Java got pulled as an acceptable browser plug in.

Java is pretty badly in needs of a redesign with way too much legacy stuff.  Too may classes should have been interfaces.  Too many concrete classes being used the default type to methods.    Missing functionality and poor security designs, inability to produce isolated spaces or load/unload classes for dynamic behavior.   Unfortunately, this sort of language shift is much too big for someone like me to enact.

KOLANICH:
I'm sorry for the late response.

1. With KS the new dynamic classpath changing works very well. Thank you for implementing this feature.
2. >java.lang.String is in module java.base of loader 'bootstrap') 
was likely caused not by classloaders conflict by by something else. The issue was solved with a hack of transforming a scala string into python string, and then back into java string. Then I have redone the architecture of the backend (I had to because KSC has changed) and the issue has gone. Likely some types were used in the places they were not expected in.

Thrameos:
Good to here the dynamic classloader is working for you.   If you see additional enhancements that would be nice be sure to put in a new  ticket.

I think that I still need to do some redesign of the JPype start up routine so that we have the ability to load some classes into the default class loader.   At least org.jpype needs to be in the system classloader.   This came up with the JDBC drivers recently.  There apparently are cases in which classes must be in the system classloader to work properly.  I could use an agent or specify the default classloader as being something other than the usual system classloader but that won't work once I have the reverse bridge.

PR #852, opened 2020-09-17
Do not fail on non-present folders

nils-braun:
The function `find_possible_homes` tests a given list of locations.
On very minimal systems (e.g. docker images), some of the folders might not be present (e.g. the `/usr/lib/jvm`). This however leads to a `FileNotFoundError` when calling `os.listdir`.

This PR fixes this problem.


marscher:
Thanks for your contribution. Adding a further test shouldn't be necessary in my opinion.

PR #855, opened 2020-10-11
Another stab at fixing Ctrl+C behavior

### LINKED ISSUE #556: Ctrl+C Segfault on Python/IPython

Thrameos:
This deals with a segfault in the Ctrl+C path.  There was a race condition that was leading to two exceptions on the stack.  I added addition interlocks and set up pattern which if a Ctrl+C is issued prior to calling Java the interrupt is cleared so that we don't get a sleep interrupted after a Ctrl+C in certain cases. 

This seems like a pretty important PR for 1.1 release.   It does not deal with the non-interactive case which still segfaults in the stack unwinds poorly.  That will need to be a followup, but it is less important.

Fixes #556 again

### DIFF
--- a/native/common/include/jp_context.h
+++ b/native/common/include/jp_context.h
@@ -77,6 +77,8 @@ class JPGarbageCollection;
 
 void assertJVMRunning(JPContext* context, const JPStackInfo& info);
 
+int hasInterrupt();
+
 /**
  * A Context encapsulates the Java virtual machine, the Java classes required
  * to function, and the JPype services created for that machine.
@@ -216,6 +218,7 @@ private:
 	JPClassLoader *m_ClassLoader;
 
 public:
+	JPClassRef m_ContextClass;
 	JPClassRef m_RuntimeException;
 	JPClassRef m_NoSuchMethodError;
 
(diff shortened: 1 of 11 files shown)


marscher:
Thank you very much for this quick solution. This seemed hard to solve and I think your approach is well suited to handle this ki d of race condition. Are you sure that it can only occur during method calls and getters? What about array/buffer io?
In any case I approve it for now.

Thrameos:
It is sort of hard to tell which commands would have triggered the unexpected interrupt after a Ctrl+C.  I should go back and do a further audit as it can only be tested by hand currently.  But that is a larger pull request as it required adding instrumentation to the Java code so that we can trigger the path and then testing of each branch condition that leads to it.   I was also worried that it would eat a lot of cycles as this type of problem usually causes lots of crosses of the JNI boundary, so my solution had not add much to the expected path given the exception after Ctrl+C was a rare problem.   Thus the solution with a local cache on the C++ side and native hooks to communicate with it.    The only time that it trigger will each time is if something triggers the interrupt handler in Java and the main thread is busy so another thread is repeatedly getting the check condition which will go to Java and find out that it isn't the thread holding the flag.   Not horrible, but given there were a bunch of edge cases this was the most reasonable solution I could come up with. 

We still have one segfault possibility on the non-interactive option that I have yet to resolve.  But it is better to crash than freeze when running large jobs that need to be stopped, so it hasn't been a high priority.

Unfortunately we have a leak report so I will need to do further work before we can cut a release.

PR #863, opened 2020-10-16
Fixed variadic method resolution

Thrameos:
This is a last minute bug fix regarding resolution rules for variadic arguments.   This bug has been around for a while (at least 0.6.2).   Variadic methods need to be considered polymorphically when deciding if they are to be hidden with respect to other methods.   I wasn't able to handle the corner case where one variadic method was being compared to another, but the more frequent case of a specific to a variadic have been resolved.   I added a test case.   

I would like this added to the 1.1.0 release if possible.   I was going to push the release out today, but this bug is holding up some other groups so it seems like a good candidate for a rush.


Thrameos:
@marscher any chance you can look this one over?

marscher (approved):

PR #870, opened 2020-10-22
Add import for importlibs.util

### LINKED ISSUE #869: Python-3.9 importlib api breakage
Current release 1.1.1 is incompatible (and untested) with Python-3.9

``` python
Python 3.9.0 | packaged by conda-forge | (default, Oct 14 2020, 22:59:50) 
[GCC 7.5.0] on linux
>>> import jpype
>>> jpype.startJVM()
  File "/home/marscher/sources/jpype/jpype/_core.py", line 215, in startJVM
    _jpype.startup(jvmpath, tuple(args),
AttributeError: module 'importlib.util' has no attribute 'find_spec'
>>> 
```

Thrameos:
Fixes #869

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,9 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 Latest Changes:
 - **1.1.2_dev0 - 2020-10-21**
+
+  - Add importlib.util to fix instability in Python 3.9
+
 - **1.1.1 - 2020-10-21**
 
   - Fixed packaging problem on linux.  
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -6,7 +6,9 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 Latest Changes:
 - **1.1.2_dev0 - 2020-10-21**
 
-  - Add importlib.util to fix instability in Python 3.9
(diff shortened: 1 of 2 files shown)


doc/CHANGELOG.rst
```diff
@@ -5,6 +5,9 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 Latest Changes:
 - **1.1.2_dev0 - 2020-10-21**
+
+  - Add importlib.util to fix instability in Python 3.9
```
gdubya, line 9: Is it specific to 3.9? i was under the impression that it's a more general Python issue.

Thrameos: Unfortunately I cant replicate it on 3.5 to 3.8 which are regularly tested. It was reported for 3.9 and one patch release of 3.7.  I suspect it is actually a site issue just starting python and checking sys.modules finds util with a working find_spec on all the versions i have available.  Some configuration must be out there which doesnt load it.  

If you have a recommended log entry that better describes the problem, just post it and I will make an update.  Unfortunately as I cant replicate it I cant determine what is trigger it.

Thrameos: I updated the comment to be less specific about the version and more about the problem.   I also added comments in the code on the kludge line to make it clear the purpose.

Thanks for submitting a review of the PR!

marscher (approved):
Thank you!

PR #872, opened 2020-10-23
Fix cflags for unix

Thrameos:
This is a shot at fixing the distribution size for unix.   I changed the patterns so we can explicitly set the debug and optimization flags from platform.  The requires scrubbing all the existing flags in the sysconfig.

Also fixes a bug with coverage and tracing.   Not sure why the extension position is hard coded, but the number drifted when we added the jar to the extension list.

### DIFF
--- a/setupext/build_ext.py
+++ b/setupext/build_ext.py
@@ -173,18 +173,6 @@ class BuildExtCommand(build_ext):
     2. try to disable warning -Wstrict-prototypes is valid for C/ObjC but not for C++
     """
 
-    # extra compile args
-    copt = {'msvc': [],
-            'unix': ['-ggdb', ],
-            'mingw32': [],
-            }
-    # extra link args
-    lopt = {
-        'msvc': [],
-        'unix': [],
-        'mingw32': [],
-    }
-
     user_options = build_ext.user_options + [
         ('android', None, 'configure for android'),
         ('makefile', None, 'Build a makefile for extensions'),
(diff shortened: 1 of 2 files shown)


setupext/build_ext.py
```diff
@@ -203,40 +191,41 @@ def initialize_options(self, *args):
...
+            for r in remove_args:
+                args = list(filter((r).__ne__, args))
+
+            cfg_vars[k] = " ".join(args)
         super().initialize_options()
 
     def _set_cflags(self):
         # set compiler flags
         c = self.compiler.compiler_type
+        jpypeLib = self.extensions[1]
```
marscher, line 214: this static indexing will hurt us in the future, won't it? (it already did, eventually put an assertion to match the desired name at least).

marscher (approved):

marscher:
Should we consider using an newer c++ standard, eg. 17? I think on Windows we already use it

setupext/build_ext.py
Thrameos: I address this with a list comprehension.   Should be a bit more future proof.

Thrameos:
It is true that standard 11 is rather ancient.  We can play around with other standards at some point.   Not sure what is the oldest compiler that we are likely to encounter currently.   The only reason for the VC to use a later standard is that the nothrow logic didn't get turned on without it.   As we are not currently using any later C++ other features I am not seeing much in the way of compelling reasons to change.   My time is a bit limited today (even through it is listed as my scheduled day off), so I am going to just try to cut a patch release.

Issue #876, opened 2020-10-28
PyInstaller support?


altendky:
Is there existing PyInstaller support or might it be added?  I am of course hoping this is a trivial thing to debug for someone with actual Java and JPype knowledge.  `:]`  I have been using javabridge for a few years but might be switching to JPype.  I don't do much to care about 'advanced' features but having the wheels would be nice.  I am presently working in Linux but would use this on both Linux and Windows.  I am getting the following `SystemError`.

```pytb
Traceback (most recent call last):
  File "j.py", line 6, in <module>
    jpype.startJVM()
  File "jpype/_core.py", line 221, in startJVM
    _jpype.startup(jvmpath, tuple(args),
SystemError: java.lang.ClassNotFoundException: org.jpype.classloader.DynamicClassLoader
```

As some probably irrelevant reference, this is for a private project I have that uses https://github.com/altendky/ccstudiodss for access to TI's Code Composer Studio DSS debugging tooling.

Here are some commands to recreate a trivial example failure from within a clone of the JPype repository.  I don't like to leave recursive delete commands laying around but maybe remove `build/`, `dist/`, and `venv/` before each run.

```
git rev-parse HEAD
git status
python3.8 -m venv venv
venv/bin/python --version --version
venv/bin/pip install --upgrade pip setuptools wheel
venv/bin/pip install . pyinstaller
venv/bin/pip freeze
cat > j.py << EOF
import os

import jpype
import jpype.imports
from jpype.types import *

for key, value in sorted(os.environ.items()):
    print(f'{key!r}: {value!r}')

print('+++ about to start JVM')
jpype.startJVM()
print('+++ JVM started')
EOF
venv/bin/pyinstaller j.py
venv/bin/python j.py
dist/j/j
```

<details> <summary>Expand for full console session</summary>

```bash
$ git rev-parse HEAD
b6d3996fe8d0ac58695aa9867ecf141dfc3196ec
```
```bash
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        j.py

nothing added to commit but untracked files present (use "git add" to track)
```
```bash
$ python3.8 -m venv venv
```
```bash
$ venv/bin/python --version --version
Python 3.8.6 (default, Oct  8 2020, 19:34:05) 
[GCC 9.3.0]
```
```bash
$ venv/bin/pip install --upgrade pip setuptools wheel
[...]

PR #877, opened 2020-10-28
Add PyInstaller hook

altendky:
#876

Draft for:
- [x] Documentation
- [x] Write test
- [x] Actually run the test
- [x] Check with my real program
  - JPype worked fine... then I realized my \_big\_ problem was going to be pytest, not Java access...  \*sigh\*

### DIFF
--- a/jpype/_pyinstaller/hook-jpype.py
+++ b/jpype/_pyinstaller/hook-jpype.py
@@ -1,8 +1,6 @@
 import os
 from pathlib import Path
 
-from PyInstaller.utils.hooks import relpath_to_config_or_make
-
 import jpype
 
 
--- a/jpype/_pyinstaller/hook-jpype.py
+++ b/jpype/_pyinstaller/hook-jpype.py
@@ -1,0 +1,13 @@
+import os
+from pathlib import Path
+
+from PyInstaller.utils.hooks import relpath_to_config_or_make
+
+import jpype
+
+
+fspath = getattr(os, 'fspath', str)
+
+jar_path = Path(jpype.__file__).parent.parent.joinpath('org.jpype.jar')
+
+datas = [[fspath(jar_path), relpath_to_config_or_make(fspath(jar_path))]]
--- a/jpype/_pyinstaller/hook-jpype.py
+++ b/jpype/_pyinstaller/hook-jpype.py
@@ -10,4 +10,4 @@ fspath = getattr(os, 'fspath', str)
 
 jar_path = Path(jpype.__file__).parent.parent.joinpath('org.jpype.jar')
 
-datas = [[fspath(jar_path), relpath_to_config_or_make(fspath(jar_path))]]
+datas = [[fspath(jar_path), '.']]
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -26,7 +26,7 @@ steps:
     python -m pytest -v --junit-xml=build/test/test.xml test/jpypetest --checkjni  
   displayName: 'Test JDK 11'
 
-# lack of src/ seems to confuse entry points
+# presence of src/ seems to confuse entry_points so `cd` elsewhere
 - script: |
     pip install .
     mkdir empty
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -26,7 +26,7 @@ steps:
     python -m pytest -v --junit-xml=build/test/test.xml test/jpypetest --checkjni  
   displayName: 'Test JDK 11'
 
-# presence of src/ seems to confuse entry_points so `cd` elsewhere
+# presence of jpype/ seems to confuse entry_points so `cd` elsewhere
 - script: |
     pip install .
     mkdir empty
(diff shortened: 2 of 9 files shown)


jpype/_pyinstaller/hook-jpype.py
```diff
@@ -0,0 +1,13 @@
...
+from pathlib import Path
+
+from PyInstaller.utils.hooks import relpath_to_config_or_make
+
+import jpype
+
+
+fspath = getattr(os, 'fspath', str)
+
+jar_path = Path(jpype.__file__).parent.parent.joinpath('org.jpype.jar')
```
Thrameos, line 11: We will need additional jars in the future as jpype is likely to depend on asm for things like object extension.   But this is sufficient for now.  We can add a glob based pattern when the number of items increases.

altendky:
@Thrameos, there are a couple aspects here that don't obviously fit into the way the rest of the project is.

1. PyInstaller has a setup for being able to run all installed PyInstaller hook tests.  This requires that those tests be installed.  We could run the tests in our own CI without this but we wouldn't quite fit into the PyInstaller plan.

2. PyInstaller uses `entry_points` for packages to publish their hooks.  This seems to require real installation rather than just the in-place build present in the CI tests.  So, I added a real installation step.

Side note, even if you do install the `entry_points` still don't work if you are in the root of the project since Python finds the `jpype/` directory instead of the installed package.  Or something...  I didn't dig into this hard but it seems to be another case of `src/` being a relevant project layout.  Anyways, that's what the `empty/` directory in CI is about.

I have my own opinions on these classic packaging choices but my goal here is definitely not to push those.  Let me know where you stand on these discrepancies and I'll adjust accordingly.

Thanks for your time today.

(and yes, still need to look at where to put this in the documentation...)

Thrameos:
With regard to the side tests, perhaps you can create a directory to operate in, then copy in whatever is needed to testing.   That would get you away from the jpype development version.   I am also not sure that we want to do the pyinstaller tests as part of the module testing as I am not sure what will happen when the normal pip test gets run.   Fortunately, I don't think a lot uses it as the CI directly calls pytest, so maybe not a big deal.  Though maybe I am missing a portion of the issue as I haven't studied it.

Unfortunately, I can understand how painful it is to work through CI issues.   I recommend that you use the disables on the other jobs so that you can speed the CI suite up while working through the patterns.   That is usually what I do so that I don't have to wait for the full suite to run.   Although Azure doesn't seem to have a hard cap on the jobs, it does have a job system which will reduce the number of cores when you try to debug a CI issue so reducing the total jobs usually helps it get through faster.

If it is too much of a burden you could just add another job rather than trying to work with in the existing jobs.   This isn't likely an architecture issue, so it would be fine to create a special job like the documentation task.  Just showing pyinstaller works on one would likely catch most issues.

altendky:
I've done way too much CI work to be worried by this little bit.  `:]`  But if I end up sucking down your resources, sure, we can certainly address that.  But for now, what I have done is at least in a working state.  I did overlook the `setup.py test` configuration though.  Not something I've seen a lot of lately.  Anyways, I only really raised these points because I don't like to come into a new project and contribute changes that don't really fit in.  So, I brought them up directly with at least attempted explanations.  If you are comfortable with the structure of the changes as-is, I am too.  If not then I'll adjust them with the corresponding tradeoffs.

The side note about Python finding the `jpype/` directory instead of the installed package was addressed by the `mkdir empty; cd empty` code in the Azure script.  Nothing needs to be copied around.  This is just an explanation.

Moving the PyInstaller test code out of the package and into the root `tests/` directory would be more consistent but means people won't run it when they ask PyInstaller to test all installed hooks.  Probably not a big loss.

For testing in CI we do need to do an installation, which looks to be new, such that the `entry_points` are setup.  Only question is if you want it to have a different structure than it does presently.  If you would like it only tested once, I can set that up.  I personally tend to go a bit crazy with the CI matrixing ending up with some projects having about 60 environments they run the full tests in...  So I would default to leaving it where it is and testing several times.

(an example of maybe too-many-combinations that I've set up... https://github.com/pytest-dev/pytest-twisted/actions/runs/332959941)

altendky:
I'm not sure if the link below will work well but you might want to take a look at the failed mac-3.8 build for b9d6be347e4e6399c7babccef5a722477cef02df.

```
AssertionError: Potential leak found
```

https://dev.azure.com/jpype-project/jpype/_build/results?buildId=707&view=logs&jobId=26347dc7-d814-5675-76ba-5a5891cd3096&j=26347dc7-d814-5675-76ba-5a5891cd3096&t=a256f958-ff63-5f57-fd8e-15df5a6ca1b1

Thrameos:
That mac resource leak has been bouncing on and off for a while.   I think it is an issue which how the leak checker is working rather than something real.   We use a statistical check on the resources expecting that we will force a garbage collect during the process.  Unfortunately, not every JVM and operating system releases the pages after the free.   This can cause a leak to be reported.  If it was a real leak it should have shown in all of the different builds.

marscher:
LGTM

PR #878, opened 2020-10-28
Add booleans to primitive convert list

hinerm:


hinerm:
This PR can be tested with a snippet such as:
```
from java.util import ArrayList
from java.lang import Boolean
MR = jpype.JClass("org.jpype.manager.MethodResolution")
param = ArrayList()
param.add(Boolean.TYPE)
MR.isMoreSpecificThan(param, param)   # I believe this should return true, which it does now after this patch
```

Thrameos:
LGTM.

PR #879, opened 2020-10-28
Fix ambiguity of matching array/single-var type

hinerm:


hinerm:
@Thrameos One other ambiguity we found.. does this fix seem reasonable to you?

Thank you for being so responsive!

Thrameos:
Seems good to me, but can you construct a test case where there ambiguity occurred so I can verify it?   I just want to make sure this doesn't introduce any rule conflicts.   We still have the horrible case of two variadic functions that I punted on, but seems like are making good progress.  (I should put an issue to add a MethodResolver testbench so we have clearly defined behavior going forward.)

@marscher Does this type of bug fix constitute a patch fix or minor release?  If it is minor then I should do a bumpversion so that our release number for the next version is on track?

marscher:
It seems like an edge case to me, but if you think it is kind of urgent to patch, then we can trigger the pipeline. Can I add you to PyPI, so you can push wheels by your own? I guess this would further speed up the process.

hinerm:
>but can you construct a test case where there ambiguity occurred so I can verify it?

@Thrameos yes definitely! Sorry I didn't have time to figure out how to run `pytest` properly yesterday. I got it now and tests will be incoming.

hinerm:
@Thrameos ok I added two test cases, `conflict4` and `conflict5`, that will fail if you revert https://github.com/jpype-project/jpype/pull/879/commits/a63d7c2b14d22edb71531cbfc0aa3d6816c871ce. Note that `conflict5` will randomly pass or fail pre-patch because the method choice was arbitrary for a single-param method. In any case, with this fix it should hopefully be nice and deterministic.

Thrameos:
LGTM.

@marscher  I believe this PR is targeting scyjava/scyjava#18.   My guess is they are looking for roll out in mid-November.   We always get a few bug reports after a minor release so I am fine will waiting a week or two and seeing what else comes in.   I was more interest as to whether you thought if this sort of bug release should be 1.1.3 or 1.2.0 when we do release?  

It is clear that we need to do a major release if we remove an existing api, and we should do a minor release if something gets added.   But this is more subtle.   If someone requests 1.1.x they may get 1.1.3 or they may get earlier.  The earlier will produce the erroneous reporting but only on this challenging case where a method was overloaded between "foo(double)" and "foo(double...)".  I would guess that it is bug level and so it should just be a patch, but as you are the maintainer I think it is your call.

Thus far in this version we have two PR for MethodResolution bugs and one for pyinstaller which is being worked on.   I would be fine with those being patch release because they don't add anything to our public API.  There is the dynamic class exposure PR but I believe when that goes through we have to call it a minor release to 1.2.

Thrameos (approved):
All looks good.  Thanks for the contribution.

hinerm:
@Thrameos is there any chance of a `1.1.3` or `1.2.0` release this week? We have a conference coming up next week and it would be awesome to be able to have the new version.

Thank you, as always!

Thrameos:
I think we can have a release in that time frame.    It will mostly be a bug fix release, but we do have the added features that you can add a jar to the classpath after the JVM is started.

@marscher I haven't heard back from you regarding the Python 3.9 distribution.   Is this something you want me to take a look at?

marscher:
That would be great. Wirhout proper access to the azure resource, I don't have the needed insight how I broke the configuration.
3.9 is default now for the popular Anaconda distribution, so I expect complains soon, if we would not ship it soon.
Thanks for taking a shot.

Thrameos:
Okay I will try to get it updated on Turkey day.   I am still waiting from an email from you so that we can set up a webex to see what is wrong with those project settings.   It would be good to get them fixed.  Not that I plan to get run over by a bus, but every so often they decide they need to me go work out some place remotely for a month or two at a time.

Thrameos:
@hinerm A quick survey now that you have worked through the converting to JPype?  Was there anything else that needed to be addressed for switching to JPype?   Were there features that you would like to see added or additional documentation that would have made the process any easier?

hinerm:
@Thrameos I thought Jpype was very well documented. There were some syntax changes from pyjnius, but typically things made sense and were resolvable when consulting the docs. I think overall it even simplified our code.

One thing we're excited for is the potential of bidirectionality with Java. That will just be an added bonus for us.

I know we have people who want to use our project in their daily research, so if we run into any issues we'll be sure to let you know. But so far, from what I can tell, the transition has been pretty smooth. We appreciate using a product that is well-maintained.

Oh! Manually installing dev versions of jpype to test changes on windows was a pain for me, as a total python newbie, because I had to install and conda-fy a visual studio "*developer* command prompt." But that's not your fault :-p

Thrameos:
@hinerm I am working on pressing forward on the release, but Python 3.9 on windows is not looking promising.   For reasons completely beyond my understanding the internal module will not load properly from a pyc file.  Thus the first time you call jpype everything is fine; the second time it hits the cache and fails.  The exact same code function perfectly in 3.8.

There are some possibilities.
- Python 3.9.0 on windows is just busted and we need to find a place to report the bug or find similar behavior in their bug tracker so that we know it will be addressed.  We can't make it function until 3.9.1.
- Python 3.9.0 changed the start of routine requirements for modules and we need to make a corresponding code change with a ``#define``.  I saw one deprecation warning but nothing I did seemed to affect it.
- My test system is busted and this is a local error.   But given we see something similar I am not sure this one is likely.

I don't really have an issue with cutting a release that does not function for Windows, but it will pretty quickly hit the anaconda release system and fail there as well.  I will keep you posted.

Thrameos:
@hinerm The release cycle is started (my renter bailed due to job loss so I had to deal with it over the weekend).  Assuming nothing fails in the build it should be ready to start distribution tomorrow.   If it fails (which sometime happens if a new feature breaks the release script) then I should be able to correct it tomorrow evening time frame.    So I am guess it should hit the distributions on Tuesday.

ctrueden:
@Thrameos @marscher Thanks a lot for pushing on the release! It will be really helpful to simplify the environment setup instructions.

@Thrameos Since you asked about additional things needing to be addressed: there are still overloaded method resolution issues in certain cases. We'll try to file PR(s) for those if possible. And I vaguel recall encountering one other small issue during the past 48 hours of coding... I will try to give more details in the coming days.

Thrameos:
Which instructions specifically?  I dont recall having env instructions.  But i may be mistaken.

I am not sure I would expect a set to match a java collection.  Sets are unordered and lists are ordered.  But you are correct that Python has free conversions from ordered and unordered types.  So from that prospective it would make sense.  I wonder though if there ever is a case where it would introduce an ambiguity between an overload with ordered and unordered Java interfaces?  Seem like a rare case.

ctrueden:
> Which instructions specifically? I dont recall having env instructions.

Apologies—I meant that the JPype release will make our tutorial's setup instructions simpler, because we won't have to pin the environment to a hash on GitHub. Not an action item for anyone on the JPype team. :smile: 

> I am not sure I would expect a set to match a java collection.

My intuition was that a Python set could be used in place of a Java `Set`, in the same way that a Python list can be used in place of a Java `List`. Because Java `Set` implements `Collection`, I thought this would work. In Java, `Collection`s are not necessarily ordered, only iterable, same as in Python. No?

Thrameos:
I agree.  I will see if I can make set match.

Thrameos:
Okay should testable in #899.   Unfortunately it is currently taking a double conversion path as there is no direct method to get from set to Object[] included.   But as I am working on the bridge in which a set will directly be a Java collection (without the need for any conversion) then I don't think this is an issue going forward.

Please note that one downside of the reverse bridge is that certain types such as int, long, list, set, etc will not be converted to Java but instead just wrapped with a Java type which implements the nearest behavior.  For example in your set example arguments "1,2,3" are currently going to java.lang.Long but instead they may end up as java.lang.Number.   Python "Sets" currently go to Java ArrayList when passed as Java "Collection", but that is a side effect of limitations of converter implementation, and not the plan going forward. 

I have tried to make it clear in the documentation that we just guaranteed to get something that implements a nearest type, but not necessarily a particular type.  Unfortunately, it is pretty easy to write code that assumes one of these types accidentally.  The hard one will be again deciding what to do with Strings.   Java made this very complicated by forcing strings to be a concrete class rather than an interface.  Thus ``Object[:](("a","b","c"))`` leaves open two possible interpretations, an array with 3 Java strings or and array with 3 ``java.lang.CharSequence`` which point to Python string instances.  Given strings are imutable and almost nothing takes a CharSequence I will likely make special cases for these sort of types.  I may have to bump to version 2.0 just to avoid having this listed as breakage.

PR #886, opened 2020-11-12
Documentation of issue

### LINKED ISSUE #885: JPype crashes when JIT compilation is enabled
Hi,

I'm running into an issue where JPype crashes when JIT compilation is enabled, I am forced to use -Xint when starting the JVM to prevent the failure, which can happen in arbitrary places.

It is hard for me to tell which environment crashes first: the JVM or the Python process, see error message below. I've tried on different JREs (Oracle 1.8.0_151 & 1.8.0_172, OpenJDK 11.0.1) and the problem remains. I am on MacOSX 10.14.6 and uses Jpype 1.1.2 on Python 3.8.6.

I don't know if and why that could be the issue but the Java code I'm running is written originally in Kotlin. Is there anything I can try to reenable JIT compilation, since without it performances are extremely poor?

Thank you,
Karl

Error message:
```
Fatal Python error: Bus error
Thread 0x000070000261d000 (most recent call first):
<no Python frame>
#
#
#  Internal Error (sharedRuntime.cpp:549), pid=9833, tid=0x0000000000000307
#  guarantee(cb != NULL && cb->is_nmethod()) failed: safepoint polling: pc must refer to an nmethod
#
Fatal Python error: Aborted
```

Thrameos:
Fixes #885

### DIFF
--- a/doc/userguide.rst
+++ b/doc/userguide.rst
@@ -3182,6 +3182,23 @@ this block as a fixture at the start of the test suite.
 This code enables fault handling and then returns the default handlers which
 will point back to those set by Java.
 
+Another issue was reported with interactions between JIT and faulthandlers.
+The Java Just-In-Time (JIT) compiler determines when a portion of code needs to
+be ported into machine code to improve performance.  When it does so it
+triggers either a SEGSEGV or SEGBUS depending on the machine architecture which
+breaks out any threads which are currently executing the existing code.
+Because the JIT compiler self triggers, this will cause a failure to appear in
+a call which worked earlier in the execution without an issue.
+
+If the Python fault handler interrupts this process, it will produce a "Fatal
+Python Error:" followed by either SIGSEGV or SEGBUS.  This error message then
+fails to hit the needed Java handler resulting in a crash.  This message will
+disappear if the JIT compiler is diabled with the option "-Xint".  Some modules
+such as Abseil Python Common Libraries (absl) automatically and unconditionally
+install faulthandlers as part of their normal operation.  To prevent an issue
+simply insert a call to disable the faulthandler after the module has enabled
+it.
+
 
 Unsupported Python versions
 ---------------------------
--- a/doc/userguide.rst
+++ b/doc/userguide.rst
@@ -3182,22 +3182,33 @@ this block as a fixture at the start of the test suite.
 This code enables fault handling and then returns the default handlers which
 will point back to those set by Java.
 
-Another issue was reported with interactions between JIT and faulthandlers.
-The Java Just-In-Time (JIT) compiler determines when a portion of code needs to
-be ported into machine code to improve performance.  When it does so it
-triggers either a SEGSEGV or SEGBUS depending on the machine architecture which
-breaks out any threads which are currently executing the existing code.
-Because the JIT compiler self triggers, this will cause a failure to appear in
-a call which worked earlier in the execution without an issue.
+Python faulthandlers can also interfer with the JIT compiler.  The Java
+Just-In-Time (JIT) compiler determines when a portion of code needs to be
+compiled into machine code to improve performance.  When it does, it triggers
+either a SEGSEGV or SEGBUS depending on the machine architecture which breaks
+out any threads which are currently executing the existing code.  Because the
+JIT compiler self triggers, this will cause a failure to appear in a call which
+worked earlier in the execution without an issue.
 
 If the Python fault handler interrupts this process, it will produce a "Fatal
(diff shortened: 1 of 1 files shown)


Thrameos:
@karllessard not so fast there!  You owe the project a review before the issue is actually closed.  I may not be able to fix it but I can at least document it.  Please look over the proposed documentation, make sure it would have directed you to how to recognize and fix the issue, and verify my engrish is parsable.

doc/userguide.rst
```diff
@@ -3182,6 +3182,23 @@ this block as a fixture at the start of the test suite.
...
+be ported into machine code to improve performance.  When it does so it
+triggers either a SEGSEGV or SEGBUS depending on the machine architecture which
+breaks out any threads which are currently executing the existing code.
+Because the JIT compiler self triggers, this will cause a failure to appear in
+a call which worked earlier in the execution without an issue.
+
+If the Python fault handler interrupts this process, it will produce a "Fatal
+Python Error:" followed by either SIGSEGV or SEGBUS.  This error message then
+fails to hit the needed Java handler resulting in a crash.  This message will
+disappear if the JIT compiler is diabled with the option "-Xint".  Some modules
```
karllessard, line 3196: `diabled` -> `disabled`

Also, I would mention that while adding `-Xint` prevents the problem to appear, it deteriorates substantially the performances of the Java code, so disabling faulthandler should prevail as the right solution whenever possible.

doc/userguide.rst
```diff
@@ -3182,6 +3182,23 @@ this block as a fixture at the start of the test suite.
...
+such as Abseil Python Common Libraries (absl) automatically and unconditionally
+install faulthandlers as part of their normal operation.  To prevent an issue
+simply insert a call to disable the faulthandler after the module has enabled
+it.
```
karllessard, line 3200: maybe show how to disable faulthandlers? i.e.
```python
import faulthandler
faulthandler.disable()
```

karllessard (commented):
Here some comments @Thrameos but don't count on me for correcting your engrish as my frenglish is pretty much the same :D

doc/userguide.rst
Thrameos: Done and done.

doc/userguide.rst
Thrameos: Added.

karllessard (approved):
Excellent, thanks again @Thrameos !

Thrameos:
@marscher I will leave this one open in case you feel like reviewing.  But if you don't have time I will merge it before the next release.

I looked into the underlying CPython implementation.   I really don't have a clue why it is interacting like this.   The signals are per thread issued, but the CPython implementation appears to be trying to reroute them to a single thread which is likely why they don't get back to Java properly.  But that is just a hunch.   As documented it is supposed to just print an error and then proceed to call the previous handler.

```
static void
faulthandler_fatal_error(int signum)
{
    const int fd = fatal_error.fd;
    size_t i;
    fault_handler_t *handler = NULL;
    int save_errno = errno;

    if (!fatal_error.enabled)
        return;

    for (i=0; i < faulthandler_nsignals; i++) {
        handler = &faulthandler_handlers[i];
        if (handler->signum == signum)
            break;
    }
    if (handler == NULL) {
        /* faulthandler_nsignals == 0 (unlikely) */
        return;
    }

    /* restore the previous handler */
    faulthandler_disable_fatal_handler(handler);

    PUTS(fd, "Fatal Python error: ");
    PUTS(fd, handler->name);
    PUTS(fd, "\n\n");

    faulthandler_dump_traceback(fd, fatal_error.all_threads,
                                fatal_error.interp);

    errno = save_errno;
#ifdef MS_WINDOWS
    if (signum == SIGSEGV) {
        /* don't explicitly call the previous handler for SIGSEGV in this signal
           handler, because the Windows signal handler would not be called */
        return;
    }
#endif
    /* call the previous signal handler: it is called immediately if we use
       sigaction() thanks to SA_NODEFER flag, otherwise it is deferred */
    raise(signum);
}
```

It doesn't appear possible to monkey patch the handlers either at least without potentially messing up the tables.  I can't really be proactive either, because it is costly to disable them and I don't see any way to check if they have been enabled without a performance penalty.

marscher:
I like the approach of having users to explicitly disable JIT, but isn't JIT compilation the default behavior of modern JVMs? I wonder why we haven't walked into this earlier.

Thrameos:
It is the faulthandler that is the issue.  Normally python modules leave the fault handler off.  But apparantly absl and pytest turn it on.  We have seen problems in the past, but this is the first time it was tied to JIT.   Normally faulthandler busts the GC first.

PR #887, opened 2020-11-13
Add support for building on z/OS

pitmanst:
This just let's us be able to install JPype when using z/OS. There's two problems on the platform currently:

1) `setupext/platform.py` - When running the installation with `JAVA_HOME` set, `jni_md_platform` will be set to None, but it will have found jni.ini, and so attempt to use `jni_md_platform` and error out. This sets it properly so the build can continue.

2) `setupext/build_ext.py` - When compiling the Java source files with javac, the expected/default encoding for the source file is EBCDIC on z/OS. When pulling down the source from git or similar has this in ascii and not EBCDIC, it'll throw an error when it tries to compile it. We put the -encoding to signify the expected encoding of the file. I don't expect any additional failures from this (no new failures from my test Ubuntu 18.04 machine), but it could be wrapped in a check for z/OS if needed.

### DIFF
--- a/setupext/build_ext.py
+++ b/setupext/build_ext.py
@@ -313,7 +313,7 @@ class BuildExtCommand(build_ext):
             build_dir = os.path.join(self.build_temp, ext.name, "classes")
             os.makedirs(build_dir, exist_ok=True)
             os.makedirs(dirname, exist_ok=True)
-            cmd1 = shlex.split('%s -cp "%s" -d "%s" -g:none -source %s -target %s' %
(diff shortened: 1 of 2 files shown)


Thrameos:
Will the encoding change make the resulting jar incompatible with other jars?  Do we need a special string encoder for this platform?

pitmanst:
There should not be be any incompatibilities with the resulting jar files. This change is only to specify what the .java source file is encoded in.

Thrameos:
Okay that makes sense.   I didn't realize that javac would be expecting a different encoding on some systems.

Thrameos:
This all looks good to me.   I will target this for the 1.2.0 release if that is okay.  We don't have a target release date yet, but should be in the December to January time frame.

PR #888, opened 2020-11-14
Load a class after starting JVM

Thrameos:
This PR allows for a jar to be loaded after the JVM is started.  This will allow Python modules to start the JVM and then load Java using modules that require jars to be added to the path to function properly.   To make this work the default classloader for JClass is now org.jpype.classloader.DynamicClassLoader.  There were some required enhancements to the getResource and getResources to make the dynamic loader more functional.   This PR also modifies the package directory system to update the package contents with the new jars.   

I added a CHANGELOG entry and a test for the new functionality though further documentation will be required.


Thrameos:
Example:

```
import jpype
import jpype.imports
from jpype.types import *
jpype.startJVM(classpath='project/jpype_java/dist/*')

# Try to get class which is not in the class path
try:
    print(JClass("org.objectweb.asm.AnnotationVisitor"))
except Exception as ex:
    print(ex)

# Add a class path after the load
jpype.addClassPath('lib/*')

# Try again after adding
try:
    print(JClass("org.objectweb.asm.AnnotationVisitor"))
except Exception as ex:
    print(ex)



try:
    print(JClass("jpype.array.TestArray"))
except Exception as ex:
    print(ex)

jpype.addClassPath('test/classes')

try:
    print(JClass("jpype.array.TestArray"))
except Exception as ex:
    print(ex)
```

Thrameos:
This PR should also call a clear cache on the JPackage so that we can import without needing to call JClass directly.

Thrameos:
Now properly handles JPackage and import instances for late loaded classes.

Thrameos:
@marscher  it is a bit unclear how we should go about testing this code.  We would have to have the harness build a number of jars so that we can test before and after.   But this seems like a lot of effort unless we make a lot of changes to our system.    I suppose I could just make the jars in the project directory and then check them into the repo and source distribution rather than trying to build them from within Python.  Any opinions on this?

This PR still requires a test bench and documentation before it is ready for review.

This is a large enhancement to an existing method so I guess we should target it for 1.2.

Thrameos:
@marsher I just realized that this one was never tagged for review.    It is an important features so it should likely go out.   However, I will need to resolve some conflicts.

test/jpypetest/test_imports.py
```diff
@@ -115,6 +115,20 @@ def testMRJar(self):
         self.assertTrue("B" in u)
         self.assertTrue("sub" in u)
 
+    def testAddClassPath(self):
```
marscher, line 118: could you run a second test like this in a subprocess, in order to check the "*" path which is yet missing?

marscher (approved):
What a nice improvement! Thanks a lot

test/jpypetest/test_imports.py
Thrameos: I added a test by simply importing a module in the test directory with a second jar file as it was easier than getting the subshell stuff working.   I believe it should cover what you are looking for.   Unfortunately star is only supported at the module level so this was the cleanest solution I could come up with.

Thrameos: I will likely have to disable this test for jacoco coverage tests.   Jacoco installs a "Java agent" which has a side effect of erasing the directories from jars when it recompiles the byte code.  They probably just disregarded the entries rather than passing them through.   Not sure that it is a big deal but certainly annoying for importing star.

Thrameos: Actually I may have spoken too soon.  It looks like autopep8 is messing with my test.   I will have to find a way to keep the second import in place.

PR #889, opened 2020-11-14
Updates to quickstart guide

Thrameos:
We have had a lot of changes to make Python statements more compact such as arrays and lambdas.   This update brings the quickstart guide up to date with those changes.


marscher (approved):
thanks!

PR #895, opened 2020-11-21
Fix for MRJAR bug.

### LINKED ISSUE #838: Cannot import some Lucene classes using OpenJDK 11
I am using Lucene 8.6.0 on a project of mine. I am using Python 3.8.2 on Pop! OS (Ubuntu) and Python 3.8.5 on RHEL 7.8.

The following code fails on OpenJDK 11 and OpenJDK 14 but works just fine on OpenJDK 8:

```
$ export CLASSPATH=lib/lucene-core-8.6.0.jar
$ python
Python 3.8.2 (default, Jul 16 2020, 14:00:26) 
[GCC 9.3.0] on linux
>>> import jpype
>>> import jpype.imports
>>> print(jpype.getDefaultJVMPath())
/usr/lib/jvm/java-14-openjdk-amd64/lib/server/libjvm.so
>>> jpype.startJVM(jpype.getDefaultJVMPath())
>>> from org.apache.lucene.search import BooleanClause
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 652, in _load_unlocked
AttributeError: type object 'org.apache.lucene.search.BooleanClause' has no attribute 'loader'

```
Am i doing something wrong?

Thrameos:
This fixes a bug that results from MRJAR format.  Java fails to live up to its contract for MRJAR files with regard to getResources.

> Finds **all** the resources with the given name. A resource is some data (images, audio, text, etc) that can be accessed by class code in a way that is independent of the location of the code. (emphasis added)

Instead with MRJAR files it only finds the directory that belongs to the version specified even it that directory contains only one or two files.   Thus, breaking our package directory system.  

This PR creates work arounds for the bug by probing to see if the Jar is a MRJAR and interrogating both directories for contents.   Worst case it will probe the directories an extra time and the version ones will overwrite the non-version ones which is consistent with the specifications.

Fixes #838


jpype/imports.py
```diff
@@ -198,7 +198,7 @@ def find_spec(self, name, path, target=None):
             try:
                 # Use forname because it give better diagnostics
                 cls = _jpype.JClass("java.lang.Class").forName(name)
-                return _jpype.JClass(cls)
+                raise ImportError("Class `%s` was found but was not expected" % name)
```
marscher, line 201: This message will always be overwritten by the one of ImportError in line 204, is this intended?

native/java/org/jpype/pkg/JPypePackageManager.java
```diff
@@ -383,6 +383,26 @@ private static void getJarContents(Map<String, URI> out, String packageName)
...
+        int index = schemePart.indexOf("!");
+        if (index != -1)
+        {
+          if (schemePart.substring(index + 1).startsWith("/META-INF/versions/"))
+          {
+            int index2 = schemePart.indexOf('/', index + 20);
+            schemePart = schemePart.substring(0, index + 1) + schemePart.substring(index2);
+            URI resource2 = new URI(resource.getScheme() + ":" + schemePart);
+            Path path3 = getPath(resource2);
+            collectContents(out, path3);
```
marscher, line 402: why can't we break here, after treating the special case? If there are more jars to inspect?

marscher (approved):
Thanks for treating this special case. Never heard of this special Jar format before though.

marscher:
Would you mind adding a test case for this case (e.g. a very tiny special jar and match the contents in the unit test)?

jpype/imports.py
Thrameos: Yes intent was to always issue the import error with some cause.  Thus the line above should always have failed.  The question being what to do if it did not fail.  So i issued a new exception saying this should have failed which will be the cause of the actual failure mesg.  I can try to make it cleaner but it should never happen.

native/java/org/jpype/pkg/JPypePackageManager.java
Thrameos: We have to search both the non version case and the versioned case.  It is a weird format.   Mind you we could use this format to make org.jpype run on many versions but the structure of setting one of these up is not that fun.  So instead i use reflection when we have version issues.

Thrameos:
I was aware of multi release jars but i hadn't seen one in the wild.  They are a tool when you wanted to use some new Java feature but still want to support old versions.  Mostly use on the Java 8 to 9 transition because only 8 does not get the layer.  There is also the potentionally horrible case where it has several version layers 9 and 11.  Not sure how i can handle that.

Okay i will try to create a test.  The hard part will be making the test pass on both 8 and 9 as they require different outcomes.  In 8 it should dir to class A only.    In 9 it would dir to class A and class B.  We were getting a dir of class B only which was unacceptable.  So i guess the check must always contain class A and if jdk version gt 9 must also contain class B.

Thrameos:
I added the test as requested.  I left the build script under the project directory in case the jar file ever gets lost.   This PR will be merged once it clears the CI.

marscher:
Thanks for fixing this tough one and creating a proper test.

PR #897, opened 2020-11-30
Cycle

Thrameos:
Start patch cycle for 1.2.0


marscher:
Is master without this PR ready to create a source distribution (python setup.py dist)?

marscher:
If not, I wonder from which branch you've created the 1.2.0 binaries

marscher:
never mind, I just got confused :)

marscher:
When creating the source archive from the 1.2.0 branch, I get the following error (which is a minor detail), but should be fixed.
```
Generating test/classes/jpype/doc/Test.html...
...
test/harness/jpype/doc/Test.java:35: error: attribute not supported in HTML5: summary
subprocess.CalledProcessError: Command '['javadoc', 'test/harness/jpype/doc/Test.java', '-d', 'test/classes/']' returned non-zero exit status 1.
```
OpenJDK version 14, I guess with the older versions this was not a problem. I will switch to the tested (11?) JDK version for the generation of the archive.

PR #899, opened 2020-11-30
Conversion for set and iterable.

Thrameos:
This is a draft PR that adds some additional conversions for Python set as well as alternatives to hit Java Iterable.   This PR requires documentation, change log, and tests before it can be included.

### DIFF
--- a/jpype/protocol.py
+++ b/jpype/protocol.py
@@ -37,7 +37,7 @@ _JBuffer = _jpype._JBuffer
 
 if sys.version_info < (3, 8):  # pragma: no cover
     from typing_extensions import Protocol, runtime_checkable
-    from typing import Sequence, Mapping  # lgtm [py/unused-import]
+    from typing import Sequence, Mapping, Collection, Set  # lgtm [py/unused-import]
     from typing import SupportsFloat, Callable  # lgtm [py/unused-import]
 
     @runtime_checkable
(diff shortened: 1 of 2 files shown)


jpype/protocol.py
```diff
@@ -68,12 +70,23 @@ def _JFileConvert(jcls, obj):
...
     return _jclass.JClass('java.util.Arrays').asList(obj)
 
 
+@_jcustomizer.JConversion("java.lang.Iterable", instanceof=Set)
+@_jcustomizer.JConversion("java.util.Collection", instanceof=Set)
+def _JSetConvert(jcls, obj):
+    # set does not satisfy PySequence_Check and collection is too broad as it
+    # would let dict be converted, so we are going to have to convert twice
+    # for now
+    return _jclass.JClass('java.util.Arrays').asList(list(obj))
```
marscher, line 87: why not use Javas HashSet or TreeSet here?

Thrameos: We can use any Iterable type which contains the required elements.   We could manually create a HashSet, but as the method we are passing it to isn't actually  using the Set<> API this would simply be an expensive conversion without much benefit.  It is true the converting to a HashSet would allow the Iterable to cast to a Set at some future point, though I don't think may Java code would expect this to work.

```
   void foo(Iterable i)
   {
      Set set = (Set) i;
      ...
    }
```

Technically we should be wrapping to an unmodifiableSet using Collections as the Iterable since it cannot be modified currently.   I haven't tested what happens if we modified the list, though as it is backed by a Object[] it should cause an error as expected.

Long term, we won't actually convert anything here.   Once the reverse bridge is complete we can just pass Python sets to the converter to get a python.lang.PySet instance which implements Set<> and Iterable<> so problem solved.

PR #910, opened 2020-12-17
Add aarch64 wheel support

odidev:
Fix: Add aarch64 wheel build support using qemu. 
Few build failures are there on Windows not related to my changes.

### DIFF
--- a/.azure/release.yml
+++ b/.azure/release.yml
@@ -34,6 +34,11 @@ stages:
           plat: manylinux2010_x86_64
           image: quay.io/pypa/manylinux2010_x86_64
           python.architecture: x64
+        64Bit2014:
+          arch: aarch64
+          plat: manylinux2014_aarch64
+          image: quay.io/pypa/manylinux2014_aarch64
+          python.architecture: aarch64
         64Bit:
           arch: x86_64
           plat: manylinux1_x86_64
--- a/.azure/scripts/wheels-linux.yml
+++ b/.azure/scripts/wheels-linux.yml
@@ -10,6 +10,13 @@ steps:
     path: dist
 
 - script: |
+    sudo apt-get update
+    sudo apt-get install -y qemu qemu-user-static qemu-user binfmt-support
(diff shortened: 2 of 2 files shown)


marscher:
@odidev It seems there is a syntax error with changed YAML files, as the azure pipeline did not even start. This kind of error reporting is undesirable (from m$, but what would you expect?)

@Thrameos could you spot this mistake without digging deep?

Thrameos:
@marscher I can see why it did not trigger.   This is a release script patch not a build script patch, so since no sources were changed and no files touched that would change the outcome of a build there was no reason for the build pipeline to run.   Unfortunately there is no way to test a release script PR as release scripts have to be triggered manually.  

If you are fine with adding this code then just approve it and I will have to deal with it by merging and then making adjustments in the release script on fails.  The primary reason for this to protect keys that we may be using for signing from being triggered by user alterations.  We don't have a push to pypi direct so security on the release path is not much of a concern, but at some point it would be nice if we could sign the releases properly.

I am a bit concerned that it is added a QEMU step to all the Linux builds rather than just the one used for QEMU.   This may be solvable by simply added a conditional to QEMU triggered off the arch section, but I would likely have to take a few shots playing with the yaml file to get it to work properly.

For reference in case this happened on another PR, the controls for which files will trigger a build are specified at the start of the build script.

```
# JPype CI pipeline
pr:
  branches:
    include:
    - master
    - releases/*
  paths:
    include:
    - .azure/build.yml
    - doc/*
    - setup.py
    - setupext/*
    - jpype/*
    - native/*
    - test/*
 ```

Currently any changes outside of that list of directories will not trigger the build.

marscher:
We should add QEMU only for aarch64 I'd guess. @odidev Could you give it a shot again, please?

odidev:
Pushed changes to add QEMU only for aarch64 builds.

marscher (approved):
Thanks, this looks sane to me. Thanks a lot @odidev

PR #912, opened 2020-12-20
Bug fix for Python 3.9

Thrameos:
This fixes that problem with importing JPype on window 3.9.   A single static variable initializer was calling PyDict_New in pyjp_class.cpp which is not allowed because library loading does not hold the GIL.  However, a bug in Python which was not releasing the GIL was allowing the bad code to function.  When they corrected the GIL release bug it yanked the rug out from under us.  

Unfortunately as it did warrant a note in the change log in Python 3.9 for "Windows" nor had anyone reported this message as a symptom, it was a cold bug search for a bug that disappeared depending on the path that was taken.

This PR is end state after debugging.  So not every change was necessary to fix the bug.  If anything looks objectionable I can pull it.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -12,8 +12,6 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import jpype
-import jpype.imports
 from unittest import mock
 import sys
 import os
@@ -146,6 +144,8 @@ mockModule._JPackage = _JPackage
 mockModule._JClassHints = _JClassHints
 mockModule._hasClass = lambda x: False
 sys.modules['_jpype'] = mockModule
+import jpype
+import jpype.imports
 
 # For some reason jpype.imports does not work if called in sphinx. Importing
 # it here solved the problem.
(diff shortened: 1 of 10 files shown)


marscher:
Your windows users will be grateful. Thanks for fixing this! The other changes seem sane and helpful as well.

marscher (approved):

marscher:
btw, unittest.mock doesn't seem to add the  `__name__` attribute for mocked objects (needed for buildthedocs service).

Thrameos:
I am trying to figure out the doc one now.   I can't replicate it locally yet.  Once I do I guess I should start the patch release process again.

Thrameos:
I figured out the ``__name__`` problem.  We were missing two of our TypeMock statements.   In early versions of Sphinx this was being silently ignored, but now it is triggering an error.   I monkey patched it so that it would produce a FIXME which will appear in the documentation rather than crashing for now.   Figuring out which TypeMock is missing is nearly impossible as the name of the class never appears on the stack trace.

marscher:
I remember that this was painful when working for another project using c-extensions. Thanks a lot for figuring this out and of course for fixing the GIL related initialization.

PR #919, opened 2021-01-04
Cycle

Thrameos:
This is the usual post release merge that incorporates all the fixes needed to make a release.  Items covered here

- Update the change log and bump the version.
- There was a problem with the pattern for deciding when to run the fast tests.  Due to typo it was not actually running any tests in the CI for architectures other than linux coverage tests.  This meant that some undiscovered errors crept in which had to be resolved at release time.
- Windows adding of missing directories had issues.   The URI protocol and the Windows Path specification are not compatible due to differences in Path separators.   It was a minor change to convert to a URI first before performing text merging to get the Jar URI specification.   There are still 2 extra slashes in the path, but it appears to be harmless.   I am not sure why the default path specification is missing the extra slashes as it is supposed to be a root origin path.   But apparently the code in Jar does not require it.  
- A version of Jedi is having issues.  It is possible that a bad behavior in JPype is triggering it, but it will take time to resolve which I am short on right now, so I just disabled the problematic tests.  We should make an issue to investigate it at a later point.

@marscher I deleted most of the old branches used in the release process.   I could tag on release rather than releases/vXXX if you would prefer though it is much safer to tag and release to releases/vXXX because I can be sure that I am at the right point on azure and when performing the tagging.   Working with branches is a easier than working with tags when resolving release issues so I would prefer to keep the current procedure.  We can delete the branches after the release clears if you feel the extra branches are an issue.  Just update the release procedure under doc/release.rst so that next release is executed differently.  At minimum we still need ``master`` for incoming PRs, ``release`` which has the current release head (which is the only one that has non ``_dev`` version tags, and of course ``cycle`` which is used to bump the master so that we don't ever get confusion between master and releases.

### DIFF
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -25,12 +25,12 @@ steps:
 - script: |
     python -m pytest -v --junit-xml=build/test/test.xml test/jpypetest --checkjni
   displayName: 'Test JDK 11'
-  condition: eq(variables.jpypetest.fast, 'false')
+  condition: eq(variables['jpypetest.fast'], 'false')
 
 - script: |
     python -m pytest -v --junit-xml=build/test/test.xml test/jpypetest --checkjni --fast
   displayName: 'Test JDK 11 (fast)'
-  condition: eq(variables.jpypetest.fast, 'true')
+  condition: eq(variables['jpypetest.fast'], 'true')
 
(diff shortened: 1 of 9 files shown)


marscher:
I'm really sorry for this stupid typo mistake in the CI. I hope it didn't cause too much trouble. Luckily you executed most tests locally during development. Thank you very much for this quick micro release.

marscher (approved):

PR #925, opened 2021-01-14
TypeError when the platform was not recognized

krowvin:
Script crashed because None on line 113 is not a String and cannot be used with os.path.join(str, str, str)

Suggest changing string to empty string or other 

Full error:
Your platform 'sunos5' is not being handled explicitly. It may work or not!
  File "setup.py", line 45, in <module>
    Path('native', 'embedded', '*.cpp')], platform=platform,
  File "REDACTED/jpype-master/setupext/platform.py", line 123, in Platform
    distutils.log.info("Add JNI directory %s" % os.path.join(java_home, 'include', jni_md_platform))
  File "REDACTED/lib/python3.6/posixpath.py", line 92, in join
    genericpath._check_arg_types('join', a, *p)
  File "REDACTED/lib/python3.6/genericpath.py", line 149, in _check_arg_types
    (funcname, s.__class__.__name__)) from None
TypeError: join() argument must be str or bytes, not 'NoneType'

### DIFF
--- a/setupext/platform.py
+++ b/setupext/platform.py
@@ -110,7 +110,7 @@ def Platform(include_dirs=None, sources=None, platform=sys.platform):
         jni_md_platform = 'zos'
 
     else:
-        jni_md_platform = None
+        jni_md_platform = ''
         distutils.log.warn("Your platform '%s' is not being handled explicitly."
                            " It may work or not!", platform)
 
--- a/setupext/platform.py
+++ b/setupext/platform.py
@@ -109,7 +109,7 @@ def Platform(include_dirs=None, sources=None, platform=sys.platform):
         distutils.log.info("Add zos settings")
         jni_md_platform = 'zos'
 
-    elif platform == 'solaris':
+    elif platform == 'sunos5':
         distutils.log.info("Add solaris settings")
         jni_md_platform = 'solaris'
         
--- a/setupext/platform.py
+++ b/setupext/platform.py
@@ -109,6 +109,10 @@ def Platform(include_dirs=None, sources=None, platform=sys.platform):
         distutils.log.info("Add zos settings")
         jni_md_platform = 'zos'
 
+    elif platform == 'solaris':
+        distutils.log.info("Add solaris settings")
+        jni_md_platform = 'solaris'
+        
     else:
         jni_md_platform = ''
         distutils.log.warn("Your platform '%s' is not being handled explicitly."


Thrameos:
Shouldn't you also add a 'sunos5' to the platform settings so that the platform is covered properly?   

@marscher I am not sure if it was a feature or a bug that unrecognized platforms force you to define the new platform as opposed to attempting to run without a platform definition.   Crashing on a bad type is not the right behavior, but should we not just have it stop and perhaps have a "--platform" option to setup.py if the user desires to force another platforms configuration.

krowvin:
Going off of what Thrameos is suggesting, adding something like:

```
elif platform.startswith('sunos'):
        distutils.log.info("Add sunos settings")
        jni_md_platform = 'sunos'
```

Would solve my case (It won't always be sunos5). 

However, had there not been a type error the program would have continued on fine. If you want the program to crash here without continuing you could throw a custom exception? I thought this platform was merely informational. Is it not?

I went into the file manually and changed the None to the suggested edit I made '', and the build continued.

Thrameos:
I think it is more than informational.  It is supposed to be the name of the directory under JAVA_HOME/include which contains jni_md.h.  For current OpenJDK installations that file should be included in the C++ include path when building the CPython module.   

For example on my system which reports as linux the file is in $JAVA_HOME/include/linux

```
nelson85@wl-9768367:~/devel/open/jpype/setupext$ ls /usr/lib/jvm/java-11-openjdk-amd64/include
classfile_constants.h  jawt.h  jdwpTransport.h  jni.h  jvmti.h  jvmticmlr.h  linux/
                                                                                                                   ^^^^ 
```

I haven't seen a lot of installations of systems like ibm, sun, or hpux but those may have different directory structures.   If it compiles fine without the include your system likely lacks that directory so it compiled without the include.

marscher:
On some systems there are other type bindings for standard java types. So I'm happy to include such a change. Just setting jni_md_platform to None is indeed a mistake. We can safely set it to the empty string, so the later join will not fail.

marscher:
@krowvin please include your proposed platform patch as well.

krowvin:
I confirmed that it is "solaris" and not "sunos5" despite is printing sunos5 in other places. 
`$JAVA_HOME/include/solaris`
There is indeed a directory within the include called solaris. 

And i'm sure this would have been an issue in the end. I'm currently going through other compilation issues as this production server does not have gcc on it. (Will need to compile on another server most likely)

Added my changes for this tot he pull request.

marscher:
Thank you for your contribution!

PR #929, opened 2021-01-23
Handle weird error in windows with partially loaded classes.

### LINKED ISSUE #928: JVM crash on Windows when loading a class with (partially) unsatisfied dependencies
On Windows, when loading a class through JPype which has partially unsatisfied dependencies, e.g. in return types, the JVM crashes.

When loading the same class through the "java" binary, as well as with JPype on Linux, I get a NoClassDefFoundError as expected.

This situation can occur when Java projects are compiled with compile-time only dependencies (e.g. for unit tests) but still included in the final JARs. The classes concerned should lot be loadable (NoClassDefFoundError), but should not crash the JVM either.

Minimal example
```java
public class TestClass {
        System.out.println("Hi Freaks");
    }

    public static UnsatisfiedClass methodWithUnsatisfiedReturnType() {
        return null;
    }
}
```
```java
public class UnsatisfiedClass {
}
```

To build:
```
javac TestClass.java
rm UnsatisfiedClass.class
```

Running with "java" on Windows:
```
D:\jptest>java TestClass
Error: Unable to initialize main class TestClass
Caused by: java.lang.NoClassDefFoundError: UnsatisfiedClass
```

JPype 1.2.1 on Windows:
```
(base) D:\jptest>python
Python 3.7.3 (default, Apr 24 2019, 15:29:51) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> import jpype
>>> jpype.__version__
'1.2.1'
>>> jpype.getDefaultJVMPath()
'C:\\Program Files\\Java\\jdk-11.0.8\\bin\\server\\jvm.dll'
>>> jpype.startJVM(classpath='.')
>>> jpype.JClass('TestClass')
#
#
#  EXCEPTION_UNCAUGHT_CXX_EXCEPTION (0xe06d7363) at pc=0x00007ff9a3684048, pid=7912, tid=6524
#
# JRE version: Java(TM) SE Runtime Environment 18.9 (11.0.8+10) (build 11.0.8+10-LTS)
# Java VM: Java HotSpot(TM) 64-Bit Server VM 18.9 (11.0.8+10-LTS, mixed mode, tiered, compressed oops, g1 gc, windows-amd64)
# Problematic frame:
# C  [KERNELBASE.dll+0x14048]
#
# No core dump will be written. Minidumps are not enabled by default on client versions of Windows
#
# D:\jptest\hs_err_pid7912.log
#
#   https://bugreport.java.com/bugreport/crash.jsp
#
```
Full crash dump: [hs_err_pid7912.log](https://github.com/jpype-project/jpype/files/5855419/hs_err_pid7912.log)

JPype 1.2.1 on Linux (not exactly the same environment through):
```
(venv) [mihostet@... pyt]$ python
Python 3.6.5 (default, Jun 15 2019, 23:43:46) 
[GCC 7.3.0] on linux
>>> import jpype
>>> jpype.startJVM(classpath='.')
>>> jpype.JClass('TestClass')
  File "ClassLoader.java", line 522, in java.lang.ClassLoader.loadClass
  File "ClassLoaders.java", line 178, in jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass
  File "BuiltinClassLoader.java", line 581, in jdk.internal.loader.BuiltinClassLoader.loadClass
java.lang.ClassNotFoundException: java.lang.ClassNotFoundException: UnsatisfiedClass

  File "org.jpype.JPypeContext.java", line -1, in org.jpype.JPypeContext.newWrapper
  File "org.jpype.manager.TypeFactoryNative.java", line -2, in org.jpype.manager.TypeFactoryNative.newWrapper
  File "org.jpype.manager.TypeManager.java", line -1, in org.jpype.manager.TypeManager.populateMembers
  File "org.jpype.manager.TypeManager.java", line -1, in org.jpype.manager.TypeManager.createMembers
  File "org.jpype.manager.TypeManager.java", line -1, in org.jpype.manager.TypeManager.createMethodDispatches
  File "Class.java", line 1904, in java.lang.Class.getMethods
  File "Class.java", line 3191, in java.lang.Class.privateGetPublicMethods
  File "Class.java", line 3166, in java.lang.Class.privateGetDeclaredMethods
  File "Class.java", line -2, in java.lang.Class.getDeclaredMethods0
Exception: Java Exception

  File "/afs/cern.ch/user/m/mihostet/tmp/pyt/venv/lib/python3.6/site-packages/jpype/_jclass.py", line 99, in __new__
    return _jpype._getClass(jc)
java.lang.NoClassDefFoundError: java.lang.NoClassDefFoundError: UnsatisfiedClass
```

Thrameos:
This PR deals with the unusual case in which a class is loaded but we can't not enumerate methods.   This corner case can come up in certain situations such as exposed Testing class dependencies.   Unfortunately, the exception happened when we already had some resources created leading to a crash in windows.   Preemptively checking resources prior to starting the class creation process appears to clean it up.

Unfortunately this is another one of those cases that requires a special jar file to execute properly.  I debated whether this was worth a test given this is a horrible corner case that is unlikely to happen again, but decided that safety is the best course of action.

Fixed #928


native/java/org/jpype/manager/TypeManager.java
```diff
@@ -397,6 +397,12 @@ private ClassDescriptor createClass(Class<?> cls, boolean special)
 
   private ClassDescriptor createOrdinaryClass(Class<?> cls, boolean special, boolean bases)
   {
+    // Verify the class will be loadable prior to creating the class.
+    // If we fail to do this then the class may end up crashing later when the
+    // members get populated which could leave us in a bad state.
+    cls.getMethods();
```
marscher, line 403: So this will raise a Java exception, if the methods dependencies are not resolvable?

Thrameos: Yes that is Java's behavior though you don't really see that from the documentation.   Apparently Java does a lazy load of the methods without resolving them.  When you attempt to traverse the methods it generates a runtime exception.

https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getMethods--

Unfortunately there is no way to traverse a class and find only those methods that are linked.   I really don't remember this behavior from earlier versions of Java.  Perhaps this was introduced in Java 9 when they did the module change.

Thrameos: I should likely also add a call to ``getDeclaredMethods`` as there are methods that appear on the declared list and do not appear on the ``getMethods`` list.   So there may be another way to fail if the protected method is also unresolvable.   

There is also the mystery of why for one database driver it fails the first time and succeeds the second.   I am worried that we some major library will have a test class unresolvable and somehow we won't work with it.   Of course if that were the case than nothing that operates on reflection would work.  But this is really trending on being a Java bug.   It is not clear that Java ``getMethods`` is supposed to fail like this, but it does.

marscher: Eventually the exception thrown in getMethods is an implementation detail we cannot rely upon? Did you check if later major versions of the JDK do declare the throw?
I absolutely agree upon calling getDeclaredMethods as well, as it will add more failure safety. In the past JPype also allowed to call private/protected methods, so resolving their dependencies as well should be a good thing. I wouldn't consider it good practice of a library, if its classes have private implementation testing methods. So I'm not sure, how common this pattern would be.

marscher: Constructors could contain unresolved deps as well, right?

Thrameos: Well so far based on my testing I have seen ClassNotDefError and ClassNotFoundException both of which would be runtime errors or worse.   So it is not really unexpected that they didn't get a throws declaration.   Throws declarations are usually reserved for checked exceptions though unchecked can declared for documentation.  I am guessing that the Java system wasn't really intended to have classes that did not fully resolve.   The problem here is that rather than returning a list that we could iterate over and skip unresolved methods, they defined these as an array so it is all or none.   Later caching operations to speed up the JVM left us in the state we have.

I agree that we could just check getDeclaredMethods, getDeclaredFields, and getDeclaredConstructors.  In theory because we always probe the super class first this should cover all the options.

I do have some question on how to resolve this.   One approach would be fall back to the base class rather than failing.   If some internal class like DrillImpl can't be loaded because it was supposed to be a private implementation class and method we are returning is "Driver" which we can do automatically then there is another option.   JPype misbehaves a bit in that it   automatically wraps to the most derived class even if that class should be unavailable due to the module system.   I use this feature often in my work, but there is no requirement that we must do that when a class we are importing wasn't meant for accessing.   I think it would be possible to return the super rather than dying.   But then you can get silent failures, so I am not sure if that is a good idea.

marscher: I'd not skip unresolved methods, as API users would be confused of partially loaded classes (mismatching to documentation) -- a clear error message would be preferable then.

If wrapping super classes, instead of impl classes, that sounds like a good idea.
Why would we want to access things, which are not intended to be (of course it is cool and we can hack around)? I remember we had a similar discussion, when talking about respecting the visibility of private/public attributes.

Which silent failures could this cause? I can only think of abstract methods (just raising an exception as they should be overridden by children).

Thrameos: Lets consider the concequences.  We have a return of type X but the actual type is Y.  We try to build a wrapper for type Y but get an exception so we instead return wrapper X.  In Java this isnt an issue as the static type is X and they would have to cast, but in Python there is only the dynamic type Y that is expected.

If X is ``Driver`` and Y is ``DrillImpl`` then all makes sense because X is a useable class.  Where is gets ugly is what is X is Object?  Now it gets ugly.  Does the user have to cast or do we go to the super type?  What if the supertype is Object as all the class has is an Interface?  What if it has multiple interfaces?

The problem is that Java public classes no longer mean public.  Many times implementation classes are public just so they can be created from the package due to the lack of a friend concept.  Thus we cant tell is Y is really a public class or a private class which is public due to module access.

Returning the static type all the time is not an option as operating on a list would be a pain and very unpythonic.

It seems like an error is the safest, but there is no way to avoid the error and just get the base wrapper.  Thus we end up stuck as bad jars become completely unusable.

This is a hard problem.  No option is particularly good.

Thrameos: As far as abstract classes.  If you get the abstract class then you can access any member declared in the abstract even if not defined.  Those method still come up in getMethods and we always do virtual lookup.  It is no different than interfaces.   You  can test this by casting a wrapper to its abstract base class and using dir and calling the methods.   Last I checked that functionality all worked.

PR #965, opened 2021-04-29
Arrays generated wrong array type

### LINKED ISSUE #944: Crash on array operations
The following code was found to crash jpype version 1.2.1 under linux with open JDK 11.   It is not clear if it is simply memory exhaustion, a reference counting issue, or a memory overrun in an array operation.  The error log indicates some type of garbage collection issue.

```
import jpype
import numpy as np

jpype.startJVM()

gamma = jpype.JDouble[:](np.array(range(0,256)))
for i in range(100000):
    gamma2 = jpype.JDouble[512]
    gamma2[0:256] = gamma
    gamma2[256:] = 0.5*np.array(gamma)
```

Thrameos:
This PR corrects a bug in the generation of arrays.   When ``jpype.JDouble[10]`` was call it produced an instance of ``double[10][]`` rather than ``double[10]`` as intended.   Oddly the bad wrapper passed most tests because the wrapper class was set to the type for ``double[]``.  It only failed because the memory was becoming corrupt when JNI calls on the bad wrapper eventually trigger a fault.   

I finally figured out when I passed the object to a field and the resulting field type came back with a different array type.   I can add tests for this bug but it is pretty obscure because the JNI is trusting on these memory accesses and thus there is no way to guarantee a failure.

Fixes #944

### DIFF
--- a/.azure/build.yml
+++ b/.azure/build.yml
@@ -84,6 +84,7 @@ jobs:
       mac-3.9:
         imageName: "macos-10.14"
         python.version: '3.9'
+        jpypetest.fast: 'true'
 
   pool:
     vmImage: $(imageName)
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,10 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 Latest Changes:
 - **1.2.2_dev0 - 2021-01-03**
+  
+  - Fixed a bug with arrays created using the short cut.  The wrong type
+    was being returned.
+
 - **1.2.1 - 2021-01-02**
 
   - Missing stub files added.
--- a/native/python/pyjp_class.cpp
+++ b/native/python/pyjp_class.cpp
@@ -712,10 +712,9 @@ static PyObject *PyJPClass_array(PyJPClass *self, PyObject *item)
 	if (PyIndex_Check(item))
 	{
 		long sz = PyLong_AsLong(item);
-		JPClass *cls = self->m_Class->newArrayType(frame, 1);
-		jvalue v;
-		v.l = (jobject) cls->newArrayOf(frame, sz);
-		return cls->convertToPythonObject(frame, v, true).keep();
+		JPArrayClass *cls = (JPArrayClass*) self->m_Class->newArrayType(frame, 1);
+		JPValue v = cls->newArray(frame, sz);
+		return cls->convertToPythonObject(frame, v.getValue(), true).keep();
 	}
 
 	if (PySlice_Check(item))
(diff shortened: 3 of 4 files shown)


marscher (approved):
So the conversion failed because of the binding to jvalue.l instead of JPValue? Could you please  elaborate a bit, what this different wrapping implies.

Anyways, thanks for fixing this!

Thrameos:
Not exactly.  The problem was calling ``newArrayOf`` vs ``newArray``.   The first call gets the requested type so ``jpype.JDouble[:]`` would give ``double[]``.   But ``newArrayOf`` adds another layer of dereferencing so we get ``double[][]``.   Here is where things went wrong.   We then call ``double[]->createPythonWrapper(double[][])``  because here is no checks on the type of object to be wrapped it packs the wrong object into the wrapper.  We could call ``isinstance()`` but that is a big performance hit because we have to marshal across the JNI boundary.  99% of the time this isn't an issue because we are wrapping types where the wrapper was determined by Java on a return, but in this incorrect call we have a forced mismatch.   

After this point things get weird.   The object and the class are mismatched so it should fail immediately when we ask to setting an array element  to a ``double`` when the element type is ``double[]`` or map it to memory.  Instead it succeeds and even performs the correct action.  It even accepts passing the wrong wrapper in to set a field of type ``double[]`` and Java seems to be okay with it for array accesses and the rest, so if you create a bad wrapper and pass it to Java it acts normally.   Because our type matcher is using the type we say rather than the actual Java class everything appears fine for those tests as well.  My conclusion is every call we were testing turned out to be unchecked by JNI and thus it could evade detection even by some fairly elaborate tests.  Eventually something in the memory management system will bust when the corrupt memory gets garbage collected as a double can't be cast to an object, but up until that point everything works fine.

Arrays in the JVM are sort of half way objects.   While certain protections exist that cause casting errors with real objects, it appears that arrays lack that safety feature.  They appear to be objects with an arbitrary scratch area appended to the end.   I can see this exact sort of error happening within the JVM implementation and being very hard to track down as the result is so unexpected.

Bottom line: I hit the bug on some fairly complex code that was able to tease out the differences between it enough to locate the source of the issue.   The earlier report just don't help because as a memory issue the origin of the fault was very distant.  I boiled it down to a very short test case, and even with that I had to read the code ten times because it wasn't obvious to me that ``newArrayOf`` was the incorrect call.  So no fault on the original review both the function and tests all look completely reasonable.

About the only way to prevent this completely in the future would be to make the instrumented diagnostics mode compilation with an extra safety check where ``createPythonWrapper`` double checks the object type before accepting it and fails otherwise.  That would have made a test fail in the coverage stage.  But it would be a lot of work (20 methods) for an error that which is fairly unlikely to get repeated.  My work load has been pretty high recently so I am not getting much time on this project.

PR #969, opened 2021-05-10
Remove line in terminate.

Thrameos:
Remove a line that is not valid after removing the freeResources was removed.

### DIFF
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -347,8 +347,6 @@ def _JTerminate():
     try:
         import jpype.config
         # We are exiting anyway so no need to free resources
-        if _jpype.isStarted():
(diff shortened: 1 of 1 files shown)


jpype/_core.py
```diff
@@ -347,8 +347,6 @@ def _JTerminate():
     try:
         import jpype.config
         # We are exiting anyway so no need to free resources
```
ShaheedHaque, line 349: Is

ShaheedHaque (commented):
Isn't the comment now superfluous?

jpype/_core.py
ShaheedHaque, line 349: Isn't the comment now superfluous?

PR #976, opened 2021-05-13
Additional diagnostics

### LINKED ISSUE #974: Improve error message
Per https://stackoverflow.com/a/66529772/14731 the error message found at https://github.com/jpype-project/jpype/blob/24a2b95aefc2a59e7cf2c362a92ee5c6b13eb94f/jpype/imports.py#L201 is misleading. Consider clarifying that the class must be public but was X instead.

Thrameos:
Fixes #974


marscher (approved):
Thanks for making it more fail safe.

PR #982, opened 2021-05-30
Changes needed for Python 3.10 GC changes.

Thrameos:
This PR contains a series of changes required to operate on Python 3.10.   Some changes in the GC system imposed additional requirements which were not satisfied.   I intend this for the 1.3 release, but the required changes warrant review before inclusion.  Once this is approved I can continue with the release.

This PR may need additional updates if it fails the CI.   Some changes for 3.10 may be incompatible with older Python versions. 

One reference counting problem was discovered with the PyFrame objects that this fixes.   The rest are changes to fix the GC.   We had one under count on a reference which required a manual increment.   It seems like there may be a bug in Python regarding importing modules.

### DIFF
--- a/native/python/include/pyjp.h
+++ b/native/python/include/pyjp.h
@@ -158,6 +158,8 @@ PyObject  *PyJPClass_FromSpecWithBases(PyType_Spec *spec, PyObject *bases);
 PyObject  *PyJPValue_alloc(PyTypeObject* type, Py_ssize_t nitems );
 void       PyJPValue_free(void* obj);
 void       PyJPValue_finalize(void* obj);
+int        PyJPValue_traverse(PyObject *self, visitproc visit, void *arg);
+int        PyJPValue_clear(PyObject *self);
 
 // Generic methods that operate on any object with a Java slot
 PyObject  *PyJPValue_str(PyObject* self);
--- a/native/python/include/pyjp.h
+++ b/native/python/include/pyjp.h
@@ -50,6 +50,13 @@ extern "C"
 {
 #endif
 
+// Needed to write common code with older versions
+#if PY_VERSION_HEX<0x03080000
+// Introduced in Python 3.8
+#define Py_TRASHCAN_BEGIN(X, Y)
+#define Py_TRASHCAN_END
+#endif
+
 PyMODINIT_FUNC PyInit__jpype();
 
 /**
--- a/native/python/include/pyjp.h
+++ b/native/python/include/pyjp.h
@@ -51,7 +51,7 @@ extern "C"
 #endif
 
 // Needed to write common code with older versions
-#if PY_VERSION_HEX<0x03080000
+#ifndef Py_TRASHCAN_BEGIN
 // Introduced in Python 3.8
 #define Py_TRASHCAN_BEGIN(X, Y)
 #define Py_TRASHCAN_END
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -6,6 +6,8 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 Latest Changes:
 - **1.2.2_dev0 - 2021-01-03**
 
+  - Fixes for memory issues found when upgrading to Python 3.10 beta.
+
   - Add additional diagnositics for importing of non-public class.
 
   - Fixed issue with classes with unsatified dependencies leading to a crash
--- a/native/common/jp_context.cpp
+++ b/native/common/jp_context.cpp
@@ -273,11 +273,6 @@ void JPContext::initializeResources(JNIEnv* env, bool interrupt)
 	jmethodID startMethod = frame.GetStaticMethodID(contextClass, "createContext",
 			"(JLjava/lang/ClassLoader;Ljava/lang/String;Z)Lorg/jpype/JPypeContext;");
 
-	// Find the native library
-	JPPyObject import = JPPyObject::call(PyImport_AddModule("importlib.util"));
-	JPPyObject jpype = JPPyObject::call(PyObject_CallMethod(import.get(), "find_spec", "s", "_jpype"));
-	JPPyObject origin = JPPyObject::call(PyObject_GetAttrString(jpype.get(), "origin"));
-
 	// Launch
 	jvalue val[4];
 	val[0].j = (jlong) this;
@@ -287,10 +282,14 @@ void JPContext::initializeResources(JNIEnv* env, bool interrupt)
 
 	if (!m_Embedded)
 	{
-		PyObject *import = PyImport_AddModule("importlib.util");
-		JPPyObject jpype = JPPyObject::call(PyObject_CallMethod(import, "find_spec", "s", "_jpype"));
+		JPPyObject import = JPPyObject::call(PyImport_AddModule("importlib.util"));
+		JPPyObject jpype = JPPyObject::call(PyObject_CallMethod(import.get(), "find_spec", "s", "_jpype"));
 		JPPyObject origin = JPPyObject::call(PyObject_GetAttrString(jpype.get(), "origin"));
 		val[2].l = frame.fromStringUTF8(JPPyString::asStringUTF8(origin.get()));
+		import.incref();  // The documentation specifies that PyImport_AddModule must return a 
+	                  // new reference, but that is not happening in Python 3.10
+	                  // so we are triggering a gc assertion failure.  To prevent
+	                  // the failure manually up the reference counter here.
 	}
 	m_JavaContext = JPObjectRef(frame, frame.CallStaticObjectMethodA(contextClass, startMethod, val));
 
--- a/native/common/jp_exception.cpp
+++ b/native/common/jp_exception.cpp
@@ -528,7 +528,6 @@ PyTracebackObject *tb_create(
 	frame->f_trace = NULL;
 	frame->f_valuestack = NULL;
 	frame->f_localsplus[0] = NULL;
-	frame->f_stackdepth = 0;
 
 	// Allow GC on the frame
 	PyObject_GC_Track(frame);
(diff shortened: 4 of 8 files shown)


native/python/include/pyjp.h
```diff
@@ -50,6 +50,13 @@ extern "C"
 {
 #endif
 
+// Needed to write common code with older versions
+#if PY_VERSION_HEX<0x03080000
+// Introduced in Python 3.8
+#define Py_TRASHCAN_BEGIN(X, Y)
```
marscher, line 56: Can you please tell, why you override this mechanism? Is this the trouble maker for 3.10?  The code here https://github.com/python/cpython/blob/main/Include/cpython/object.h#L461 says, that this safety routine got introduced in 3.10, but not 3.8.

marscher:
Thanks for taking care about this. I guess these bugs were hard to squish. I guess you had good reasons to disable the newly introduced GC stack overflow guard.

native/python/include/pyjp.h
Thrameos: I think you misunderstand the code.  The intent of the lines is to define the GC stack overflow guard for version 3.6 and 3.7 where those symbols were not defined so that I don't have to place version guards all over the place when I add the proper TRASHCAN handling in JPype.  They appeared in 3.8 but for some reason I did not trigger a bug until 3.10 when further GC changes changed a harmless error into a segmentation fault which is forcing me to include it.  

So as I read it the code says "if the Python version is before 3.8 then introduce the GC stack overflow as empty".  Did I write it incorrectly?

I may have to disable the trashcan in 3.8 and 3.9 if it is a problem but that would be ``#undef`` statements.

marscher:
Eventually I didn't read it correctly (the body after define). Without a definition, there is nothing happening. This wouldn't cause any trouble in these Python versions, as long as these new GC changes won't be back ported.
Forgive me, if I'm all wrong, it's very late here and I'm already tired..

On May 30, 2021 11:36:07 PM UTC, Karl Nelson ***@***.***> wrote:
>@Thrameos commented on this pull request.
>
>
>
>> @@ -50,6 +50,13 @@ extern "C"
> {
> #endif
> 
>+// Needed to write common code with older versions
>+#if PY_VERSION_HEX<0x03080000
>+// Introduced in Python 3.8
>+#define Py_TRASHCAN_BEGIN(X, Y)
>
>I think you misunderstand the code.  The intent of the lines is to
>define the GC stack overflow guard for version 3.6 and 3.7 where those
>symbols were not defined so that I don't have to place version guards
>all over the place when I add the proper TRASHCAN handling in JPype. 
>They appeared in 3.8 but for some reason I did not trigger a bug until
>3.10 when further GC changes changed a harmless error into a
>segmentation fault which is forcing me to include it.  
>
>So as I read it the code says "if the Python version is before 3.8 then
>introduce the GC stack overflow as empty".  Did I write it incorrectly?
>
>I may have to disable the trashcan in 3.8 and 3.9 if it is a problem
>but that would be ``#undef`` statements.
>
>
>-- 
>https://github.com/jpype-project/jpype/pull/982#discussion_r642144194

-- 

Thrameos:
No problem. 

Sorry for any confusion.  I didn't explain it very clearly.  3.10 has introduced 4 separate segmentation faults when I hit it in testing.

-  PyJPMethod was causing issues with its creation policy.   I may have to remove the test of whether JMethod is derived from Function as there is clearly an issue going on with our hack.
- The GC system added the requirements that all GC classes must have a traverse and clear which we did not.  They do not have a NULL check in the GC so that was creating a segfault on GC_TRACK command.  It is a documented requirement so I don't expect it to change when 3.10 is released.
- We had a reference counting bug on the stack trace we added.  The requirement appears to be that a frame references.   The original code in Python I modeled on did not.   Unfortunately this could be a change of requirements just like heap type referencing where they added a requirement by fixing a bug that requires us to test to see if the reference count has increased in 3.7.2.  
- The reference counting requirement changes from `new reference` to `borrowed reference` which broke some of our code.  Worst case we end up leaking `importlib.utils` in old versions, but that can only happen once on startup so not fatal.
-
(My usual procedure it to run our pytest test bench under gdb with python-debug version.  If the error message is too much of mess I then run each pytest script separately to see if I can get a clear picture of the symptoms which manifest in each script hoping to be able to boil it down to a short fatal script.   Often the error starts in one test and then crashes in another so running each script to puzzle out which test script has the least steps to replicate make it easier.  The method works great assuming there is one error causing an issue.  But it has the downside that when there are actually multiple independent problems, I have to bash my head until I realize my assumptions were bad.)

Likely these were bugs in JPype that were just exposed by a new Python version, like the windows initialization bug.   In theory fixing those bugs should work fine in older versions, but it does not appear the case so I am going to have to wrap more code in ``#if`` statements so that JPype works both for old and new versions.   I mistakenly hit the review button assuming that it would be easy to get both working, but it looks like the changes in 3.10 were actually API breaking so I will have to debug several versions of Python to get this PR completed.

As these are all reference counting bugs they are about the worst possible to trace down as the fault point is deep in the GC collector far, far from where the original insult occurred.   And because there were 4 of them rather than a common cause I have been scratching my head without much progress for 2 weeks as I was looking for some way to instrument to find the source.   Also 3.10 broke a bunch of stuff in distutils which left me hard pressed to get the problems replicated in the first place.   I ended up having to build Python several times just to get the environment compatible.  (which is why I almost always avoid beta versions)  This one has been a real drag.  Of course if I can't fix it, noone else likely will be able to fix it either so I just have to knuckle down and get it done.

Unfortunately 2 builds of our release pattern which were introduced by a user were not tagged to a version so our release script is building 3.10 rather than 3.9 versions like intended which is what is sending me to make these fixes.  I had all this on the release branch that I was pushing out by once I looked at the scope of the changes I decide better to send it to a review.

Thrameos:
I am going to try a different approach with PyFrame.   The implementation is so different starting in 3.10 (and previously in pre 3.5) that the amount of code I am having to replicate from Python to make the frame is unwieldy.  Python did not give a method for creating frames externally but it can create a frame from a PyThreadState, PyCodeObject, and a globals dictionary.  Unfortunately we don't have a valid thread state as we are faking the objects.  So if I want to use only defined API I either have to mutate the existing thread state to set up my fake frames and then restore it after, or create an empty threadstate with the one required field and feed it in.    Obviously messing with the thread state has risks while faking it is just a speed concern (on a low usage error path).   

I am hoping this will make our code more portable in the future.  I tried with 3 different versions on my system, so hopefully it will clear the CI.

native/python/include/pyjp.h
Thrameos: Should I have defined this as 
```
#ifndef Py_TRASHCAN_BEGIN
#define Py_TRASHCAN_BEGIN(X, Y)
#define Py_TRASHCAN_END
#endif
```
instead, rather that using the Python version?   It would be safer if they decide to backport.

marscher:
Yes, that's what I meant.

On May 31, 2021 1:36:35 AM UTC, Karl Nelson ***@***.***> wrote:
>@Thrameos commented on this pull request.
>
>
>
>> @@ -50,6 +50,13 @@ extern "C"
> {
> #endif
> 
>+// Needed to write common code with older versions
>+#if PY_VERSION_HEX<0x03080000
>+// Introduced in Python 3.8
>+#define Py_TRASHCAN_BEGIN(X, Y)
>
>Should I have defined this as 
>```
>#ifndef Py_TRASHCAN_BEGIN
>#define Py_TRASHCAN_BEGIN(X, Y)
>#define Py_TRASHCAN_END
>#endif
>```
>instead, rather that using the Python version?   It would be safer if
>they decide to backport.
>
>
>-- 
>https://github.com/jpype-project/jpype/pull/982#discussion_r642162875

-- 

native/python/include/pyjp.h
marscher: Yes, exactly.

Thrameos:
Is this one okay to go?

Thrameos:
Also any thoughts regarding the ThreadState kludge?   

I implemented the dummy object rather than manipulating the current one, but I could try the alternative approach.   The issues with the dummy object is most fields are not filled out in the thread state (and are currently uninitialized) and they could potentially call almost anything in the thread state which may generate a seg fault in future versions.   

The alternative of manipulation of the current state, is that any Python calls can take that state as a hidden argument so altering it and then restoring it runs the risk that something like multithreading consults it between resulting in problems.

The other alternative is to dup the thread state in some fashion with memory copy that then alter the dup.   This is much more expensive and thus would have to be called once at the start of the creation process.   But it would also be doable.  

None of the options are great.  The lack of a safe API in Python is problematic and it would only take a short amount of time to patch Python to be much safer.   Taking a giant object just to get two fields (the frame reference, and starting in 3.10 indirectly accessing the builtin dict if not found in the globals) is not a very good design.   Unfortunately, my organization is still placing a hold on my Python work as they apparently can't sign the Python developer contract and have forbid me from signing.  I am still planning to stage a revolt but COVID restrictions have prevented me from organizing.   I won't be holding my breath.  The last time I had such a dispute it took close to 7 years to solve but that was with an external sponsor rather than simply the intellectual property department.

marscher:
I you open up PRs on Python in your free-time (with your personal/other Github account), what concerns would this super-useful department raise? Would/could they even relate this to your person?

Even if the operation of copying is expensive, it seems like the most fail-safe approach.

marscher (approved):
Until we have a proper public API offered by Python, this workaround is appreciated.

Thrameos:
Unfortunately, they have given me the "do not proceed".  I asked can I simply do this on my personal time and was informed that  I don't have such a thing under my contract unless I wish to sever my contract.  To be fair I hold a large number of patents and have to work with IPO often.   Unless overridden that decision will stand.

They understand that contributing to software that is used by many people in the organization, but they have directions from above that the terms in the Python contract are unacceptable.   The issue here is that it is a giant game of telephone.   Someone along the way has decide to hear some vague directive in a way that likely was not intended.   And unless I can go over or around I will remain stuck (and this is not the first time.)   Last time I met with the department lawyer and he basic told them to accept JPype as legitimate work that was not a conflict of interest, but I would need access to him to do such a thing again.   I have a large number of potential supporters that I need to meet face to face to come up with a reasoned solution.

PR #984, opened 2021-06-04
Documentation of sql shutdown issue

### LINKED ISSUE #983: sql shutdown
Don't know if this is a JPype issue, or one for HSQLDB (2.6.0). I'm a bit new to JPype, so I hope this isn't a triviality I missed in the docs.

This isn't about shutting down the JVM.

I needed a way to shut down an hsqldb server remotely. Server was started on its host with:

```
java -cp lib\hsqldb.jar org.hsqldb.server.Server --database.0 file:db;readonly=true --dbname.0 dbremote
```

Let's say for now the db host is on `192.168.2.100`. I connect to it in `python/JPype` with:

```
with jpype.dbapi2.connect("jdbc:hsqldb:hsql://192.168.2.110/dbremote") as cx:
    with cx.cursor() as cur:

    #
    # do some stuff with cur
    #
```

It works.

But let's say one wants to remotely shutdown that database through `python/JPype`, for *whatever* reason.

Usually one would do a `cur.execute("shutdown")` (or with `shutdown immediate`) within the above code. It *does work*, the remote database is shutdown... but a connection error is also given out, that can be traced back to the closing of the implicit prepared statement that JPype does in the background (`cur._statement.close()`).

I won't give the whole traceback, just the main initial cause:
```
  File "org.hsqldb.jdbc.JDBCPreparedStatement.java", line -1, in org.hsqldb.jdbc.JDBCPreparedStatement.close
  File "org.hsqldb.ClientConnection.java", line -1, in org.hsqldb.ClientConnection.execute
  File "org.hsqldb.error.Error.java", line -1, in org.hsqldb.error.Error.error
  File "org.hsqldb.error.Error.java", line -1, in org.hsqldb.error.Error.error
org.hsqldb.HsqlException: <exception str() failed>
```

I *guess* that when we come out of the cursor's `with`, when the "cleanup" is done, the prepared statement's `close()` method tries to communicate with the database for some maintenance, but isn't there anymore. Or something else I don't know.

Fortunately, JPype is flexible enough to let us have a workaround: rather than `cur.execute("shutdown")`, one can dig a bit more and use a simple (non prepared) statement:

```
# First, provoke a cleanup that will also set `cur._statement = None`,
# and prevent further call to `cur._statement.close()`
# (if no other prepared statement done)
cur._finish()

# Ultimately, avoid dbapi2's `execute`, and rather
# call hsqldb's `createStatement` and execute simple SQL
cur._jcx.createStatement().execute("shutdown")

# After the above, we still have `cur._statement = None`,
# and coming out of `with` won't trigger
# the problematic closing of `cur._statement`
```

Not certain if this is "clean" though... `_finish` is almost closing `cur` (`cur._closed` not put to `True`).
Perhaps better to reconnect to the db after the connection is closed just to do the shutdown. (?)

Is there a better way to do a non-prepared query statement?

The above is an edge case, probably not something one would normally do, but here it is anyway, in case someone else stumbles on the same problem.

Thrameos:
Fixes #983

@kysko Can you make sure this covers the problem so that I can close the issue?

### DIFF
--- a/doc/dbapi2.rst
+++ b/doc/dbapi2.rst
@@ -390,7 +390,7 @@ is then automatically closed will then produce an exception.
 This exception is due to a conflict between dbapi2, Java, and HSQLDB
 specifications.  Dbapi2 requires that statements be executed as prepared
 statements, Java requires that closing a statement yields no action if the
-connect is already closed, and HSQLBD sets the ``isValid`` to false but not
+connection is already closed, and HSQLBD sets the ``isValid`` to false but not
 ``isClosed``.  Thus executing a shutdown through dbapi2 would be expected to
 close the prepared statement on an invalid connection resulting in an error.
 
--- a/doc/dbapi2.rst
+++ b/doc/dbapi2.rst
@@ -376,7 +376,40 @@ so that the custom return type accurately reflects the column type.
           # Custom return converter here
           return rc
    JSON = JSONType("JSON")
-  
+
+
+Interactions with prepared statements
+-------------------------------------
+
+Certain calls can be problematic for dbapi2 depending on driver.  In
+particular, SQL calls which invalidate the state of the connection will issue
+an exception when the connection is used.   In HSQLDB the statement
+``cur.execute('shutdown')`` invalidates the connection which when the statement
+is then automatically closed will then produce an exception.
+
+This exception is due to a conflict between dbapi2, Java, and HSQLDB
+specifications.  Dbapi2 requires that statements be executed as prepared
+statements, Java requires that closing a statement yields no action if the
+connect is already closed, and HSQLBD sets the ``isValid`` to false but not
+``isClosed``.  Thus executing a shutdown through dbapi2 would be expected to
+close the prepared statement on an invalid connection resulting in an error.
+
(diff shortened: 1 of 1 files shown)


kysko:
Yes, this covers the case I exposed.
Thanks.

doc/dbapi2.rst
```diff
@@ -376,7 +376,40 @@ so that the custom return type accurately reflects the column type.
...
+Certain calls can be problematic for dbapi2 depending on driver.  In
+particular, SQL calls which invalidate the state of the connection will issue
+an exception when the connection is used.   In HSQLDB the statement
+``cur.execute('shutdown')`` invalidates the connection which when the statement
+is then automatically closed will then produce an exception.
+
+This exception is due to a conflict between dbapi2, Java, and HSQLDB
+specifications.  Dbapi2 requires that statements be executed as prepared
+statements, Java requires that closing a statement yields no action if the
+connect is already closed, and HSQLBD sets the ``isValid`` to false but not
```
marscher, line 393: if the connection is already closed,

marscher (approved):

PR #999, opened 2021-08-24
Docfix

### LINKED ISSUE #940: Does Jpype support with Java SpringBoot annotation classes defined in jar?
Currently, I am trying to use one of the helper java code which is written using SpringBoot framework with an annotation like @Component, @Autowired .

while calling those object method I am getting NullPointerException, Does anyone come across this type of issue?

### LINKED ISSUE #991: Allow Pass-by-reference semantics for arrays
Check out this MWE:

1. Java code that modifies an array entry:
```java
package com.foo;

public class ArrayModifier {
    public static void changeArray(int[] array) {
        array[0] = 42;
    }
}
```

2. Python code that calls `changeArray()`:
```python
# jpype setup code
# ...

from com.foo import ArrayModifier

array = [1, 2, 3]
ArrayModifier.changeArray(array)
print(array)
```

This prints `[1, 2, 3]` instead of the expected `[42, 2, 3]`, suggesting that the array has been passed by value, not by reference. Why is that, and is this intended? Is there any way to avoid this and replicate pass-by-reference semantics for arrays passed between Python and Java?

### LINKED ISSUE #998: Ambiguous conversion of Python lists into either Java arrays or Lists
I noticed that, if a java method has two overloads accepting an array and an iterable, respectively, passing a Python list (or any other sort of iterable) to that method produces an exception.

Consider for instance the class:
```java
class Foo {
  Y method(X[] args)
  Y method(Iterable<X> args) 
} 
```
this can be used from Python as follows:
```python
foo = Foo() 
foo.method([a1, a2, a3,...]) 
```

I would expect the second overload to be called in this case (meaning that the Python list is converted into a Java iterable by Jpype), yet an exception is raised instead.

Is this intended?

IMHO whenever the actual argument is a Python iterable, and an overload exists accepting an iterable, that overload should be preferred.

Thrameos:
This PR 

Fixes #998
Fixes #940
Fixes #991

- Corrects a formatting issue that was resulting in missing examples in the quick guide regarding the casting operator.
- Added section on casting operator to the user guide
- Added section on array operators to the user guide
- Added section on pass by reference to get back results from modified Java arrays.
- Added section on annotations to limitations

### DIFF
--- a/doc/quickguide.py
+++ b/doc/quickguide.py
@@ -43,14 +43,6 @@ def python(s):
 """ % s
 
 
-def generic(s):
-    return """
-.. code-block::
-
-    %s
-""" % s
-
-
 def entry(Desc=None, Java=None, Python=None, Notes=None):
     global footnotes, footnotecounter
     if not Java:
@@ -296,7 +288,7 @@ entry("Checking if Java object wrapper", None,
 # Casting
 entry("Casting to a specific type",
       java("BaseClass b = (BaseClass)myObject;"),
-      generic("b = (BaseClass) @ myObject"),
+      python("b = (BaseClass) @ myObject"),
       "Matmul(@) is used as the casting operator.")
 endSection()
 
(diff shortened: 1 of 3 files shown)


gciatto (approved):

marscher (approved):
Thanks for fixing these issues!

Issue #1003, opened 2021-09-22
Some driver do not support setAutoCommit


abij:
The autoCommit setting is not supported by some java drivers.

- I read Apache Drill does not support it.
- Also the Spark JDBC (for Databricks) ([simba-spark-driver](https://databricks.com/spark/jdbc-drivers-download)) I'm using does not support it.

I'm prep a simple try-except around this setting in a PR.

```python
# line ~ 455 in dbapi2.py
def __init__(self, jconnection, adapters, converters, setters, getters):
        self._jcx = jconnection
        # Required by PEP 249
        # https://www.python.org/dev/peps/pep-0249/#commit
         self._jcx.setAutoCommit(False)
```

abij:
The exception raised when trying to `jpype.dbapi2.connect(url + token, driver="com.simba.spark.jdbc.Driver")`, to have some context:

```
~\...\lib\site-packages\jpype\dbapi2.py in __init__(self, jconnection, adapters, converters, setters, getters)
    454         # https://www.python.org/dev/peps/pep-0249/#commit
    455         # try:
--> 456         self._jcx.setAutoCommit(False)
    457         # except Exception as ex:
    458         #     if ex.getClass().getSimpleName() != 'SQLFeatureNotSupportedException':

java.sql.SQLFeatureNotSupportedException: java.sql.SQLFeatureNotSupportedException: [Simba][JDBC](10220) Driver does not support this optional feature.
```

Thrameos:
Yeah the PEP requirement that autocommit must be set.   Trying to set it and failing if it does not seems reasonable.

PR #1004, opened 2021-09-22
Add try block around setting autoCommit

abij:
PR for #1003.


Thrameos:
This seems reasonable.  Thanks for the contribution.

PR #1012, opened 2021-10-17
java.lang.Class and arrays conversion hints

### LINKED ISSUE #1008: java.lang.Class JConversion is ignored when looking for overloads for a method
When a JConversion is registered for `java.lang.Class`, it is silently ignored

Example:
```java
package org.acme;

public class MyClass {
    public static String getClassSimpleName(Class<?> clazz) {
        return clazz.getSimpleName();
    }
}
```

```python
import jpype.imports
from jpype import _jcustomizer
from jpype import JClass

class ProxyClassInstance:
    def __init__(self, proxy):
        self.proxy = proxy


@_jcustomizer.JConversion("java.lang.Class", instanceof=ProxyClassInstance)
def _ProxyClassConvert(jcls, obj):
    return obj.proxy

pype.startJVM(classpath=['target/issue-reproducer-8.11.0.Final.jar'])

from org.acme import MyClass

out = MyClass.getClassSimpleName(JClass('java.lang.String'))
assert out == 'String'

out = MyClass.getClassSimpleName(ProxyClassInstance(JClass('java.lang.String')))
assert out == 'String'
```

Expected:
Terminated without output OR an Error is raised when the JConversion is defined saying "conversion to java.lang.Class is not allowed"

Actual:
```
TypeError: No matching overloads found for *static* org.acme.MyClass.getClassSimpleName(ProxyClassInstance), options are:
	public static java.lang.String org.acme.MyClass.getClassSimpleName(java.lang.Class)
```

JConversion does work for other types (like `java.util.Collection`)

Use Case:
OptaPlanner has a lot of interfaces that take classes in order to determine what Objects to send in constraint streams:
Ex:
```java
constraintFactory.from(Lesson.class)
    .join(Lesson.class, Joiners.equal(Lesson::getRoom))
    .penalize("Same Room", HardSoftScore.ONE_HARD)
```

OptaPy generates proxy Java classes for Python class, which then can be reference in the constraint stream. It is stored on the type itself, meaning it is possible to do an implicit conversion from type to Java class by referencing its java class attribute, allowing for the OptaPy code to look like this:

```python
constraint_factory.from_(Lesson) \
    .join(Lesson, Joiners.equal(lambda lesson: lesson.room) \
    .penalize("Same Room", HardSoftScore.ONE_HARD)
``` 

However, this issue is preventing such an implicit conversion being registered, forcing the python code to look like this instead:

```python
LessonClass = get_class(Lesson)

constraint_factory.from_(LessonClass) \
    .join(LessonClass, Joiners.equal(lambda lesson: lesson.room) \
    .penalize("Same Room", HardSoftScore.ONE_HARD)
``` 

Reproducer:
https://github.com/Christopher-Chianelli/issue-reproducer/tree/jpype-conversion-customization

Thrameos:
Fixes #1008

### DIFF
--- a/.azure/build.yml
+++ b/.azure/build.yml
@@ -22,19 +22,19 @@ variables:
 jobs:
 - job: Deps
   pool:
-    vmImage: "ubuntu-16.04"
+    vmImage: "ubuntu-latest"
   steps:
   - template: scripts/ivy.yml
 
 - job: Documentation
   pool:
-    vmImage: "ubuntu-16.04"
+    vmImage: "ubuntu-latest"
   steps:
   - template: scripts/documentation.yml
 
 - job: Coverage
   pool:
-    vmImage: "ubuntu-16.04"
+    vmImage: "ubuntu-latest"
   dependsOn: Deps
   steps:
   - template: scripts/deps.yml
@@ -42,7 +42,7 @@ jobs:
 
 - job: Tracing
   pool:
-    vmImage: "ubuntu-16.04"
+    vmImage: "ubuntu-latest"
   steps:
   - template: scripts/tracing.yml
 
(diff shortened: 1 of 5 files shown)


marscher (approved):

marscher:
Thank you!

PR #1017, opened 2021-11-26
fix - #1016: Make functional conversion check argument count

### LINKED ISSUE #1016: Ambiguous Overloads for functional interfaces with different argument counts
# Description

When a Java method have multiple overloads with functional interfaces, an Ambiguous Overloads Type Error is raised when attempting to call the method with a Python Function.

In particular, this is valid in Java:

```java
public class MyClass {
    public static String apply(Function<String, String> mapper) {
        return mapper.apply("1");
    }

    public static String apply(BiFunction<String, String> mapper) {
        return mapper.apply("1", "2");
    }

        System.out.println(MyClass.apply(number -> "Number " + number))
        // "Number 1"
        System.out.println(MyClass.apply((number1, number2) -> "Number " + number1 + " and " + number2))
        // "Number 1 and 2"
    }
}
```

But this is not valid in Python
```python
import jpype
import jpype.imports

jpype.startJVM(classpath=['...'])


def single_arg_function(arg):
    return f'Number {arg}'


def two_arg_function(arg1, arg2):
    return f'Number {arg1} and {arg2}'


from org.acme import MyClass
assert str(MyClass.apply(single_arg_function)) == 'Number 1'
assert str(MyClass.apply(two_arg_function)) == 'Number 1 and 2'
```

```
TypeError: Ambiguous overloads found for org.acme.MyClass.apply(function) between:
public static java.lang.String org.acme.MyClass.apply(java.util.function.BiFunction)
public static java.lang.String org.acme.MyClass.apply(java.util.function.Function)
```

# Workaround

If you know exactly the functional interfaces you are using, cast to proxies using `inspect.signature`

```python
def cast(python_function):
    parameters = inspect.signature(python_function).parameters
    arg_count = len(parameters)
    if arg_count == 1:
        return JObject(python_function, Function)
    elif arg_count == 2:
        return JObject(python_function, BiFunction)
    else:
        raise Exception(f'Unsupported number of arguments: {arg_count}')


assert str(MyClass.apply(cast(single_arg_function))) == 'Number 1'
assert str(MyClass.apply(cast(two_arg_function))) == 'Number 1 and 2'
```

# Use Case

Make it easier to use Java Libraries in Python. Getting an ambiguous type error confuse Python users unfamiliar with Java, potentially causing them to abandon trying to use the technology. There are a few places in OptaPlanner (particular `Joiners`) where there are multiple overloads, same number of args, but different Functional Interfaces (with different parameter counts). Thus, it possible to do something like `Joiners.lessThan(item -> item.id, otherItem -> otherItem.item.id)` and `Joiners.lessThan((item1, item2) -> item2.id, otherItem -> otherItem.item.id)`  because the functional interfaces have a different argument count. I would like Python users to have the same experience:  `Joiners.lessThan(lambda item: item.id, lambda otherItem: otherItem.item.id)` and `Joiners.lessThan(lambda item1, item2: item2.id, lambda otherItem: otherItem.item.id)`. Currently, to achieve this behaviour, I would need to create a custom Joiners class that uses the casting workaround posted above.

# Reproducer
https://github.com/Christopher-Chianelli/issue-reproducer/tree/jpype-ambiguous-types-for-functional-interfaces-with-different-argument-counts

# Expected Behaviour

If there are overloads with the same argument counts but have functional interfaces with different argument counts, and a function is passed, inspect the function parameter count and cast it to the functional interface that have the same number of arguments. If no interface have the same number of parameters, raise a No Overload found Error. This does not solve the case where the Functional interfaces have the same parameter count, where casting must be used (the same as Java).

Christopher-Chianelli:
Motivation:

For a Java class with multiple overloads, reducing required casting and reduced TypeError caused by Python function accepting a different argument count than a Java `@FunctionalInterface`.

```java
public class MyClass {
    public static String apply(Function map) {
        return "Function apply";
    }

    public static String apply(BiFunction map) {
        return "BiFunction apply";
    }
}
```

Overloads:
```python
import jpype
import jpype.imports
from jpype import JImplements, JOverride, JObject

jpype.startJVM(classpath=['...'])

from org.acme import MyClass

assert str(MyClass.apply(lambda a: a)) == "Function apply"
assert str(MyClass.apply(lambda a, b: a)) == "BiFunction apply"
```

Old Behaviour:
```
TypeError: Ambiguous overloads found for org.acme.MyClass.apply(function) between:
	public static java.lang.String org.acme.MyClass.apply(java.util.function.BiFunction)
	public static java.lang.String org.acme.MyClass.apply(java.util.function.Function)
```

New Behaviour:
```
>>>  str(MyClass.apply(lambda a: a))
'Function apply'
>>>  str(MyClass.apply(lambda a, b: a))
'BiFunction apply'
```

Bad Casts
```
assert str(MyClass.apply(JObject(lambda a: a, BiFunction))) == 'Function apply'
assert str(MyClass.apply(JObject(lambda a, b: a, Function))) == 'BiFunction apply'
```

Old Behaviour:
```
TypeError: <lambda>() takes 1 positional argument but 2 were given
```

New Behaviour:
```
TypeError: Unable to cast 'function' to java type 'java.util.function.BiFunction'
```

This commit modify functional conversion to check argument count:

1. If the Java functional interface have more parameters than
   the Python function and the Python function does not accept
   variable arguments, conversion type is none

2. If the Java functional interface have less parameters than
   the number of required positional parameters of the Python
   function, conversion type is none

3. Otherwise, conversion type is implict.

I tested on my issue-reproducer and I can confirm it fixes the issue. I added some tests, but I couldn't figure out how to run them locally.

Fixes #1016

### DIFF
--- a/native/common/include/jp_typemanager.h
+++ b/native/common/include/jp_typemanager.h
@@ -41,6 +41,7 @@ public:
 	JPClass* findClassForObject(jobject obj);
 	void populateMethod(void* method, jobject obj);
 	void populateMembers(JPClass* cls);
+    int interfaceParameterCount(JPClass* cls);
 
 private:
 	JPContext* m_Context;
@@ -50,6 +51,7 @@ private:
 	jmethodID m_FindClassForObject;
 	jmethodID m_PopulateMethod;
 	jmethodID m_PopulateMembers;
+    jmethodID m_InterfaceParameterCount;
 } ;
 
 #endif // _JPCLASS_H_
--- a/native/common/jp_functional.cpp
+++ b/native/common/jp_functional.cpp
@@ -41,64 +41,64 @@ public:
 		if (!PyCallable_Check(match.object))
 			return match.type = JPMatch::_none;
 
-        // Modified from https://stackoverflow.com/a/1117735
-        bool has_parameter_count = false;
-        // Get the __code__ attribute of the function, which contains details about the function
-        // Note: __code__ is func_code in Python 2.x
-        PyObject* function_code_object = PyObject_GetAttrString(match.object, "__code__");
-        if(function_code_object) {
-            // get the argument count
-            PyObject* parameter_count_obj = PyObject_GetAttrString(function_code_object, "co_argcount");
-            if(parameter_count_obj) {
-                has_parameter_count = true;
-                int optional_parameter_count = 0;
-                int flags = 0;
-                const int ACCEPT_VARGS_MASK = 0x04; // From https://docs.python.org/3/reference/datamodel.html
-                PyObject* flags_obj = PyObject_GetAttrString(function_code_object, "co_flags");
-                if (flags_obj) {
-                    flags = PyLong_AsLong(flags_obj);
-                    Py_DECREF(flags_obj);
-                }
-                PyObject* optional_parameter_tuple = PyObject_GetAttrString(match.object, "__defaults__");
-                if (optional_parameter_tuple && optional_parameter_tuple != Py_None) {
-                    optional_parameter_count = PyTuple_Size(optional_parameter_tuple);
-                    Py_DECREF(optional_parameter_tuple);
-                }
-                const int parameter_count = PyLong_AsLong(parameter_count_obj);
-                const int java_parameter_count = cls->getContext()->getTypeManager()->interfaceParameterCount(cls);
-                const bool is_varargs = (flags & ACCEPT_VARGS_MASK) == ACCEPT_VARGS_MASK;
+		// Modified from https://stackoverflow.com/a/1117735
+		bool has_parameter_count = false;
+		// Get the __code__ attribute of the function, which contains details about the function
+		// Note: __code__ is func_code in Python 2.x
+		PyObject* function_code_object = PyObject_GetAttrString(match.object, "__code__");
+		if(function_code_object) {
+			// get the argument count
+			PyObject* parameter_count_obj = PyObject_GetAttrString(function_code_object, "co_argcount");
+			if(parameter_count_obj) {
+				has_parameter_count = true;
+				int optional_parameter_count = 0;
+				int flags = 0;
+				const int ACCEPT_VARGS_MASK = 0x04; // From https://docs.python.org/3/reference/datamodel.html
+				PyObject* flags_obj = PyObject_GetAttrString(function_code_object, "co_flags");
+				if (flags_obj) {
+					flags = PyLong_AsLong(flags_obj);
+					Py_DECREF(flags_obj);
+				}
+				PyObject* optional_parameter_tuple = PyObject_GetAttrString(match.object, "__defaults__");
+				if (optional_parameter_tuple && optional_parameter_tuple != Py_None) {
+					optional_parameter_count = PyTuple_Size(optional_parameter_tuple);
+					Py_DECREF(optional_parameter_tuple);
+				}
+				const int parameter_count = PyLong_AsLong(parameter_count_obj);
+				const int java_parameter_count = cls->getContext()->getTypeManager()->interfaceParameterCount(cls);
+				const bool is_varargs = (flags & ACCEPT_VARGS_MASK) == ACCEPT_VARGS_MASK;
 
-                // def my_func(x, y=None) should be both a Function and a BiFunction
-                // i.e. the number of parameters accepted by the interface MUST
-                // 1. Be at most the maximum number of parameters accepted by the python function (parameter_count)
-                //    (Unless the function accept a variable number of arguments, then this restriction does not
-                //     apply).
-                // 2. Be at least the minumum number of parameters accepted by the python function
-                // (parameter_count - optional_parameter_count = number of required parameters).
-                // Notes:
-                // - keywords vargs does not remove restriction 1
-                // - keyword only arguments are not counted.
-                if ((!is_varargs && parameter_count < java_parameter_count) ||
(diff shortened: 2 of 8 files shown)


Christopher-Chianelli:
Was able to learn how to run the test suite from the GitHub checks. Failure reproduce locally using the test script, but unable to reproduce in a venv. Quite odd.

Christopher-Chianelli:
Note: the Java Method not having vargs does not disable minimum required arguments constraint; vargs in Java are passed as an array, thus for the following code:
```java
public class MyClass {
    public static String vargsApply(VargsFunction mapper) {
        return mapper.apply("1", "2", "3");
    }

    @FunctionalInterface
    public interface VargsFunction {
        String apply(String... args);
    }
}
```

```python
import jpype
import jpype.imports
from jpype import JImplements, JOverride, JObject

jpype.startJVM(classpath=['...'])

from org.acme import MyClass

assert str(MyClass.vargsApply(lambda a, b, c: a)) == "1"
```

causing the following error currently in JPype:
`TypeError: <lambda>() missing 2 required positional arguments: 'b' and 'c'`

the PR changes the error to:
```
TypeError: No matching overloads found for *static* org.acme.MyClass.vargsApply(function), options are:
	public static java.lang.String org.acme.MyClass.vargsApply(org.acme.MyClass$VargsFunction)
```

Christopher-Chianelli:
Figured out the issue: `PyObject_GetAttrString` sets the error indicator, which is then checked and raised in convert. Adding `PyErr_Clear();` before method return fixes the issue.

(There were several other issues in the test module, naming the interfaces not being annotated with `@FunctionalInterface` and wrong string content)

Christopher-Chianelli:
@Thrameos ping

Thrameos:
I am not opposed to this in theory, but there is a serious cost to doing a large amount of work in the check routine.   The most likely issue to induce bugs the type caching issue which may force this code to be on the conversion rather than the check side of the match function.   When JPype matches and overload it hashes the types of the arguments which bypasses the overload resolution.   Adding more logic on the match such that determining the overloads based on the number of arguments could trigger the cache and thus potentially trigger the cache to skip a better overload.

I am guessing that is what you are trying to bypass in the method dispatch.   

Maybe it would be better to clear the cache if the match is poor.
```
                // Anything better than explicit constitutes a hit on the cache
                if (bestMatch.m_Type > JPMatch::_explicit)
                        return true;
                else
                        // bad match so forget the overload.
                        bestMatch.m_Overload = 0; 
```

Christopher-Chianelli:
I did catch a cache bug early on (IIRC, it resolved the first overload, but not the second, but it been a while), hence the modification to `jp_methoddispatch.cpp` (which might of make it bypass the cache?).

I am guessing from "which may force this code to be on conversion rather than the check side of the match function" that `convert` can return a value/set something on JPMatch that makes it a `None` match (even if `matches` return a not None match type)?

I'll try reproducing the cache issue I encountered w/o the changes to `jp_methoddispatch.cpp` to get the details.

Christopher-Chianelli:
Was able to reproduce on tests just by reverting the method dispatch change; The cache causes an ambiguous resolution to be found between a prior overload + the actual overload (in the test, SingleArg is tested first, followed by a NoArgs; note the third case, TwoArgs, is not listed in the ambiguous resolution list). The full error is below:
```
____________ OverloadTestCase.testFunctionalInterfacesWithDefaults _____________

>   JP_PY_CATCH(NULL); // GCOVR_EXCL_LINE

native/python/pyjp_method.cpp:110: 

>   JP_TRACE_OUT;

native/common/jp_methoddispatch.cpp:204: 

>   JP_TRACE_OUT;

native/common/jp_methoddispatch.cpp:195: 

>   JP_RAISE(PyExc_TypeError, ss.str());
E   Exception: C++ Exception

native/common/jp_methoddispatch.cpp:149: Exception

self = <test_overloads.OverloadTestCase testMethod=testFunctionalInterfacesWithDefaults>

    def testFunctionalInterfacesWithDefaults(self):
        def my_fun(x, y=None):
            return 'my_fun'
    
        def my_fun_vargs(x, *vargs):
            return 'my_fun_vargs'
    
        def my_fun_kwargs(x, **kwargs):
            return 'my_fun_kwargs'
    
        def my_fun_kw(*, keyword_arg=None):
            return 'my_fun_kw'
    
        class my_fun_class:
            def __call__(self):
                return 'my_fun_class'
    
        test2 = self.__jp.Test2()
        self.assertRaisesRegex(
            TypeError, 'Ambiguous overloads found', test2.testFunctionalInterfaces, my_fun)
        self.assertRaisesRegex(
            TypeError, 'Ambiguous overloads found', test2.testFunctionalInterfaces, my_fun_vargs)
        self.assertRaisesRegex(
            TypeError, 'Ambiguous overloads found', test2.testFunctionalInterfaces, my_fun_class)
        self.assertEqual('SingleArg', test2.testFunctionalInterfaces(my_fun_kwargs))
>       self.assertEqual('NoArgs', test2.testFunctionalInterfaces(my_fun_kw))
E       TypeError: Ambiguous overloads found for jpype.overloads.Test2.testFunctionalInterfaces(function) between:
E       	public java.lang.String jpype.overloads.Test2.testFunctionalInterfaces(jpype.overloads.Test2$NoArg)
E       	public java.lang.String jpype.overloads.Test2.testFunctionalInterfaces(jpype.overloads.Test2$SingleArg)

test/jpypetest/test_overloads.py:225: TypeError
_______ OverloadTestCase.testFunctionalInterfacesWithDifferentSignatures _______

```

PR #1032, opened 2022-02-27
Fix for javadoc on Java14+

Thrameos:
Fix issue with test when using later versions of javadoc.


Thrameos:
Unfortunately given that javadoc has basically no standards for their output this sort of extra code will break once in a while.  I made the trivial fix that allows both versions so that the jpypetest works properly on both versions.   I am merging so that I have a release candidate with a working test bench.

PR #1033, opened 2022-02-27
Remove 3.5 from test matrix

Thrameos:
It appears that Python 3.5 is no longer supported on azure.  Removing from test matrix.


Thrameos:
Merging this so we have a working test bench.

PR #1046, opened 2022-04-11
Azure issues

Thrameos:
Trying to make sure all versions work.

### DIFF
--- a/.azure/build.yml
+++ b/.azure/build.yml
@@ -62,19 +62,25 @@ jobs:
       linux-3.9:
         imageName: "ubuntu-latest"
         python.version: '3.9'
+      linux-3.10:
+        imageName: "ubuntu-latest"
+        python.version: '3.10'
       windows-3.6:
-        imageName: "vs2017-win2016"
+        imageName: "windows-2019"
         python.version: '3.6'
       windows-3.7:
-        imageName: "vs2017-win2016"
+        imageName: "windows-2019"
         python.version: '3.7'
       windows-3.8:
-        imageName: "vs2017-win2016"
+        imageName: "windows-2019"
         python.version: '3.8'
-        #windows-3.9:
-        #imageName: "vs2017-win2016"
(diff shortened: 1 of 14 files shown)


Thrameos:
@pelson The issue with the broken test is something wrong is numpy not jpype on Windows.

```
import jpype
import numpy as np
jpype.startJVM()

u = [np.uint64(13085891031060623360)]
print("%.8f"%u[0])
print("%.8f"%np.float32(u[0]))
print("%.8f"%np.array(u, dtype=np.float32)[0])
print("%.8f"%np.array(jpype.JFloat[:](u), dtype=np.float64)[0])
```

Correct behavior:  (all conversion path yield the same result)
```
13085891031060623360.00000000
13085891514045825024.00000000
13085891514045825024.00000000
13085891514045825024.00000000
```

Windows behavior: (one of the paths is different)
```
13085891031060623360.00000000
13085890414534197248.00000000
13085891514045825024.00000000
13085891514045825024.00000000
```

I switched to allclose which will ignore this issue for now.

Thrameos:
@marscher This was a simple set of changes to fix the build system.  There were minor changes in the test to fix deprecation warnings.   That required changing from "np.bool, np.int, np.float" to "np.bool_, np.int_, np.float_".   There was also one change because numpy is broken on some conversions that requires the the answers are close rather than exact.    I am disabling 3.6 from the testing matrix to add 3.10.   3.6 had a number of hacks which will need to be cleaned up for now.   The main reason for dropping 3.6 is some of the latest packages required for testing on 3.10 were conflicting with 3.6 in one of the test runs.   I am sure that I could find some set of packages that worked for both 3.10 and 3.6, but this huge time sink that I don't have time to resolve currently for an old version of python.

As there was no source code changed (everything is on the test side) and maintaining a working CI state is needed for releases and PRs, I am going to merge this one without further review.  Hopefully this won't cause you any grief.  If is it a problem you can roll back.

PR #1052, opened 2022-04-23
Fixed jni header warning logic

astrelsky:
This should be pretty self explanatory. Simply moves the warning outside the loop.


Thrameos:
LGTM.

Thrameos (approved):

PR #1054, opened 2022-04-23
Replacement for #1017

### LINKED ISSUE #1051: TypeError: Unable to cast function/method to java type 'java.lang.Runnable'
```python
import jpype
import jpype.imports

class MyClass:

    def fun(self):
        print("this doesn't")

def fun():
    print('this works')

if __name__ == '__main__':
    jpype.startJVM()
    from java.lang import Runnable
    mc = MyClass
    cb = Runnable @ fun
    cb.run()
    cb = Runnable @ mc.fun
    cb.run()
```

I'm going to assume this is caused by #1017. This is not a problem in jpype 1.3.0.

Thrameos:
This implementation is aware for functions and methods when checking argument counts.  

Fixes #1051

### DIFF
--- a/native/common/jp_functional.cpp
+++ b/native/common/jp_functional.cpp
@@ -32,6 +32,7 @@ JPFunctional::~JPFunctional()
 {
 }
 
+
 class JPConversionFunctional : public JPConversion
 {
 public:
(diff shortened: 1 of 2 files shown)


Thrameos:
@astrelsky Please verify this corrects your issue.
@Christopher-Chianelli Please look over this and review to see if it is a reasonable replacement.

astrelsky:
> @astrelsky Please verify this corrects your issue. @Christopher-Chianelli Please look over this and review to see if it is a reasonable replacement.

Seems to work.

Christopher-Chianelli (approved):
From what I see, the main changes are:

- Using PyFunction_GetCode to get the code attribute of functions/methods (versus accessing it like a PyObject).
- Adding a case for bound functions (where the self argument is removed via the function "__get__" descriptor). This is done via the `PyMethod_Self(match.object)` check, which returns a not null instance if the method is bound.

Looks good, given tests pass and correct overloads are used when ran locally.

Thrameos:
@marscher  This one should be ready for review.  I likely should add a note about it in the changelog.

marscher (approved):

marscher:
Thanks for the provided fix and testing @Thrameos and @Christopher-Chianelli, @astrelsky

PR #1055, opened 2022-04-23
Fix jchar segfault

### LINKED ISSUE #1053: String concatenation with JChar causes segfault
When I attempt to use the `+` operator to concatenate a JChar and a Python string I get a seg fault:

``` python
In [1]: import jpype

In [2]: jpype.__version__
Out[2]: '1.3.0'

In [3]: jpype.startJVM()

In [4]: jpype.JChar("x")
Out[4]: 'x'

In [5]: jpype.JChar("x") + "test"
#
#
#  SIGSEGV (0xb) at pc=0x0000000000000000, pid=11156, tid=11156
#
# JRE version: OpenJDK Runtime Environment 21.9 (17.0.2+8) (build 17.0.2+8)
# Java VM: OpenJDK 64-Bit Server VM 21.9 (17.0.2+8, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# C  0x0000000000000000
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h" (or dumping to /home/jim/Projects/ITP/pythonprocessing/py5generator/core.11156)
#
# /home/jim/Projects/ITP/pythonprocessing/py5generator/hs_err_pid11156.log
#
#   https://bugzilla.redhat.com/enter_bug.cgi?product=Fedora&component=java-17-openjdk&version=34
#
Aborted (core dumped)
```

Reversing the order of the operands gives a `TypeError: cast of null pointer` exception.

The workaround in both cases is to use `str`:

``` python
In [5]: "test" + str(jpype.JChar("x"))
Out[5]: 'testx'

In [6]: str(jpype.JChar("x")) + "test"
Out[6]: 'xtest'
```

There may be a good reason why that operation is not allowed and using `str` is a simple enough workaround, but I wouldn't expect this to trigger a segfault.

My computer is using Fedora Linux, python 3.8.12, and OpenJDK Java 17

Thrameos:
JChar operators did not consider the case in which the order of arguments was reversed.  The add method called the wrong slot when adding to unicode variables.   Both problems are fixed.

Fixes #1053

### DIFF
--- a/native/python/pyjp_char.cpp
+++ b/native/python/pyjp_char.cpp
@@ -63,11 +63,17 @@ static int isNull(JPValue *javaSlot)
 	return 1;
 }
 
+static PyObject* notSupported()
+{
+	PyErr_SetString(PyExc_TypeError, "unsupported operation");
+	return 0;
+}
+
 static int assertNotNull(JPValue *javaSlot)
 {
 	if (!isNull(javaSlot))
 		return 0;
-	PyErr_SetString(PyExc_TypeError, "cast of null pointer");
+	PyErr_SetString(PyExc_TypeError, "jchar cast of null pointer");
 	return 1;
 }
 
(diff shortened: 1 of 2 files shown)


Thrameos:
@hx2A please verify the behaviors make sense.

hx2A:
@Thrameos Looks good to me. Your unit tests are very thorough. Thank you!

marscher (approved):
Thanks for fixing this!

PR #1056, opened 2022-04-23
Remove dunder methods from java.

### LINKED ISSUE #1050: Glorious Segfault
Reproduction is fairly straightforward. I already took the liberty of finding the cause which I ended up doing with the Ghidra debugger. Having 3 separate debuggers for each language didn't suffice because it ended up being due to an exception from an exception being sent to python that was detected and causing a `JPypeException` to be thrown before going back to python and late enough in c++ to not know where it came from.

To reproduce simply create an interface such as the following and then attempt to import it in python.

```java
public interface BadInterface {
    public void __del__();
}
```

Here is what happens when it the type is created.

https://github.com/jpype-project/jpype/blob/0bacb9e57003f74aec05029caf58e1c707882b9c/native/python/pyjp_class.cpp#L1191-L1193

The `PyJPClass_Type->tp_new` there ends up calling `PyJPClass_new` which then gets up to this point which sets the Python exception and returns `NULL`.
https://github.com/jpype-project/jpype/blob/0bacb9e57003f74aec05029caf58e1c707882b9c/native/python/pyjp_class.cpp#L102-L107
 This then get's detected in the `JP_PY_CHECK` macro in `JPPyObject::call` which causes the `JPypeException` to be thrown. What happens from there I have no idea and probably shouldn't jump in without the necessary kevlar.

Hopefully this makes enough sense to hit the ground running at least.

Thrameos:
If a Java method contains ``__`` then the resulting class will often have unexpected and buggy behavior as it will try to repoint Python internals to Java methods.   This PR removes any method which starts with double underscore from the class definition.   Double underscore is a Python convention for a dunder method and also violates the conventions for Java.  This may break code which was doing "very bad things."  We could rename those as something like ``j__del__`` or something if it seems necessary to try to support such an oddly named function.  But for now I removed the dangerous java defined dunder functions.  (Adding an underscore just made ``__del_`` cause breakage, so our normal conventions don't apply.)

Fixes #1050

@astrelsky  I am not sure what you are doing trying to declare a dunder method in Java.  This would normally be accomplished with a class customizer.   I was not able to replicate the segfault, but I can prevent it from reaching that branch.   If you have additional clarifications it would help.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -6,6 +6,10 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 Latest Changes:
 - **1.3.1_dev0 - 2021-06-05**
 
+  - Fixed issue when Java classes with dunder methods such as  ``__del__`` 
+    caused conflicts in Python type system.   Java method which match dunder 
+    patterns are longer translated to Python.
+
   - Fix issue with numpy arrays with no dimensions resulting in crash..
 
   - Support for user defined conversions for java.lang.Class and array types.
--- a/jpype/_jclass.py
+++ b/jpype/_jclass.py
@@ -124,7 +124,9 @@ def _jclassPre(name, bases, members):
         k2 = pysafe(k)
         if k2 != k:
             del members[k]
-            members[k2] = v
+            # Remove unmappable functions
+            if k2:
+                members[k2] = v
 
     # Apply customizers
     hints = _jcustomizer.getClassHints(name)
--- a/jpype/_pykeywords.py
+++ b/jpype/_pykeywords.py
@@ -28,6 +28,8 @@ _KEYWORDS = set((
 
 
 def pysafe(s):
+    if s.startswith("__"):
+        return None
     if s in _KEYWORDS:
         return s + "_"
     return s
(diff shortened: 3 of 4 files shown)


astrelsky:
I wasn't trying to do it myself. I was actually using stubgenj and iterating over every single package when I initially encountered it.

The whole reason I spent the time to debug it myself was when I saw what caused it I pretty much said "oh no, this may be doing some very bad things" as you said above. But it turned out to just be an uncaught exception.

astrelsky:
@Thrameos I'm afraid I don't think this actually solves the problem. The use of the dunder methods is already guarded as the exception is thrown when the finalizer conflict is detected. The problem is that this exception goes uncaught in some conditions.

The sample with `__del__` in the interface was only one cause of an uncaught exception causing a segfault on import. I used that as an example because it is super simple and easy to reproduce. I haven't tracked down the other ones yet nor do I know if they are actually any different. I provided a complete example yesterday in the open issue but am commenting here so that this isn't merged under the pretense of resolving it.

Thrameos:
Okay I will try to replicate it.

Thrameos:
@marscher Any chance you can review this one so I can ship it?   I can't replicate the orginally reported error, but there was an issue with dunder methods.   If you declare code with ``__del__`` specifically the class will fail because ``__del__`` is needed to clear the Java resources.   Dunder methods are not a thing in Java and leaving them available is just asking for a conflict.   Thus I removed them from the safe methods list.

marscher (approved):

Thrameos:
Excellent.  I will try again to get a release out Saturday.

PR #1070, opened 2022-05-24
Add OpenBSD support on setupext/platform.py

zeloff:
Tested on OpenBSD 7.1 with OpenJDK 11.


marscher (approved):

marscher:
thanks for updating the setup to support OpenBSD

PR #1092, opened 2022-09-16
jpype#1091: allow null characters in strings when convertStrings=True

### LINKED ISSUE #1091: convertStrings=True cut Java returned Strings with NULL bytes short
#  Summary

When `jpype.startJVM(convertStrings=True)` is used, returned Strings with NULL bytes are cut short.
This behavior is not observed when `jpype.startJVM(convertStrings=False)` is used.

## Example Reproducer

Code is available at: https://github.com/Christopher-Chianelli/issue-reproducer/tree/jpype-null-bytes-in-string

Java class:
```java
package org.acme;

public class MyClass {
    public static int getLength(String argument) {
        return argument.length();
    }

    public static String getArgument(String argument) {
        return argument;
    }
}
```

Python code
```python
import jpype
import jpype.imports

jpype.startJVM(classpath=[...], convertStrings=True)

from org.acme import MyClass

# Passing a string with NULL bytes to Java works
assert MyClass.getLength('\u0000') == 1

# Reading a returned string with NULL bytes from Java fails
out = MyClass.getArgument('String\u0000with\u0000NULL\u0000Bytes')
print(out)

assert out == 'String\u0000with\u0000NULL\u0000Bytes'
```

### Expected Behavior

`out == 'String\u0000with\u0000NULL\u0000Bytes'`

### Actual Behavior

`out == 'String'`

Christopher-Chianelli:
Previously, `PyUnicode_FromString(str.c_str())` was used when convertStrings=True. If the string contains null bytes, this will terminate the string early. To allow strings with null bytes to be returned, `PyUnicode_FromStringAndSize(str.c_str(), str.length())` is now used instead, which create a Unicode object from the pointer with the given length (instead of the position of the first NULL character), thus allowing null bytes in the string.

Fixes #1091

### DIFF
--- a/native/common/jp_stringtype.cpp
+++ b/native/common/jp_stringtype.cpp
@@ -47,7 +47,7 @@ JPPyObject JPStringType::convertToPythonObject(JPJavaFrame& frame, jvalue val, b
 		if (context->getConvertStrings())
 		{
 			string str = frame.toStringUTF8((jstring) (val.l));
-			return JPPyObject::call(PyUnicode_FromString(str.c_str()));
+			return JPPyObject::call(PyUnicode_FromStringAndSize(str.c_str(), str.length()));
 		}
 	}
 
--- a/test/harness/jpype/str/Test.java
+++ b/test/harness/jpype/str/Test.java
@@ -31,8 +31,12 @@ public class Test
     return "memberCall";
   }
 
-  public String callWithNullBytes() {
+  public static String callWithNullBytes() {
     return "call\0With\0Null\0Bytes";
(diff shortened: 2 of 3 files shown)


Thrameos:
You should add tests using a mixture of Nulls with international characters (16 and 32 byte encoded) to make sure that it properly handles all cases.

Christopher-Chianelli:
Done; included tests that test strings with null bytes that SHOULD be 16 byte encoded/ 32 byte encoded.

Thrameos:
LFTM

PR #1100, opened 2022-10-27
Cycle

Thrameos:
Post release merge for the next patch cycle.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ jpypeJar = Extension(name="org.jpype",
 
 setup(
     name='JPype1',
-    version='1.4.1_dev0',
+    version='1.4.1',
     description='A Python to Java bridge.',
     long_description=open('README.rst').read(),
     license='License :: OSI Approved :: Apache Software License',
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -79,7 +79,7 @@
     package_dir={'jpype': 'jpype', },
     package_data={'jpype': ['*.pyi']},
     install_requires=['typing_extensions ; python_version< "3.8"',
-        'packaging ; python_version< "3.10"'],
```
marscher, line 82: I suggest we also disallow Python 3.12 for the time being. Otherwise guess we would run into trouble someday

marscher (approved):

setup.py
marscher: The pattern was introduced here https://github.com/jpype-project/jpype/pull/1031/files
But we have just used it to depend on a minimum requirement. Do you think adding a constraint on the upper bound as well will introduce more trouble than it solves? If the C-API changes in 3.12 again and JPype won't compile any more, we are forced to do a new release, leaving a broken version on PyPI. If we pin the version, we will get issues raised by folks who want to use the new version.

PR #1104, opened 2022-10-31
[distribution] do not ship *.class files in test/ for source dist

### LINKED ISSUE #1101: source distribution includes test binaries (e.g. class/jar files)
I guess we do not want that. The code from MANIFEST.in (graft test/) is from 2014 and is obsolete IMO.

marscher:
fixes #1101

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,8 @@
 recursive-include native *.h *.xml *.java *.jar *.class
 recursive-include examples *
 include *.rst doc/* LICENSE
-# include everything under test/
+# include test/
 graft test
+prune test/classes/*
 graft setupext
 global-exclude *.pyc


marscher:
Thanks for noticing that, I wasn't aware. Is it safe to run this pipeline without actually releasing something we don't want to?

Thrameos:
Sure I trigger the release one on azure anytime I want to see if the release works.  It just makes a bunch of artifacts.  Since the pipeline isnt tied to PyPI with a push  (I copy and check the wheels manually) you are free to test as much as you need.

marscher:
I tried pushing this branch to origin/releases/test, but it seems the pipeline didn't run. Unfortunately I cannot login via Github to  Azure, just receiving an error message from their side.

Thrameos:
Darn.  I was worried about that.   I know that I granted credentials to run the pipeline, but there was always a problem with your account.  If you can send me the azure login name I can try to refresh it.

As for now I did a manual trigger.  Should complete in about 10 to 15 minutes.

https://dev.azure.com/jpype-project/jpype/_build/results?buildId=1056&view=results

Thrameos:
Looks like this one is safe to merge.   Our pipelines have improved since the our move to azure and it looks like we don't need the class files to test the packages anymore.   

LGTM!

Thrameos (approved):

PR #1106, opened 2022-11-01
Fixes #1105

### LINKED ISSUE #1105: memoryview(direct ByteBuffer) has incorrect nbytes field
There appears to be a bug in the nbytes on some systems when using direct byte buffer.  This prevents them from being used with io.readinto

Bug replication
```
>>> import jpype
>>> import jpype.imports
>>> jpype.startJVM()
>>> from java.nio import ByteBuffer
>>> memoryview(ByteBuffer.allocateDirect(100)).nbytes
25096256
```

Thrameos:
While attempting to use directbyte buffers with Python io, I found that the nbytes was never set which caused random and unpredictable behaviors.   Fixed and added test.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,9 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 Latest Changes:
 - **1.4.2_dev0 - 2022-10-26**
+
+  - Fixed direct byte buffers not reporting nbytes correctly when cast to memoryview.
+
 - **1.4.1 - 2022-10-26**
   
   - Fixed issue with startJVM changing locale settings.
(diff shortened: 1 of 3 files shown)


marscher (approved):
Good catch!

Thrameos:
It was my bad.   There are multiple shape and offset, itemsize fields in the buffer and some APIs use one field and some use the other.   A lot of them are optional or redundant so it is not clear what is needed.  So when I tested with different things it appeared to work (mostly numpy transactions).  But apparently some other interfaces use the len field which is named differently in the Python view than it is in the C view.   So it wasn't clear that it was broken until I hit this interface. 

I will need to go through all our buffer transactions and make sure none of the rest have a similar bug.  Though now that I know what to look for it should be quick.

PR #1110, opened 2022-11-11
Allow FunctionalInterface conversion when the annotation is not present

### LINKED ISSUE #1107: Python lambda to FunctionalInterface conversion is too strict
Note the following line in the documentation for `FunctionalInterface` seen [here](https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html):

"However, the compiler will treat any interface meeting the definition of a functional interface as a functional interface regardless of whether or not a FunctionalInterface annotation is present on the interface declaration."

Unfortunately, jpype does **not** allow conversion from a lambda to a functional interface if the Java interface does not have the `@FunctionInterface` annotation. However, this should not be the case and any functional interface should be convertable.

The below java and python code snippits provide a simple demonstration of the problem.

```java
package tmp;

public class App {
    
    public static void printNonAnnotated(NonAnnotatedFunctionInterface messageGetter) {
        System.out.println(messageGetter.getMessage());
    }
    
    public static void printAnnotated(AnnotatedFunctionInterface messageGetter) {
        printNonAnnotated(messageGetter);
    }
    
        printNonAnnotated(() -> "Hello from non annotated interface");
        printAnnotated(() -> "Hello from annotated interface");
    }
    
    public static interface NonAnnotatedFunctionInterface {
        
        public String getMessage();
    }
    
    @FunctionalInterface
    public static interface AnnotatedFunctionInterface extends NonAnnotatedFunctionInterface {
    }
}
```

```python
import jpype
from jpype.imports import *

if __name__ == '__main__':
    jpype.startJVM(classpath="tmp.jar")
    from tmp import App
    App.main([])
    App.printAnnotated(lambda : "Hello from Python annotated interface")
    App.printNonAnnotated(lambda : "Hello from Python non annotated interface")
```

Running the above python code results in the following output:

```
Hello from non annotated interface
Hello from annotated interface
Hello from Python annotated interface
  File "C:\Users\astre\Documents\tmp\main.py", line 9, in <module>
    App.printNonAnnotated(lambda : "Hello from Python non annotated interface")
TypeError: No matching overloads found for *static* tmp.App.printNonAnnotated(function), options are:
        public static void tmp.App.printNonAnnotated(tmp.App$NonAnnotatedFunctionInterface)
```

astrelsky:
I will have a look at the current tests tomorrow and see what I can add to throw the kitchen sink at this.

I tried to follow the formatting as close as possible.

Fixes #1107

### DIFF
--- a/native/java/org/jpype/JPypeUtilities.java
+++ b/native/java/org/jpype/JPypeUtilities.java
@@ -55,7 +55,7 @@ public class JPypeUtilities
     }
   }
 
-  static Method getFunctionalInterfaceMethod(Class cls)
+  public static Method getFunctionalInterfaceMethod(Class cls)
   {
     if (!cls.isInterface() || cls.isAnnotation() || isSealed.test(cls))
       return null;
--- a/native/java/org/jpype/JPypeUtilities.java
+++ b/native/java/org/jpype/JPypeUtilities.java
@@ -1,11 +1,47 @@
 package org.jpype;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandleProxies;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.function.Predicate;
 
 public class JPypeUtilities
 {
+  
+  // a functional interface can only re-declare a public non-final method from Object
+  // this should end up being an array of equals, hashCode and toString
+  private static final Method[] OBJECT_METHODS =
+    Arrays.stream(Object.class.getMethods())
+      .filter(m -> !Modifier.isFinal(m.getModifiers()))
+      .toArray(Method[]::new);
+  
+  private static final Predicate<Class> isSealed;
+
+  static
+  {
+    Predicate<Class> result = null;
+    try
+    {
+      Method m = Class.class.getMethod("isSealed");
+      MethodHandle handle = MethodHandles.publicLookup().unreflect(m);
+      result = MethodHandleProxies.asInterfaceInstance(Predicate.class, handle);
+    } catch (IllegalAccessException e)
+    {
+      // it's a public method so this should never occur
+      throw new IllegalAccessError(e.getMessage());
+    } catch (NoSuchMethodException e)
+    {
+      // if isSealed doesn't exist then neither do sealed classes
+      result = c -> false;
+    }
+    isSealed = result;
+  }
 
   public static Path getJarPath(Class c)
   {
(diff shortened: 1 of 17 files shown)


native/java/org/jpype/JPypeUtilities.java
```diff
@@ -1,11 +1,47 @@
...
 import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.function.Predicate;
 
 public class JPypeUtilities
 {
+  
+  // a functional interface can only re-declare a public non-final method from Object
```
astrelsky, line 17: Will fix the indentation here.

Thrameos:
Does this cide all work with jdk 8?

astrelsky:
> Does this cide all work with jdk 8?

It *should*. I don't think I used anything that wasn't available in jdk 8. The reason I got a `MethodHandle` to `Class.isSealed` is to keep jdk 8 compatibility and still keep the check to prevent a potentially confusing exception further down the road.

Thrameos:
Okay.  I will verify it on jdk8 when it is ready for review.  I dont think our cloud currently tests old versions.

astrelsky:
I added some test but unfortunately they are failing at the moment. There is something lingering elsewhere that is causing the cast to fail and I'm not sure where it is.

astrelsky:
@Thrameos any ideas where the cast might be failing?

Thrameos:
So two problems.   

It didn't check that the method exists for setting the FUNCTIONAL around line 460 in ``native/java/org/jpype/manage/TypeManager.java``.  

Unfortunately, doing that also exposes a race condition during initialization.   Normally it won't search for the function name in the ClassDescriptor until after the type system is initialized, but once we start triggering on functions that could be considered functional but don't have the annotation we trigger on things like ``java.lang.invoke.TypeDescriptor`` which is created very early in the process before all the hooks are set up.   We never explicitly load this class but it was fetched by something basic like ``java.lang.Class`` which is why the system is still booting as order is critical until ``Object`` and ``Class`` are completed.   Thus we end up calling ``getFunctional`` before the hook was installed.   I tried some rearrangements of the startup sequence, but no luck yet.   As with all race conditions if takes a while to find the proper ordering which is a bit too challenging for most contributors (especially if I have to rearrange code to make it order free.)  I will keep you posted when I work something out.

Thrameos:
Okay I put in a pull request with the recommended fix.   Hopefully it resolved the issue with the testbench.

astrelsky:
@Thrameos I think it is ready for review. The option to request a review is not available.

Thrameos:
It all looks fine to me.  Though I am somewhat worried that we changed the definition of a public method in ``org.jpype.JPypeContext`` though I consider most of that stuff private so likely not an issue.   (If I ever get helpers for the EPYPJ branch I was going to rearrange that portion anyway.)  So I am going to put it to the master maintainer for review.

astrelsky:
> It all looks fine to me.  Though I am somewhat worried that we changed the definition of a public method in ``org.jpype.JPypeContext`` though I consider most of that stuff private so likely not an issue.   (If I ever get helpers for the EPYPJ branch I was going to rearrange that portion anyway.)  So I am going to put it to the master maintainer for review.

That definition change may have been unintentional and something I did while troubleshooting. I'll look again after work but it should be trivial to avoid that change.

Thrameos:
Not it was required because of the order of operations.    When we make it recognize classes as Lambdas before ``java.lang.Class`` is constructed that means we need to create hooks before the context was finished constructing.   So either I need to make that method static or I need to split the context start up into a before hooks and after hooks, or I could try to get more information passed into the JNI calls so that we don't have to probe back.  

Making this method static was the shortest path to solution.   And given noone is supposed to be accessing those JPypeContext methods outside of diagnostics it should be fine.  The other solutions are not terribly hard to implement, but they are more involved and thus would make for a much larger PR that your original intent.   So lets see if the simplest solution is acceptable first.  But is the maintainer wants a better solution then I will make it so.

Sadly if you remove the static then the system will crash because with a NullPointerException or hook install failure during initialization.

astrelsky:
Oh I didn't realize you made that change (or I did forget about it and it just worked out well)
Either way, if there is someone using it, it would still work and would just generate a warning about how a static method should be accessed in a static way blah blah.

marscher (approved):

PR #1113, opened 2022-11-30
Added additional level to break ambiguity when derived classes are used

Thrameos:
This is a proposed change to the method resolution such that using a derived type when matching a method rather than the base class will not alter the method resolution.

Example:

Methods
```
   public void method(boolean b, Base v);
   public void method(Object o, Base v);
```

Will successfully resolve ``method(True, Base())`` to ``method(boolean, Base)`` and ``method(JObject(True), Base())`` to ``method(Object, Base)``, but ``method(True, Derived())`` will currently produce an ambiguity.   This PR fixes the issue by making a new level between exact and implicit.


Thrameos:
It is somewhat unclear if user defined conversions should be consider to be derived or implicit.   They are automatic but create a new instance so I am going to leave them as implicit for now.  But another way to view them is that they substitute for some Java type so they should be considered to be as good as the resulting type which would make them either exact or derived depending on the situation.

Thrameos:
@marscher It appears that the solution was only slightly more complex than proposed.  I forgot that ``java.lang.Object`` has a special conversion path.  But otherwise, it was almost exactly on target.   It is still a bit unclear how Python object should play into this.  Currently they are at the same level as Java implicit conversions (such as automatic boxing and primitive promotions), with the sole exception of Python strings which currently are given the level of "exact".    I could promote them to a higher level than automatic conversions (add a user level between derived and implicit, or I could bump them to derived.   If they are at the exact level then if you have two possible Python conversions it would chose the first unconditionally which likely isn't correct.   

I don't think this will bust the cache as the Python hash codes when you change types should still hold.  But it took some examination to be sure.

test/jpypetest/test_conversion.py
```diff
@@ -144,3 +144,9 @@ def testCharBytes(self):
         with self.assertRaises(TypeError):
             ja[0:1] = [object()]
         ja[0:1] = ["\u265E"]
+
+    def testDerived(self):
+        # Test Java Object any conversion
+        self.assertEqual(jpype.JClass("java.lang.Object")._canConvertToJava(jpype.JClass("java.lang.String")("a")), "derived")
```
marscher, line 150: Is there actually another way to test this. I would like to avoid testing private implementation (even though this interface hasn't changed in a decade).

marscher (approved):

Thrameos:
Will merge this once I up the minor number.

Issue #1147, opened 2023-09-26
Support Python 3.12


EwoutH:
[Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html) is releasing in [exactly one week](https://peps.python.org/pep-0693/), Monday 2023-10-02. It would be great if JPype could support Python 3.12 fully at the release, including testing in CI and wheels uploaded to PyPI.

<details><summary>Summary – Python 3.12 Release highlights</summary>

> New grammar features:
> - [PEP 701: Syntactic formalization of f-strings](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep701)
> 
> Interpreter improvements:
> - [PEP 684: A Per-Interpreter GIL](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep684)
> - [PEP 669: Low impact monitoring for CPython](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep669)
> 
> New typing features:
> - [PEP 688: Making the buffer protocol accessible in Python](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep688)
> -  [PEP 692: Using TypedDict for more precise **kwargs typing](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep692)
> -  [PEP 695: Type Parameter Syntax](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep695)
> - [PEP 698: Override Decorator for Static Typing](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep698)
> 
> Important deprecations, removals or restrictions:
> -  [PEP 623](https://peps.python.org/pep-0623/): Remove wstr from Unicode
> -  [PEP 632](https://peps.python.org/pep-0632/): Remove the distutils package. See [the migration guide](https://peps.python.org/pep-0632/#migration-advice) for advice on its replacement.
> -  [gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3.12/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3.12/library/venv.html#venv-explanation) virtual environment.</details>

marscher:
[...]

PR #1148, opened 2023-10-02
fix: Python 3.12 support; remove usage of wstr and use updated PyLong ob_digit location

Christopher-Chianelli:
- wstr was removed in CPython 3.12: https://www.python.org/dev/peps/pep-0623
- ob_digit was moved from PyObject to PyLongObject, which is accessible from long_value on PyObject

Investigating a segmentation fault with PyJP_GetAttrDescriptor

```
C  [libpython3.12.so.1.0+0x1f2882]  PyDict_GetItem+0x22
C  [_jpype.cpython-312-x86_64-linux-gnu.so+0x8df43]  PyJP_GetAttrDescriptor+0x73
C  [_jpype.cpython-312-x86_64-linux-gnu.so+0x87758]  PyJPClass_setattro+0x98
```

### DIFF
--- a/.azure/build.yml
+++ b/.azure/build.yml
@@ -65,6 +65,9 @@ jobs:
       linux-3.11:
         imageName: "ubuntu-latest"
         python.version: '3.11'
+      linux-3.12:
+        imageName: "ubuntu-latest"
+        python.version: '3.12'
       windows-3.7:
         imageName: "windows-2019"
         python.version: '3.7'
@@ -81,6 +84,9 @@ jobs:
       windows-3.11:
         imageName: "windows-2019"
         python.version: '3.11'
+      windows-3.12:
+        imageName: "windows-2019"
+        python.version: '3.12'
       mac-3.9:
         imageName: "macos-11"
         python.version: '3.9'
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -9,6 +9,7 @@ steps:
     version: 11
 
 - script: |
+    python -m pip install --upgrade pytest setuptools
     python setup.py build_ext --inplace
   displayName: 'Build module'
 
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -9,7 +9,7 @@ steps:
     version: 11
 
 - script: |
-    python -m pip install --upgrade pytest setuptools
+    python -m pip install --upgrade pytest-xdist setuptools
     python setup.py build_ext --inplace
   displayName: 'Build module'
 
@@ -24,7 +24,8 @@ steps:
   displayName: 'Install test'
 
 - script: |
-    python -m pytest -v --junit-xml=build/test/test.xml test/jpypetest --checkjni
+    # run tests in two workers.
+    python -m pytest -v -n 2 --junit-xml=build/test/test.xml test/jpypetest --checkjni
   displayName: 'Test JDK 11'
   condition: eq(variables['jpypetest.fast'], 'false')
 
--- a/native/common/jp_primitivetype.cpp
+++ b/native/common/jp_primitivetype.cpp
@@ -47,7 +47,11 @@ PyObject *JPPrimitiveType::convertLong(PyTypeObject* wrapper, PyLongObject* tmp)
 	((PyVarObject*) newobj)->ob_size = Py_SIZE(tmp);
 	for (Py_ssize_t i = 0; i < n; i++)
 	{
+        #if PY_VERSION_HEX<0x030C0000
+        newobj->ob_digit[i] = tmp->ob_digit[i];
+        #else
         ((PyLongObject*)newobj)->long_value.ob_digit[i] = ((PyLongObject*)tmp)->long_value.ob_digit[i];
+        #endif
 	}
 	return (PyObject*) newobj;
 }
--- a/native/common/jp_primitivetype.cpp
+++ b/native/common/jp_primitivetype.cpp
@@ -47,7 +47,7 @@ PyObject *JPPrimitiveType::convertLong(PyTypeObject* wrapper, PyLongObject* tmp)
 	((PyVarObject*) newobj)->ob_size = Py_SIZE(tmp);
 	for (Py_ssize_t i = 0; i < n; i++)
 	{
-		newobj->ob_digit[i] = tmp->ob_digit[i];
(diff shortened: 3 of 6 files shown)


Christopher-Chianelli:
Fixes #1147 when it actually works, but this make the code compilable at least.

cclauss:
Please add the continuous integration tests from #1153 so we can monitor progress.

Christopher-Chianelli:
There are quite a few non-trivial changes to the internals of CPython 3.12 that JPype uses, which unfortunately makes supporting CPython 3.12 far from trivial. Currently trying to fix the segmentation fault on `testGetStaticByInstance`. I am currently working on this issue in my free time, so progress will be slow; don't expect it anytime soon unless someone else does it.

Thrameos:
I am trying to cycle back to it after not having a computer for last 3 months.

marscher:
> I am trying to cycle back to it after not having a computer for last 3 months.

What the heck. That's a long time I have to say. Did you take some sabbath? :+1:

marscher:
@Christopher-Chianelli I advice you to use pytest-xdist and invoke pytest -n2 to enable all tests to run, even if there are crashes. I already pushed this to CI.

Here is an overview of all failed tests: https://dev.azure.com/jpype-project/jpype/_build/results?buildId=1129&view=ms.vss-test-web.build-test-results-tab

Thrameos:
@encukou  Bad news.... the API for extension of objects added in 3.12 appears to be a significant miss in terms of what is required for JPype.   

There are two reasons for this.

1) The JPClassType is used as a Meta class which derives from PyType and contains a significant amount of non-trivial code.   The new method gets called from within the module which passes a special flag which indicates that it is a safe operation.  If the user attempts to extend a Java class from within Python it fails with an error (which is for safety because you can't create a new Java class in Python).
2) Second it adds to the basicsize of the object which defeats the point of the hidden reserved space.   If you add to the basic size and then derive a new type say PyException then you will get the classic problem that Python objects can only be single inherited.   JPype was using true multiple inheritance.   That is it derives from multiple Python types that have layouts which would otherwise be in complete conflict because the reserve space is uses was a floating offset modeled after dictoffset.   If we add to the basicsize of the object rather than having the concept of a user reserve that floats just like dict and weak references, then you can't use this for multiple inhertance.

Unfortunately this places JPype in a seriously bad spot for 3.12.    The existing "tp_alloc" hack is currently breaking because something wrote to the space after the allocation.  And the API intended to be the replacement is short of what is required for universal multiple inheritance.  

Is the GC now using the memory after the object or is there a bug someplace that could write into that location?

Thrameos:
@encukou Eeeek!   I just found the same bug that was haunting me in the Python 3.12 code.   Assuming I am reading this right if you create a class  derived from int  (ie  `` class Foo(int): pass``) and then try to use the dictionary it will call this code....

```
PyObject **
_PyObject_ComputedDictPointer(PyObject *obj)
{
    PyTypeObject *tp = Py_TYPE(obj);
    assert((tp->tp_flags & Py_TPFLAGS_MANAGED_DICT) == 0);

    Py_ssize_t dictoffset = tp->tp_dictoffset;
    if (dictoffset == 0) {
        return NULL;
    }

    if (dictoffset < 0) {
        assert(dictoffset != -1);

        Py_ssize_t tsize = Py_SIZE(obj);  <====  This wont work because PyLong has blown away ob_size
        if (tsize < 0) {    <== These lines were for the old version of PyLong in which -size was used to mean a negative number 
            tsize = -tsize;
        }
        size_t size = _PyObject_VAR_SIZE(tp, tsize);
        assert(size <= (size_t)PY_SSIZE_T_MAX);
        dictoffset += (Py_ssize_t)size;

        _PyObject_ASSERT(obj, dictoffset > 0);
        _PyObject_ASSERT(obj, dictoffset % SIZEOF_VOID_P == 0);
    }
    return (PyObject **) ((char *)obj + dictoffset);
}
```

Thus the dictionary position will end up way beyond the end of the object into someone else's memory.

Thrameos:
I have a working copy of JPype for Python 3.12.   Next up it to work again with Python 3.10.

encukou:
> I think there is a very simple implementation for multiple inheritance base on what you included.
> The only difficulty with this implementation is that [... context omitted ...] you have to decide if mixings are virtual or not

Well, PEPs welcome... But it does seem like a real implementation would be as complex as the grand layout plan (and possibly quite similar to it, in the end).

---

> if you create a class derived from int (ie  `class Foo(int): pass`) and then try to use the dictionary it will call this code

Thanks for digging into that one! I guess it's another case of `int` optimizations invalidating assumptions :(

Doing this with ``class Foo(int): pass`` won't crash, because the Python statement will now use `Py_TPFLAGS_MANAGED_DICT` rather than `tp_dictoffset`. But it does crash when subclassing in C with negative `tp_dictoffset`, and I don't see why that should not be valid.

I filed https://github.com/python/cpython/issues/112743

Thrameos:
Regarding the PR how limited in scope do you want?

1) Moving data requested by negative basesize in front of the object like MANAGED_DICT.   In which case there can only be one, but otherwise no API breakage.

2) Moving var objects to start with offset which allows non-trivial basesize to be mixed with var objects like long and str.   Minor breakage in Python, but known amounts in other packages.  (May not solve my issue because conflict with exception, but solvable another way.)

3) Adding look up for general mixing of opaque types.   Requires additional tables to support lookup in the class structure if a class has more than one opaque types.  Requires API that can request which opaque type is for the offset.  API change now and potentially allows for fancy things at the cost of complexity.

Which is your preference?  I have finally gotten permission to sign the form.

Thrameos:
@encukou Can you give me some insight as to which PR would be most acceptable in terms of API breakage and rearrangement?   To fact that JPype currently requires access to Python private methods as opposed to using public API is currently the largest maintenance issue here, especially if my job decides that we don't have enough funding to support JPype (which it often does).   I would like to get a few PR in that improves this situation for the long term.

encukou:
Sorry, I now realize “PEPs welcome” looks too much like “PRs welcome” :(

I think this'll need a PEP -- a change proposal listing the pros and cons of possible solutions. There are too many details and people involved. And while your proposals look simple, but I don't think they'll be that simple in practice. 

1. What happens if there *is* more than one? That can't break (otherwise e.g. changing a base class from absolute `tp_basicsize` to negative would be a breaking change). Would there be a flag to request moving data to the front? Or would this happen automatically?
2. I'd really like to avoid breakage, even “known amounts”. (I don't speak for CPython as a whole, of course...)
3. This sounds like the grand unified layout, which still looks like the best solution.

PR #1152, opened 2023-10-31
findjvm.py: print() is a function in Python 3

cclauss:


marscher:
Thanks for fixing that. Did you actually try to run these examples? Do they provide any meaningful output?
I think they are not used/updated for many years.

PR #1153, opened 2023-10-31
.azure/build.yml: Add Python 3.12 to the build

cclauss:
Testing for
* #1147

This will require the changes proposed in
* #1148


marscher:
Thanks, but #1148 already adds CI support for Python 3.12

Issue #1111, opened 2022-11-14
Issue with non-ascii characters in the JVM path on Windows


villares:
Context: in Brazil, many users have names, usernames and various directories containing non-ascii characters. I found this issue because some students couldn't run the same project/tools that other students (and myself) could run.

A mininmal way to reproduce is to put the JDK in a non-ascii path location. And then try the REPL steps bellow:

```python
Python 3.10.6 (C:\Users\abav\OneDrive\Desktop\é-non-ascii-path\python.exe)
>>> import jpype
>>> jpype.getDefaultJVMPath()
'C:\\Users\\abav\\OneDrive\\Desktop\\é-non-ascii-path\\user_data\\jdk-17\\bin\\server\\jvm.dll'
>>> jpype.startJVM(jvmpath='C:\Users\abav\OneDrive\Desktop\é-non-ascii-path\user_data\jdk-17\bin\server\jvm.dll')
  File "<stdin>", line 1
    jpype.startJVM(jvmpath='C:\Users\abav\OneDrive\Desktop\é-non-ascii-path\user_data\jdk-17\bin\server\jvm.dll')
                                                                                                                ^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
>>> jpype.startJVM()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\abav\OneDrive\Desktop\é-non-ascii-path\lib\site-packages\jpype\_core.py", line 218, in startJVM
    _jpype.startup(jvmpath, tuple(args),
OSError: [WinError 126] JVM DLL not found: C:\Users\abav\OneDrive\Desktop\é-non-ascii-path\user_data\jdk-17\bin\server\jvm.dll
```

Removing the `é` character from the folder name allows the JVM to start.

This seems to be a Windows issue, tested on Windows 10 21H1, and It seems not to be present on Linux.

Thrameos:
I am going to have to depend on you to find a way to address the issue, as my local system likely won't produce the correct result.   Given that it is a SystemError rather than a Java error of some kind I am not sure if it is actually getting to the Java code.     Most likely some encoding converter when passing the string from Python to Java or in this case the C library forget to specify the encoding.

The specific code path is 

Raw char* gets passed to Windows  ``LoadLibrary()`` call
```
        virtual void loadLibrary(const char* path) override
        {
[...]

PR #1174, opened 2024-02-24
Windows locale

Thrameos:
This is an attempt to make JPype work when there are international characters in the path name.   Currently does not appear to make a significant difference because the JVM itself does not support it on the tested system.


Thrameos:
@marscher This is meant to address #1111 but as I don't have a system that supports international paths for the JVM I can't really tell if it fixes anything more than moving the failure point.  If the international testers do find it to fix the problem, I will complete this PR with the usual changelog entry, ask for a review, etc.

marscher:
Thanks at lot and sorry for the delay!

PR #1195, opened 2024-06-16
Fix `IllegalArgumentException` with non-ascii paths

### LINKED ISSUE #1194: IllegalArgumentException after using `dir(jpype.JPackage("mypackage"))` and non-ascii paths.
It appears that `dir(jpype.JPackage("mypackage"))` does not handle non-ASCII characters anywhere in its package path. This method is critical for tools like [stubgenj](https://gitlab.cern.ch/scripting-tools/stubgenj) that loop over packages using `dir`.

Consider the following working and broken cases: 
```
├── bad_workdir_à
│   ├── bad_example.py
│   ├── MyClass.java
│   ├── mypackage
│   │   └── MyClass.class
│   └── mypackage.jar
└── good_workdir
    ├── good_example.py
    ├── MyClass.java
    ├── mypackage
    │   └── MyClass.class
    └── mypackage.jar
```
Both are identical, except for the name of the working directory. The `good_example.py` and `bad_example.py` scripts are:
```python
import jpype
jpype.addClassPath("mypackage.jar")
print(dir(jpype.JPackage("mypackage")))
```

The output of `bad_workdir_à/bad_example.py` is:
```
  File "org.jpype.pkg.JPypePackage.java", line -1, in org.jpype.pkg.JPypePackage.getContents
Exception: Java Exception

  File "/home/antonlee/github.com/tachyonicClock/A/CapyMOA/jpype_bug/bad_workdir_à/bad_example.py", line 4, in <module>
    print(dir(jpype.JPackage("mypackage")))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
java.lang.java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: Bad escape
```

The output of `good_workdir/good_example.py` is:
```
['MyClass']
```

---

I would expect the output of `bad_workdir_à/bad_example.py` to be the same as `good_workdir/good_example.py`. I've attached the files I used to reproduce this bug.

[jpype_bug.zip](https://github.com/user-attachments/files/15855799/jpype_bug.zip)

tachyonicClock:
Fix #1194


Thrameos (approved):
LTGM

Thrameos:
Looks like I have some maintenance tasks before the CI will run.

tachyonicClock:
@Thrameos  I'm working on a test and it seems emoji still don't work although that appears to be a different reason.

tachyonicClock:
The emoji thing seems to be a problem with java itself
```
 java -classpath "😊/mypackage.jar" mypackage.MyClass
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.IllegalArgumentException: Error decoding percent encoded characters
        at java.base/sun.net.www.ParseUtil.decode(ParseUtil.java:218)
        at java.base/jdk.internal.loader.FileURLMapper.getPath(FileURLMapper.java:64)
        ...
```
https://youtrack.jetbrains.com/issue/IJPL-19330/Scratch-Java-file-wont-run-if-theres-an-emoji-in-the-project-name

Thrameos:
I am hoping it is safe to ignore emoji.   Although a coworker of mine had great fun using the poop emoji when we added internationalization to string conversions, I think that it is an edge case.

It is also a good idea to add a line to "doc/CHANGELOG.rst"  describing the change so that others will know it is fixes.  Thanks for taking the time to add tests.

tachyonicClock:
Since Java appears to have difficulties with emojis in classpaths, it's probably impossible to fix this issue. The important thing is to support non-English character sets that are likely in user paths. That is the problem that prompted me to open this issue in the first place. Although a poop emoji directory does sound like a good place to store certain parts of java :D.

I've added my changes to the changelog. Let me know if its not in the right place or the wording is bad.

test/jpypetest/test_non_ascii_paths.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+import unittest
+import jpype
+from pathlib import Path
+import shutil
+from multiprocessing import get_context, Queue, Process
+
+
+class TestNonAsciiPaths(unittest.TestCase):
```
marscher, line 12: please derive from common.JPypeTestCase, then you would not need to take about a running JVM.

marscher: Thanks, I should have noticed that. Would you be so kind to add this as a comment at the class declaration, please? Then nobody would later on think, it'd be a good idea to change that again.

tachyonicClock: Yea no worries :D I've just added that comment.

marscher (approved):

marscher:
Honestly I do not understand right now, why the pipeline isnt converging. On the main branch everything was fine.

tachyonicClock:
I fixed the hanging issue and packaged it with the related tests in `test_startup.py`. These are run as sub-processes using `subrun'. I'm running into an issue with them failing on Windows. I'll try to fix this later this week.

tachyonicClock:
Last week I tried fixing this on windows but ran into some other issues related to UTF-8 vs UTF-16 character encoding on Windows in the JNI.

marscher:
@tachyonicClock Thanks for your work on this so far! Any updates on this one? We would like to conclude a set of changes for the next release and I would like to include this fix if it also works on Windows.

tachyonicClock:
- Launching Java with non-ascii class paths appears impossible on Windows (https://stackoverflow.com/questions/20052455/jni-start-jvm-with-unicode-support). To resolve this issue, I tried adding the classpath after initialization since this supports Unicode. But the change to how jpype starts up is significant.
-  I fixed a problem in Python 3.8 with JDK8 todo with "zip file system provider returns doubly % encoded URIs" (`native/java/org/jpype/pkg/JPypePackageManager.java`). A pre-existing fix removed % signs if `'%2520'` was present. This did not fix doubly encoded Unicode characters and risked removing % signs that were intentional. The new version will work in these.
- Emoji now work.
- I added `project/jars/unicode_à😎`  to describe how to build the Unicode jar test case. I tried to follow the pattern of other examples, but I'm not confident I did this correctly.

jpype/_core.py
```diff
@@ -264,6 +264,21 @@ def startJVM(
...
+    Unfortunately, unicode is unsupported by this interface on windows, since
+    windows uses wide-byte (16bit) character encoding.
+    See: https://stackoverflow.com/questions/20052455/jni-start-jvm-with-unicode-support
+
+    To resolve this issue we add the classpath after initialization since jpype
+    itself supports unicode class paths.
+    """
+    for cp in _handleClassPath(classpath):
+        addClassPath(Path.cwd() / Path(cp).resolve())
+
```
Thrameos, line 281: It appears this code is going to cause some serious issues because certain types of classpath absolutely must be loaded prior to starting the JVM.   The two that come to mind are database drivers and anything containing a classloader.   

I believe there is a solution in that we just need to send a bytes with the correct encoding, but I need to test it.

tachyonicClock: I'm sorry this ended up breaking things.

> I believe there is a solution in that we just need to send a bytes with the correct encoding, but I need to test it.

I wish you luck. I spent some time struggling to get this to work, but did not try exhaustively since I found the stack-overflow response suggesting I set the properties at runtime. I found some discussion of this problem from 2017 but it does not look like it went anywhere: https://mail.openjdk.org/pipermail/jdk10-dev/2017-May/000191.html.

PR #1196, opened 2024-06-17
Changes for numpy 2.0 to fix test matrix.

Thrameos:
This PR deals with the broken test matrix following the removal for ``float_``.   I replaced it with ``float16`` though that is actually a different test.   I dealt with the required alterations to support ``half`` type for array conversion, though I can't find a good pattern to deal with byte reversed arrays so that may be untested.

### DIFF
--- a/native/common/jp_convert.cpp
+++ b/native/common/jp_convert.cpp
@@ -14,9 +14,60 @@
    See NOTICE file for details.
  *****************************************************************************/
 #include "jpype.h"
+#include <math.h>
+#include <bitset>
 
 namespace
 {
+
+template <jvalue func(void *c) >
+class Half
+{
+public:
+	static jvalue convert(void* c)
+    {
+        uint16_t i = *(uint16_t*) c;
+		uint32_t sign = (i&0x8000)>>15;
+		uint32_t man  = (i&0x7C00)>>10;
+		uint32_t frac = (i&0x03ff);
+		uint32_t k = sign<<31;
+
+		if (man == 0)
+		{
+			// subnormal numbers
+			if (frac != 0)
+			{
+				frac = frac | (frac >> 1);
+				frac = frac | (frac >> 2);
+				frac = frac | (frac >> 4);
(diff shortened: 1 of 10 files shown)


Thrameos:
Seem like I gave one more set of changes before this is complete.  Not sure why onle one version triggered it.

Thrameos:
I am unable to replace the error seen in the test suite but it most likely is a Python behavior not a JPype one.   The test was making sure that all cast types go through required paths.   Python itself through the exception that "e" type is not a simple data type.  If they choose to change that behavior then we shouldn't test against it.  

@marscher I believe this is ready to go.

native/common/jp_convert.cpp
```diff
@@ -14,10 +14,61 @@
    See NOTICE file for details.
  *****************************************************************************/
 #include "jpype.h"
+#include <math.h>
+#include <bitset>
 
 namespace
 {
 
+template <jvalue func(void *c) >
```
Thrameos, line 23: This is a crude implementation of the "half" floating type.    There is no native type in Java to support this but it is possible that someone could pass a half floating in for promotion to either float or double.   This code is exercised in the test suite.

native/common/jp_convert.cpp
```diff
@@ -385,6 +436,31 @@ jconverter getConverter(const char* from, int itemsize, const char* to)
 				case 'd': return &Convert<double>::toD;
 			}
 			break;
+		case 'e':
+			if (reverse) switch (to[0])
```
Thrameos, line 440: I am not sure if there is every a case in which a reversed byte order half floating value would be converted.   Through the order of this operation is reasonable to define.   Reverse the bytes, promote to float then convert to Java type.

Thrameos (commented):
Not much to this other than changes to the test bench to get it back on line.   

The only odd code was to better support the half float which is valid buffer type in Python.   There is no native Java type for it.

marscher:
Thanks. LGTM

PR #1214, opened 2024-09-01
"Generic" JArray support

astrelsky:
This allows the use of a parameterized `JArray` in type annotations. There was no other way to represent a java array type.

This also includes two small fixes where you could not do `JArray(JClass)` or `JClass[0]`.

### DIFF
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -339,6 +339,7 @@ def initializeResources():
     _jpype._type_classes[object] = _jpype._java_lang_Object
     _jpype._type_classes[_jpype.JString] = _jpype._java_lang_String
     _jpype._type_classes[_jpype.JObject] = _jpype._java_lang_Object
+    _jpype._type_classes[_jpype.JClass] = _jpype._java_lang_Class
     _jinit.runJVMInitializers()
 
     _jpype.JClass('org.jpype.JPypeKeywords').setKeywords(
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -92,6 +92,18 @@ class JArray(_jpype._JObject, internal=True):  # type: ignore[call-arg]
     def of(cls, array, dtype=None):
         return _jpype.arrayFromBuffer(array, dtype)
 
+    def __class_getitem__(cls, key):
+        if key is _jpype.JClass:
+            # explicit check for JClass
+            # _toJavaClass cannot be used
+            # passing int, float, etc is not allowed
+            key = _jpype._java_lang_Class
+        if isinstance(key, (str, _jpype._java_lang_Class)):
+            key = _jpype.JClass(key)
+        if isinstance(key, _jpype.JClass):
+            return type(key[0])
+        raise TypeError("Cannot instantiate unspecified array type")
+
 
 class _JArrayProto(object):
 
@@ -104,6 +116,10 @@ class _JArrayProto(object):
     def __reversed__(self):
         for elem in self[::-1]:
             yield elem
+    
+    def __contains__(self, item):
+        # "in" works without this but this should be more efficient
+        return _jpype.JClass("java.util.Arrays").asList(self).contains(item)
 
     def clone(self):
         """ Clone the Java array.
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -102,7 +102,7 @@ class JArray(_jpype._JObject, internal=True):  # type: ignore[call-arg]
             key = _jpype.JClass(key)
         if isinstance(key, _jpype.JClass):
             return type(key[0])
-        raise TypeError("Cannot instantiate unspecified array type")
+        raise TypeError("%s is not a Java class" % key)
 
 
 class _JArrayProto(object):
@@ -116,10 +116,6 @@ class _JArrayProto(object):
     def __reversed__(self):
         for elem in self[::-1]:
             yield elem
-    
-    def __contains__(self, item):
-        # "in" works without this but this should be more efficient
-        return _jpype.JClass("java.util.Arrays").asList(self).contains(item)
 
     def clone(self):
         """ Clone the Java array.
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -96,9 +96,9 @@ class JArray(_jpype._JObject, internal=True):  # type: ignore[call-arg]
         if key is _jpype.JClass:
             # explicit check for JClass
             # _toJavaClass cannot be used
-            # passing int, float, etc is not allowed
+            # passing int, float, str, etc is not allowed
             key = _jpype._java_lang_Class
-        if isinstance(key, (str, _jpype._java_lang_Class)):
+        if isinstance(key, _jpype._java_lang_Class):
             key = _jpype.JClass(key)
         if isinstance(key, _jpype.JClass):
             return type(key[0])
--- a/jpype/_jarray.pyi
+++ b/jpype/_jarray.pyi
@@ -24,7 +24,7 @@ __all__ = ['JArray']
 T = typing.TypeVar('T')
 
 
-class JArray(typing.Collection[T], typing.Reversible[T]):
+class JArray(typing.Generic[T]):
 
     def __new__(cls, tp: typing.Type[T], dims=1) -> JArray[T]:
         ...
@@ -37,10 +37,6 @@ class JArray(typing.Collection[T], typing.Reversible[T]):
         ...
 
     def __iter__(self) -> typing.Iterator[T]:
-        ...
-
-    def __contains__(self, item: T) -> bool:
-        # this is implicitly implemented
         ...
 
     def __reversed__(self) -> typing.Iterator[T]:
@@ -60,7 +56,4 @@ class JArray(typing.Collection[T], typing.Reversible[T]):
 
     @typing.overload
     def __setitem__(self, index: slice, value: typing.Sequence[T]):
-        ...
-
-    def __class_getitem__(cls, key: T) -> typing.Type[JArray[T]]:
         ...
--- a/jpype/_jarray.pyi
+++ b/jpype/_jarray.pyi
@@ -15,7 +15,6 @@
 #   See NOTICE file for details.
 #
 # *****************************************************************************
-import abc
 import collections.abc
 import typing
 
@@ -26,18 +25,30 @@ __all__ = ['JArray']
 T = typing.TypeVar('T')
 
 
-class _JArrayGeneric(collections.abc.Sequence[T]):
-
-    @abc.abstractmethod
-    def __setitem__(self, index, value):
-        ...
-
-
-class JArray(_JArrayGeneric[T], metaclass=abc.ABCMeta):
+class JArray(collections.abc.Collection[T], collections.abc.Reversible[T]):
 
     def __new__(cls, tp, dims=1):
         ...
 
     @classmethod
     def of(cls, array, dtype=None):
+        ...
+
+    def __len__(self):
+        ...
+
+    def __iter__(self):
+        ...
+
+    def __contains__(self, item):
+        # this is implicitly implemented
+        ...
+
+    def __reversed__(self):
+        ...
+
+    def __getitem__(self, key):
+        ...
+
+    def __setitem__(self, index, value):
         ...
--- a/jpype/_jarray.pyi
+++ b/jpype/_jarray.pyi
@@ -15,7 +15,6 @@
 #   See NOTICE file for details.
 #
 # *****************************************************************************
-import collections.abc
 import typing
 
 
@@ -25,30 +24,43 @@ __all__ = ['JArray']
 T = typing.TypeVar('T')
 
 
-class JArray(collections.abc.Collection[T], collections.abc.Reversible[T]):
+class JArray(typing.Collection[T], typing.Reversible[T]):
 
-    def __new__(cls, tp, dims=1):
+    def __new__(cls, tp: typing.Type[T], dims=1) -> JArray[T]:
         ...
 
     @classmethod
     def of(cls, array, dtype=None):
         ...
 
-    def __len__(self):
+    def __len__(self) -> int:
         ...
 
-    def __iter__(self):
+    def __iter__(self) -> typing.Iterator[T]:
         ...
 
-    def __contains__(self, item):
+    def __contains__(self, item: T) -> bool:
         # this is implicitly implemented
         ...
 
-    def __reversed__(self):
+    def __reversed__(self) -> typing.Iterator[T]:
         ...
 
-    def __getitem__(self, key):
+    @typing.overload
+    def __getitem__(self, key: int) -> T:
         ...
 
-    def __setitem__(self, index, value):
+    @typing.overload
+    def __getitem__(self, key: slice) -> JArray[T]:
+        ...
+
+    @typing.overload
+    def __setitem__(self, index: int, value: T):
+        ...
+
+    @typing.overload
+    def __setitem__(self, index: slice, value: typing.Sequence[T]):
+        ...
+
+    def __class_getitem__(cls, key: T) -> typing.Type[JArray[T]]:
         ...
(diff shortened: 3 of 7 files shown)


jpype/_jarray.py
```diff
@@ -15,14 +15,34 @@
...
 
-class JArray(_jpype._JObject, internal=True):  # type: ignore[call-arg]
+T = typing.TypeVar('T')
+
+
+class _JArrayGeneric(collections.abc.Sequence[T]):
+    
+    @abc.abstractmethod
+    def __setitem__(self, index, value):
+        ...
```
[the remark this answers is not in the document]
astrelsky: Yes, that is the point stupid bot

jpype/_jarray.py
```diff
@@ -15,14 +15,34 @@
...
 
 
 __all__ = ['JArray']
 
 
-class JArray(_jpype._JObject, internal=True):  # type: ignore[call-arg]
+T = typing.TypeVar('T')
+
+
+class _JArrayGeneric(collections.abc.Sequence[T]):
```
pelson, line 31: Note that this could live in a `pyi` file (like https://github.com/jpype-project/jpype/blob/master/jpype/_jcollection.pyi).

I can definitely see the advantage of it being in the same file though - whatever we choose, we should be consistent.

jpype/_jarray.pyi
```diff
@@ -0,0 +1,43 @@
...
+import typing
+
+
+__all__ = ['JArray']
+
+
+T = typing.TypeVar('T')
+
+
+class _JArrayGeneric(collections.abc.Sequence[T]):
```
astrelsky, line 29: It just occurred to me that this can be removed entirely. I will take care of that tonight

test/jpypetest/test_array.py
```diff
@@ -593,6 +593,7 @@ def testShortcut(self):
         # Check Objects
         self.assertEqual(JString[5].getClass(), JArray(JString)(5).getClass())
         self.assertEqual(JObject[5].getClass(), JArray(JObject)(5).getClass())
+        self.assertEqual(JClass[5].getClass(), JArray(JClass)(5).getClass())
```
Thrameos, line 596: This seems reasonable.  Though it is a rather limited use case.  So long as it is consistent with the rest of the library I have no objections.

Thrameos (changes requested):
Everything looks good, but there needs to be a message in ``doc/CHANGELOG.rst`` as to changes to the public API.  Just one sentence should be sufficient.   Thanks for the contribution.

jpype/_core.py
```diff
@@ -339,6 +339,7 @@ def initializeResources():
     _jpype._type_classes[object] = _jpype._java_lang_Object
     _jpype._type_classes[_jpype.JString] = _jpype._java_lang_String
     _jpype._type_classes[_jpype.JObject] = _jpype._java_lang_Object
+    _jpype._type_classes[_jpype.JClass] = _jpype._java_lang_Class
```
Thrameos, line 342: This seems reasonable.  Though technically all uses of `_type_classes` needs checking.  There used to be an edge case though I don't see it any more.   The only user is ``_jarray`` which i checked.

native/python/pyjp_class.cpp
```diff
@@ -728,8 +728,9 @@ static PyObject *PyJPClass_array(PyJPClass *self, PyObject *item)
 
 	if (self->m_Class == NULL)
 	{
-		PyErr_Format(PyExc_TypeError, "Cannot instantiate unspecified array type");
-		return NULL;
+		PyObject *res = PyObject_CallMethod((PyObject *)self, "__class_getitem__", "O", item);
+		Py_DECREF(item);
+		return res;
```
Thrameos, line 733: This is the most suspect.  It appears to be using the side effect that JClass never got its javaslot filled in.   I would wonder if one could just set the slot properly and use the regular path.   May not be worth the effort, though it likely deserves a comment on the line to to state what is happening here as I am likely to get confused if I saw this without context.

jpype/_jarray.py
```diff
@@ -104,6 +116,10 @@ def __iter__(self):
     def __reversed__(self):
         for elem in self[::-1]:
             yield elem
+    
+    def __contains__(self, item):
+        # "in" works without this but this should be more efficient
+        return _jpype.JClass("java.util.Arrays").asList(self).contains(item)
```
Thrameos, line 122: Was this change documented.   It seems like a separate enhancement.  I would recommend that to avoid multiple lookups it may be best to add it to the ``_core.py`` typecache.

```
   _jpype._java__util_Array = _jpype.JClass("java.util.Array")
```

Though the type look up system is fairly fast and using "in" as part of loop is likely going to be slow anyway.

jpype/_jarray.py
```diff
@@ -92,6 +92,18 @@ def __new__(cls, tp, dims=1):
...
+            # explicit check for JClass
+            # _toJavaClass cannot be used
+            # passing int, float, etc is not allowed
+            key = _jpype._java_lang_Class
+        if isinstance(key, (str, _jpype._java_lang_Class)):
+            key = _jpype.JClass(key)
+        if isinstance(key, _jpype.JClass):
+            return type(key[0])
+        raise TypeError("Cannot instantiate unspecified array type")
+
```
Thrameos, line 106: Was this path tested?   I would worry that the code above didn't check for the case and we may segfault.

jpype/_jarray.py
astrelsky: May as well, there are several other places in jpype where Arrays is used where it could benefit such as the auto conversions.

I didn't document the change, there is no behavior change that this introduces. I think Python just automatically iterates over it and checks if there is no `__contains__` 

The reason I added it was so that it completes the Sequence interface.

native/python/pyjp_class.cpp
astrelsky: I found this part of the code to be a bit strange. The reason the java slot is not filled in is because JArray is a factory with no underlying java class. I was actually surprised that Python used `__getitem__` here instead of `__class_getitem__` but when I looked into why it is documented in the Python docs for `__getitem__` to take precedence.

jpype/_jarray.py
astrelsky: I would argue for the exception message to be changed. As far as I can tell, it is impossible to attempt to instantiate an unspecified array type.

I left the exception message as it was to discuss later.


I'm not sure how to tell the code coverage that most of the paths are covered. I did miss two paths which I'll add tonight.

native/python/pyjp_class.cpp
astrelsky: I'm not going to dig into what exactly is going on but I think it's just a side effect of Jpype's object model. I'm not going to tinker with it and will just leave a comment explaining how we got there.

Thrameos:
I would recommend trying

jpype.JArray(_jpype._JClass)

to hit that code path.  The backend classes all have unoccupied java slots unless the front end has specifically added it during initialization.

jpype/_jarray.py
astrelsky: Well, surprise... I got `List` of the array and not a `List` of the contents. Not sure how I managed to goof that up but I had assumed there was already test coverage since `in` already worked without a problem. In order to get the correct `List` I'd need to do `_jpype.JClass("java.util.Arrays").asList(*self)` which defeats the purpose so I'm going to drop this change.

jpype/_jarray.py
astrelsky: Regarding the code coverage, it was actually correct and I was missing coverage due to a copy/paste mistake.

astrelsky:
> I would recommend trying jpype.JArray(_jpype._JClass) to hit that code path. The backend classes all have unoccupied java slots unless the front end has specifically added it during initialization.

You are correct. I just check and it's actually already protected here. https://github.com/jpype-project/jpype/blob/66f8c6c07e051eb430326418798736004b0f833b/native/python/pyjp_module.cpp#L408


However, if I do `JArray[_jpype._JClass]` I get the `TypeError` from `__class_getitem__` which now says "TypeError: <class '_jpype._JClass'> is not a Java class".

Would you prefer I explicitly check for this case and raise the original TypeError? I can see how the error message in that specific case could confuse someone.

test/jpypetest/test_array.py
```diff
@@ -601,3 +602,24 @@ def testShortcut(self):
...
+
+    def testJArrayInvalidGeneric(self):
+        with self.assertRaises(TypeError):
+            jpype.JArray[object]
+    
+    def testJArrayGenericJClass(self):
+        self.assertEqual(type(JClass[0]), JArray[JClass])
+
+    def testJArrayGenericString(self):
+        self.assertEqual(type(JClass[0]), JArray["java.lang.Class"])
```
pelson, line 621: This one is wild to me - it is like a back-door to ``jp.JClass(...)``.

jpype/_jarray.py
```diff
@@ -92,6 +92,18 @@ def __new__(cls, tp, dims=1):
     def of(cls, array, dtype=None):
         return _jpype.arrayFromBuffer(array, dtype)
 
+    def __class_getitem__(cls, key):
+        if key is _jpype.JClass:
+            # explicit check for JClass
+            # _toJavaClass cannot be used
+            # passing int, float, etc is not allowed
+            key = _jpype._java_lang_Class
+        if isinstance(key, (str, _jpype._java_lang_Class)):
```
pelson, line 101: To me, the `str` based option seems like it is adding a questionable shortcut. I don't know if this is already common practice for other `__class_getitems__` in JPype?

test/jpypetest/test_array.py
```diff
@@ -601,3 +602,24 @@ def testShortcut(self):
...
+
+    def testJArrayGenericStringInvalid(self):
+        with self.assertRaises(TypeError):
+            JArray["foo.bar"] # type: ignore
```
pelson, line 625: I'm surprised there needs to be a type ignore here, given it is fine elsewhere to pass a str?

test/jpypetest/test_array.py
astrelsky: I just used it as a test for a "forward declaration". It could have been any valid class.

test/jpypetest/test_array.py
astrelsky: You will get a warning that `foo.bar` doesn't exist, which happens to be the point of the test.

jpype/_jarray.py
astrelsky: https://peps.python.org/pep-0484/#forward-references

This was the first documentation I found. Just trying to address some of the comments from my phone for now.

jpype/_jarray.pyi
```diff
@@ -0,0 +1,54 @@
...
+        ...
+
+    def __iter__(self):
+        ...
+
+    def __contains__(self, item):
+        # this is implicitly implemented
+        ...
+
+    def __reversed__(self):
```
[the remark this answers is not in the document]
astrelsky: They should already be inferred by the abstract base classes. I put the function stubs in because mypy would otherwise complain that it's not an abstract class but it's not defining those functions. I figured it was better to do this then set the meta class as abc.ABCMeta because it's not accurate.

I don't see a reason to not put in the type hints though, rather do it and ensure any type checker gets it right. Will fix it tonight.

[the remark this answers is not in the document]
astrelsky: As strange as this may sound, I copied the signatures directly from `Python3/Lib/_collections_abc.py` to ensure I got them correct. I'm going to add them anyway though.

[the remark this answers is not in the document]
astrelsky: Please sanity check the applied type hints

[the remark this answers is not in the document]
astrelsky: Ok so after further investigation, I discovered that `isinstance(JInt[5], typing.Collection)` will return `False` even though it meets the criteria. I'm assuming this is because `__contains__` is implicit and unfortunately, for some reason, even if explicitly added to `JArray` it still doesn't have the attribute.

This should finally settle everything I think.

test/jpypetest/test_array.py
pelson: It has to be any valid class available from the namespace in which it is declared. While `java` in this case is valid from this module, it isn't in general.

A better test would be to create a never-before-seen alias, and use that in the test ``from jpype import java as _java_alias`` and use it in `JArray["_java_alias.lang.Class"]`. If following the PEP, it would be expected that this works, but I don't think it will with your current implementation. 

Honestly, I think it is better to rely on ``from __future__ import annotations`` to avoid having to introduce a new API for `JClass` retrieval (i.e. don't support strings directly, since they aren't a short-cut for avoiding imports).

astrelsky: I had some thoughts about using `typing.ForwardRef` because it could be used before the JVM is started. Then it occurred to me that you are correct and just not supporting strings directly is the correct approach.

The main reason I am convinced that not directly supporting strings is correct is because, if you wanted to do it before the JVM is started, all you have to do is the following:

```py
from jpype import JArray

def example(obj: "JArray[java.lang.Object]"):
    ...

```

jpype/_jarray.pyi
```diff
@@ -0,0 +1,59 @@
...
+T = typing.TypeVar('T')
+
+
+class JArray(typing.Generic[T]):
+
+    def __new__(cls, tp: typing.Type[T], dims=1) -> JArray[T]:
+        ...
+
+    @classmethod
+    def of(cls, array, dtype=None):
```
pelson, line 33: This could be typed I think?

pelson (approved):
LGTM :+1:

jpype/_jarray.pyi
astrelsky: Unfortunately a type is considered an illegal value for `typing.Literal` so there isn't much point. It might make sense to add once the minimum supported version is 3.12 and `collections.abc.Buffer` can be used.

astrelsky: Actually I should check if you can do `typing.Type[numpy.int8]` or whatever type in combination with typing.overload and see if it helps determine the output parameter type. If it does I'll add it. If not then I'll just say it didn't work. I won't get to it until tomorrow.

astrelsky: Unfortunately overloads with `typing.Type[JBoolean]`, `typing.Type[JByte]` didn't have the wanted behavior and neither did `def of(cls, array, dtype: typing.Type[T]) -> JArray[T]:`. Unfortunately I think the best that can be done is to stick on the return type hint.

astrelsky: I had an epiphany this morning and realized it would work the way I wanted if I used a different `typing.TypeVar`

pelson:
@Thrameos - this PR is good to go from my perspective. I appreciate you are deep in Python 3.13 work, so let me know if there is anything I can do to help with this.

Thrameos:
@pelson I think long term we need to either formalize the Unstable with extra calls such that they work and get them to move the extra data that they are adding into the preheader space.   Unfortunately, I tested heavily with the allocate with extra memory version they provided and it didn't work  In addition the with extra memory is fundamentally incompatible with most recent changes in Python 3.13.   I am also happy if they just adopted the user data patch which accomplishes the same thing,   Eventually they will get a stable object model which provides for true multiple inheritance and this sort of problem won't haunt us every version.  

I am not sure what it will take to get there.   I can provide some patches or instructions for patching Python which we can provide for Python 3.14.   Moving the inline dict to its proper space was only about 10 lines.  (changing 2 macros, adding a type slot for the preheader allocation, and then the setting the free point.)   But if we can't find a way to get patches accepted it won't get better.  I am unfortunately too high a stress of person to navigate the bureaucracy that is Python (and maintain my very high stress job.)

jpype/_jarray.py
marscher: why not use a binary search here? https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Arrays.html
So you can avoid creating a temporary list.

marscher (approved):

marscher:
Thank you @astrelsky for your first contribution and taking the review process. 
And of course thanks @pelson and @Thrameos for the discussions and code reviews!

PR #1218, opened 2024-09-16
prevent misuse of PyTuple_Pack

astrelsky:
This prevents simple mistakes such as the following

https://github.com/jpype-project/jpype/blob/66f8c6c07e051eb430326418798736004b0f833b/native%2Fpython%2Fpyjp_number.cpp#L390

### DIFF
--- a/native/python/pyjp_number.cpp
+++ b/native/python/pyjp_number.cpp
@@ -270,9 +270,8 @@ static PyObject *PyJPBoolean_new(PyTypeObject *type, PyObject *args, PyObject *k
 		return nullptr;
 	}
 	int i = PyObject_IsTrue(PyTuple_GetItem(args, 0));
-	PyObject *args2 = JPPyTuple_Pack(PyLong_FromLong(i));
-	self = JPPyObject::call(PyLong_Type.tp_new(type, args2, kwargs));
-	Py_DECREF(args2);
+	JPPyObject args2 = JPPyTuple_Pack(PyLong_FromLong(i));
+	self = JPPyObject::call(PyLong_Type.tp_new(type, args2.get(), kwargs));
 	JPClass *cls = PyJPClass_getJPClass((PyObject*) type);
 	if (cls == nullptr)
 	{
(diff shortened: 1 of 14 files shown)


Thrameos:
Minor suggestion.... given that all uses of this function should be placed in a JPPyObject wrapper via ``call`` to we don't leak it would seem like combining the two would make sense.   It would take some minor reworking of a few sections which were already dangerous for exceptions.

astrelsky:
> Minor suggestion.... given that all uses of this function should be placed in a JPPyObject wrapper via ``call`` to we don't leak it would seem like combining the two would make sense.   It would take some minor reworking of a few sections which were already dangerous for exceptions.

Agreed. I'll do this tonight.

native/python/pyjp_number.cpp
```diff
@@ -404,8 +399,8 @@ JPPyObject PyJPNumber_create(JPJavaFrame &frame, JPPyObject& wrapper, const JPVa
 		jlong l = 0;
 		if (value.getValue().l != nullptr)
 			l = frame.CallBooleanMethodA(value.getJavaObject(), context->_java_lang_Boolean->m_BooleanValueID, nullptr);
-		PyObject *args = PyTuple_Pack(1, PyLong_FromLongLong(l));
-		return JPPyObject::call(PyLong_Type.tp_new((PyTypeObject*) wrapper.get(), args, nullptr));
```
astrelsky, line 408: Was this and the two below a leak?

astrelsky:
The changes were mostly painless. I only had to think about one change which was in `PyJPClass_FromSpecWithBases`. I left a comment where it should not be used in that function.

native/python/pyjp_number.cpp
Thrameos: Seems likely.  Once had leaks everwhere.   Hence need to write the C++ wrapper for python API.  Clearly this was missed.

Thrameos (approved):

Thrameos:
Merge for upcoming release.

PR #1220, opened 2024-09-30
fix JUnpickler BufferOverflowException

tachyonicClock:


marscher:
Thanks @tachyonicClock for the fix and the regression test!

test/jpypetest/test_pickle.py
```diff
@@ -76,6 +77,35 @@ def testMixed(self):
...
 
+    def testMultiObject(self):
+        """Regression test for https://github.com/jpype-project/jpype/issues/1201
+        
+        Issue occurs when a python object contains multiple java objects above
+        a certain size. The issue occurs because of a buffer overflow in
+        ``native/java/org/jpype/pickle/ByteBufferInputStream.java``.
+        """
+        JString = jpype.JClass("java.lang.String")
+        filename = "test.pic"
```
marscher, line 88: please use a `NamedTemporaryFile` here so it gets deleted after the test.

test/jpypetest/test_pickle.py
```diff
@@ -23,15 +23,16 @@
 import common
 
 
-import unittest
-
-
 def dump(fname):
```
marscher, line 26: this seems unused

marscher:
@tachyonicClock Please add a line to the changelog then I think we're good to go.
@Thrameos LGTM, what do you think?

Thrameos:
@marscher I would rather have it create static sized ByteBuffers of something approaching a page size in slices if needed rather than one for every put where the user is going to force the size of each buffer.   Imagine pushing byte[1] over and over and getting a list of ByteBuffers with one item.  ByteBufferes are fairly expensive objects.   They unfortunately are not growable in Java and we must have landed on an edge condition.  

Overall the solution is acceptable.  I would have selected a higher efficiency route, but working is better than not.

marscher:
> @marscher I would rather have it create static sized ByteBuffers of something approaching a page size in slices if needed rather than one for every put where the user is going to force the size of each buffer. Imagine pushing byte[1] over and over and getting a list of ByteBuffers with one item. ByteBufferes are fairly expensive objects. They unfortunately are not growable in Java and we must have landed on an edge condition.
> 
> Overall the solution is acceptable. I would have selected a higher efficiency route, but working is better than not.

The use case of serialization we never call put(1 byte), but the amount of bytes of the serialized object. Anyhow I've added an impl (which is not yet functional) with a growing buffer to 1201 branch in this repo. I will fix it tomorrow.

marscher (approved):

PR #1224, opened 2024-10-11
free threaded build

astrelsky:
As far as I can tell, jpype is already thread safe as it was necessary since Java is a normal programming that isn't scared of more than one thread.

Anyone looking to use the experimental free threaded build of python should be anticipating to encounter some, uh, problems anyway. [problem example](https://github.com/python/cpython/issues/125139)

@marscher I'm a complete dinosaur when it comes to ci/cd stuff so I didn't touch any of it. In order to build the free threaded version on windows I needed to do `python3.13t -m pip install -e .`, so I'm assuming it's a similar situation with the other operating systems.

### DIFF
--- a/native/common/jp_tracer.cpp
+++ b/native/common/jp_tracer.cpp
@@ -150,7 +150,7 @@ void JPypeTracer::tracePythonObject(const char* msg, PyObject* ref)
 	if (ref != nullptr)
 	{
 		std::stringstream str;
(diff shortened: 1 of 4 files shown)


marscher:
Thanks for this interesting contribution. Is the thread-free branch enabled by default when using Python 3.13 or how is toggled?

Thrameos (approved):
LGTM

astrelsky:
> LGTM

Thank you

PR #1232, opened 2024-11-10
Releases/v1.5.1

Thrameos:
Testing for release.


Thrameos:
@marscher  I am still getting random faults in the test matrix.   I am not sure if it is us or something underlaying in Python.   I will likely proceed with the release as none of my local testing shows anything unstable after many runs.  My general feeling is there may be an instability in some backported changes in the Python release system.   

It will take a while getting the release pipeline up the date.   Some of the changes in the python packaging system damaged the  release system and a number of images have shifted.  Hopefully I have some images by the end of the weekend.

Thrameos:
@pelson it seems like the conversion to pyproject.toml was incomplete.   None of the builds are properly under the release system.  (I believe it was changed in #1133)   

Unfortunately, the revised system is entirely opaque to me thus I am not sure how to proceed.  Under the setup.py system it was easy to customize to include files that were outside the scope of the regular build system. (after all Python doesn't account for java files as part of the build process.)  I can try to do things like moving org.jpype.jar to be under the python module, but that may break other things.  

I may need to add you to azure dev cloud in order for you to see the logs.  If you have a azure account that please look through the github commits to get my email and drop me a email so I can properly add you (no need to post it here.)

Thrameos:
You should be able to get all the artifacts from build.  I believe to went wrong executing the source dist which is supposed to build the jar and collect all the distibuted sources.  If something is missing from the step then everything downstrwam will fail.

The procedure I would use would be pull the source artifact and inspect it.  The disable tge rest if the release build (to save time or ypu can end up with 10 failed builds queued up), then trigger a new release with changes until tge source dist succeeds.

PR #1235, opened 2024-11-17
Fix for GC on windows

### LINKED ISSUE #1197: Starting JVM with jpype severely slows down pure python code in the same process
After a call to `jpype.startJVM()`, all python code is slowed down by 2x even when there are absolutely no other interactions with java/jpype. 
This is further aggravated the higher one sets heap size of the started JVM.

The problem seems to be the mechanism by which every python GC triggers a java GC. Adding this line: `jpype._jpype._collect = lambda *args, **kwargs: None` before startJVM makes the issue go away.

<details>
  <summary>Reproduction script</summary>

  ```py
  import argparse
  import random
  import time
  
  
  def generate_garbage():
      garbage = {}
      for _ in range(100_000):
          e = {}
          for _ in range(10):
              e[random.randbytes(4)] = [random.randint(10_000, 50_000)]
          garbage[random.randbytes(10)] = e
  
      return garbage
  
  
  def main():
      parser = argparse.ArgumentParser()
      parser.add_argument('--disable-jpype-collect', action='store_true', default=False)
      parser.add_argument('--start-jvm', action='store_true', default=False)
      args = parser.parse_args()
  
      import jpype
      if args.disable_jpype_collect:
          jpype._jpype._collect = lambda *args, **kwargs: None
          if not args.start_jvm:
              parser.error('--disable-collect requires --jvm')
      if args.start_jvm:
          jpype.startJVM("-Xmx8g", "-Xms8g")
  
      start = time.perf_counter()
      for _ in range(3):
          generate_garbage()
      print(f'took {time.perf_counter() - start:.2f} seconds to generate garbage')
  
  
  if __name__ == '__main__':
      main()
  ```

</details>

```shell
$ python test_jpype_slow.py 
took 2.78 seconds to generate garbage

$ python test_jpype_slow.py --start-jvm
took 5.02 seconds to generate garbage

$ python test_jpype_slow.py --start-jvm --disable-jpype-collect
took 2.74 seconds to generate garbage
```

Reproduced with jpype 1.5.0, openjdk 22, python 3.12 on Windows 11

Thrameos:
The issue appears to be with the Python bookkeeping.   When the Python memory is growing faster than the Java GC can execute the Java GC gets triggered many times resulting in a serious slow down.   I upped the limit after a forced Java trigger.   I also added some instrumentation to make diagnosis easier.   It is not a 100% but does address the issue.

Fixes #1197

### DIFF
--- a/native/common/jp_gc.cpp
+++ b/native/common/jp_gc.cpp
@@ -154,6 +154,14 @@ void JPGarbageCollection::init(JPJavaFrame& frame)
 	_SystemClass = (jclass) frame.NewGlobalRef(frame.FindClass("java/lang/System"));
 	_gcMethodID = frame.GetStaticMethodID(_SystemClass, "gc", "()V");
 
+    jclass ctxt = frame.getContext()->m_ContextClass.get();
+    _ContextClass = ctxt;
+	_totalMemoryID = frame.GetStaticMethodID(ctxt, "getTotalMemory", "()J");
+	_freeMemoryID = frame.GetStaticMethodID(ctxt, "getFreeMemory", "()J");
+	_maxMemoryID = frame.GetStaticMethodID(ctxt, "getMaxMemory", "()J");
+	_usedMemoryID = frame.GetStaticMethodID(ctxt, "getUsedMemory", "()J");
+	_heapMemoryID = frame.GetStaticMethodID(ctxt, "getHeapMemory", "()J");
+
 	running = true;
 	high_water = getWorkingSize();
 	limit = high_water + DELTA_LIMIT;
@@ -237,10 +245,24 @@ void JPGarbageCollection::onEnd()
 		Py_ssize_t pred = current + 2 * (current - last);
 		last = current;
 		if ((Py_ssize_t) pred > (Py_ssize_t) limit)
+		{
 			run_gc = 2;
+            limit = high_water + (high_water>>3) + 8 * (current - last);
+        }
 
-		//		printf("consider gc %d (%ld, %ld, %ld, %ld) %ld\n", run_gc,
-		//				current, low_water, high_water, limit, limit - pred);
+#if 0
+        {
+			JPJavaFrame frame = JPJavaFrame::outer(m_Context);
+		    jlong totalMemory = frame.CallStaticLongMethodA(_ContextClass, _totalMemoryID, nullptr);
+		    jlong freeMemory = frame.CallStaticLongMethodA(_ContextClass, _freeMemoryID, nullptr);
+		    jlong maxMemory = frame.CallStaticLongMethodA(_ContextClass, _maxMemoryID, nullptr);
+		    jlong usedMemory = frame.CallStaticLongMethodA(_ContextClass, _usedMemoryID, nullptr);
+		    jlong heapMemory = frame.CallStaticLongMethodA(_ContextClass, _heapMemoryID, nullptr);
+			printf("consider gc run=%d (current=%ld, low=%ld, high=%ld, limit=%ld) %ld\n", run_gc,
+					current, low_water, high_water, limit, limit - pred);
+            printf(" java total=%ld free=%ld max=%ld used=%ld heap=%ld\n", totalMemory, freeMemory, maxMemory, usedMemory, heapMemory);
+        }
+#endif
 
 		if (run_gc > 0)
 		{
--- a/native/common/jp_gc.cpp
+++ b/native/common/jp_gc.cpp
@@ -154,8 +154,8 @@ void JPGarbageCollection::init(JPJavaFrame& frame)
 	_SystemClass = (jclass) frame.NewGlobalRef(frame.FindClass("java/lang/System"));
 	_gcMethodID = frame.GetStaticMethodID(_SystemClass, "gc", "()V");
 
-    jclass ctxt = frame.getContext()->m_ContextClass.get();
-    _ContextClass = ctxt;
+	jclass ctxt = frame.getContext()->m_ContextClass.get();
+	_ContextClass = ctxt;
 	_totalMemoryID = frame.GetStaticMethodID(ctxt, "getTotalMemory", "()J");
 	_freeMemoryID = frame.GetStaticMethodID(ctxt, "getFreeMemory", "()J");
 	_maxMemoryID = frame.GetStaticMethodID(ctxt, "getMaxMemory", "()J");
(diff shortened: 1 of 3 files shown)


native/common/jp_gc.cpp
```diff
@@ -237,10 +245,24 @@ void JPGarbageCollection::onEnd()
...
 		last = current;
 		if ((Py_ssize_t) pred > (Py_ssize_t) limit)
+		{
 			run_gc = 2;
+            limit = high_water + (high_water>>3) + 8 * (current - last);
+        }
 
-		//		printf("consider gc %d (%ld, %ld, %ld, %ld) %ld\n", run_gc,
-		//				current, low_water, high_water, limit, limit - pred);
+#if 0
```
marscher, line 253: Did you forget to commit something? This is unconditional true, I think it should be activated upon instrumentation, right?

It'd also be nice to use a consistent formatting (tabs vs spaces)

native/common/jp_gc.cpp
```diff
@@ -237,10 +245,24 @@ void JPGarbageCollection::onEnd()
 		Py_ssize_t pred = current + 2 * (current - last);
 		last = current;
 		if ((Py_ssize_t) pred > (Py_ssize_t) limit)
+		{
 			run_gc = 2;
+            limit = high_water + (high_water>>3) + 8 * (current - last);
```
marscher, line 250: can you please explain this heuristic? What does it aim for?

Thrameos: The current logic is that if Python memory was rising above the high water (or predicted to) then it would trigger then java gc.   When the python gc is complete we were supposed to reset the high water so that we won't run the java gc again for a while.  

In Windows, it appears that rather than stop the world Python is still running. So what is happening is that in the example sent we triggered the java gc over and over and over.   To prevent that I have to raise the limit above the waves so that it won't trigger unless garbage is actually piling up.  Here is a graph of before and after.

image

native/common/jp_gc.cpp
Thrameos: The instrumentation is to barf out all of the python gc calls and all the java calls.   I am not sure anyone is likely to want that unless they are trying to debug the process.   So I currently have it conditionally compiled as false all the time.

I will try to fix up the tabs.   The problem is that some files we have are currently tab spaced and others are spaced only and I have to keep switching the settings on vim based on which file I am editing.

marscher (approved):
Thank you very much for the explanation and consistent format!

marscher:
@Thrameos Do you want this to be included in 1.5.1?

native/common/jp_gc.cpp
astrelsky: I've found it rather painful as well while working on extension support. It got to the point where I've said to hell with it and reformatted entire files and decided to just deal with fixing it later.

marscher: We should come up with some codestyle probably. Did you choose spaces over tabs or vice versa @astrelsky?

astrelsky: > We should come up with some codestyle probably. Did you choose spaces over tabs or vice versa @astrelsky?

My ide settings were set to use this formatter so it only took me two clicks to reformat a file. https://github.com/NationalSecurityAgency/ghidra/blob/master/eclipse/GhidraEclipseFormatter.xml

I learned java from working with Ghidra so I lean towards their code style with minor divergence. In many ways it directly conflicts with the styles used in jpype. I've become very nitpicky about formatting.

To actually answer your question, I used tabs (width of 4 spaces). At the end of the day I'll follow whatever code style you set up. I was aware I was creating extra work for myself by applying the formatter when I did it.

I don't think I should have any say here. As long as it's consistent I'll follow it :joy:

Thrameos:
I think it should go in.   It was a last minute fix for a critical speed bug.    It may be a while before we have another release.

native/common/jp_gc.cpp
Thrameos: It was so much easier when netbeans had native C++ support.   My only requirement is please no K&R type brace styles.   My preferences is always the GNU coding style.  https://www.gnu.org/prep/standards/html_node/Formatting.html

astrelsky: > It was so much easier when netbeans had native C++ support.
> 
> 

If you're wanting something that supports all 3 languages I've used vscode with good results. Setup can be a hassle, especially with java, but I have managed to have a python debugger, java debugger and c++ debugger all attached at the same time in the same vscode instance while running the jpype python tests through vscode. I'm not necessarily recommending it, because project configuration for java and c++ can be extremely finicky, just putting it out there.

Thrameos: Before the transition to Apache I had Netbeans configured to all of the languages, with debugging, autoformat, etc.    It was very easy to compile and test.   After Apache C++ extensions decayed to the point they couldn't even open the files and reformat them.   Thus I have been forced back to stone knives and bear skins (well using vim and doing everything from command line).   It has been a serious reduction in what I can do as refactoring is out of the question with such a setup.  

It appears that netbeans has finally readded C++ support, so I will try to get it spun up again.

astrelsky: > Before the transition to Apache I had Netbeans configured to all of the languages, with debugging, autoformat, etc.    It was very easy to compile and test.   After Apache C++ extensions decayed to the point they couldn't even open the files and reformat them.   Thus I have been forced back to stone knives and bear skins (well using vim and doing everything from command line).   It has been a serious reduction in what I can do as refactoring is out of the question with such a setup.  
> 
> It appears that netbeans has finally readded C++ support, so I will try to get it spun up again.   

I was initially doing it in a similar fashion with having to rebuild the extension using `pip install -e .` and having to recompile everything every single time. I eventually just dropped in a cmake file so I can build the native code with ninja and a gradle script for the java. It's much better not having to wait two minutes to rebuild everything for a minor change. IMO, this is mostly just a side effect of how python builds c extensions in a platform independent way.

Thrameos: It wouldn't be much of an effort for them to make an option like my "python setup.py build_ext --makefile" which just takes all the instructions they are running on a system and dumping it out in makefile format.   I would never have made it this far in the project without it.   Unfortunately with the motion towards a more opaque process it is getting harder to accomplish such development.

astrelsky: > It wouldn't be much of an effort for them to make an option like my "python setup.py build_ext --makefile" which just takes all the instructions they are running on a system and dumping it out in makefile format.   I would never have made it this far in the project without it.   Unfortunately with the motion towards a more opaque process it is getting harder to accomplish such development.

fwiw cmake does a pretty decent job with python extensions. The only thing that was lacking when I set it up was free threaded python support. It might be able to generate the netbeans project for you too.

marscher: I'd be in favor of tools which are independent of the used IDE. Otherwise everybody would be forced to use the same IDE, right?

astrelsky: > I'd be in favor of tools which are independent of the used IDE. Otherwise everybody would be forced to use the same IDE, right?

I'm always fighting someone to make something os/ide independent, even at work. Everyone should be able to use what they're most comfortable and productive with.

Thrameos: I just want a beautifier to run from command line when we check in.   So long as the style is close enough that anyone's personal IDE doesn't go insane trying process it then all is good.  Double points if the beautifier is a python module so that we can just pull it in with requirements rather than installing another tool.  I distributed the Netbeans IDE files in the project directory, but there is no expectation that anyone needs to use them.

marscher: Right, so lets run pre-commit on PRs, so everybody can do whatever they love offline. When the code goes online, we enforce the formatting (this can be done automatically, but would involve rebasing to origin after a CI run).

It is also called evolution when the young ones wants to introduce state of the art stuff. I was also young one day and my bosses hated me for introducing SVN over CVS =) Sooo complicated. Several years later we had the same thingy with Git. Now everybody uses it. So after all I wouldn't say it's a bad thing to modernize once in a while.

The productivity drop is preliminary and would probably benefit in the future.

Thrameos:
I wish it was true on the productivity.   But now a days it seems like people are being spoon fed general purpose tools that they use for every class.   My old netbeans with ant does everything their tools do plus a huge amount more.   If they wreck my performance and force me to waste hour after hour dealing with groovy/maven/random tool which they actually don’t even know how to use themselves, then it is a loss.  Better to retool them to use tools that work.  If they can use another IDE without busting up my 60k lines of code I am maintaining then great, if not then use my tools.

marscher:
It seems you're kind angry about that. Wasting time on something which does not work or boost productivity is indeed a huge loss.

Any how, please stop cross posting issues with your greatly working email client :P *just being sarcastic*

astrelsky:
> I wish it was true on the productivity. But now a days it seems like people are being spoon fed general purpose tools that they use for every class. My old netbeans with ant does everything their tools do plus a huge amount more. If they wreck my performance and force me to waste hour after hour dealing with groovy/maven/random tool which they actually don’t even know how to use themselves, then it is a loss. Better to retool them to use tools that work. If they can use another IDE without busting up my 60k lines of code I am maintaining then great, if not then use my tools.

Could be worse, I've seen people use make for Java :joy:. Does it work, yes, but is it the best tool for the job, no not really.

Issue #1242, opened 2024-11-22
Failure to start JVM with jpype 1.5.1 on Windows


villares:
With 1.5.0 on this env I could start the JVM.
```
C:\Users\alexandre.villares>python
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import jpype
>>> jpype.__version__
'1.5.0'
>>> jpype.startJVM()
Picked up JAVA_TOOL_OPTIONS: -Djava.vendor="Sun Microsystems Inc."
>>> jpype.getDefaultJVMPath()
'C:\\Users\\alexandre.villares\\Downloads\\thonny-4.1.6-portable\\user_data\\jdk-17\\bin\\server\\jvm.dll'
>>>
```
Updating to jpype 1.5.1 it fails:
```
C:\Users\alexandre.villares>python -m pip install jpype1 --upgrade
Requirement already satisfied: jpype1 in c:\users\alexandre.villares\downloads\thonny-4.1.6-portable\lib\site-packages (1.5.0)
Collecting jpype1
  Downloading jpype1-1.5.1-cp310-cp310-win_amd64.whl.metadata (5.0 kB)
Requirement already satisfied: packaging in c:\users\alexandre.villares\downloads\thonny-4.1.6-portable\lib\site-packages (from jpype1) (24.2)
Downloading jpype1-1.5.1-cp310-cp310-win_amd64.whl (356 kB)
Installing collected packages: jpype1
  Attempting uninstall: jpype1
    Found existing installation: JPype1 1.5.0
    Uninstalling JPype1-1.5.0:
      Successfully uninstalled JPype1-1.5.0
  WARNING: Failed to remove contents in a temporary directory 'C:\Users\alexandre.villares\AppData\Local\Temp\pip-uninstall-lxb7b830'.
  You can safely remove it manually.
Successfully installed jpype1-1.5.1

C:\Users\alexandre.villares>python
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import jpype
>>> jpype.__version__
'1.5.1'
>>> jpype.startJVM()
[...]

PR #1245, opened 2024-11-27
No agent

Thrameos:
This resolves the issued with the agent.  But I would like to fix the non-ascii path issue at the same time.  So additional work is still required.

Comments and reviews welcome.

### DIFF
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -109,26 +109,21 @@ def isJVMStarted():
     return _jpype.isStarted()
 
 
-def _getOption(args, var, sep) -> list[str]:
+def _getOption(args, var, sep=None, keep=False) -> list[str]:
     """ Get an option and remove it from the current jvm arguments list """
     for i,v in enumerate(args):
         if not isinstance(v, str):
             continue
-        _, _, value = v.partition('%s='%s)
+        _, _, value = v.partition('%s='%var)
         if value:
-            del args[i]
-            return value.split(sep)
+            if not keep:
+                del args[i]
+            if sep is not None:
+                return value.split(sep)
+            return value
     return []
 
-
-def _hasSystemClassLoader(args) -> bool:
-    for i in args:
-        if isinstance(i, str) and i.startswith('-Djava.system.class.loader'):
-            return True
-    return False
-
-
-def _handleClassPath(
+def _expandClassPath(
     classpath: typing.Union[typing.Sequence[_PathOrStr], _PathOrStr, None] = None
 ) -> typing.Sequence[str]:
     """
@@ -177,6 +172,7 @@ def startJVM(
     ignoreUnrecognized: bool = False,
     convertStrings: bool = False,
     interrupt: bool = not interactive(),
+    agent: bool = False,
 ) -> None:
     """
     Starts a Java Virtual Machine.  Without options it will start
@@ -196,7 +192,6 @@ def startJVM(
       classpath (str, PathLike, [str, PathLike]): Set the classpath for the JVM.
         This will override any classpath supplied in the arguments
         list. A value of None will give no classpath to JVM.
-      modulepath (str, PathList, [str, PathList]): Set the module path for JVM.
       ignoreUnrecognized (bool): Option to ignore
         invalid JVM arguments. Default is False.
       convertStrings (bool): Option to force Java strings to
@@ -210,6 +205,9 @@ def startJVM(
         transfer control to Python rather than halting.  If
         not specified will be False if Python is started as
         an interactive shell.
+      agent (bool): Start as agent.  This allows for certain
+        privilaged operations, but required java.instrumentation
+        module.
 
     Raises:
       OSError: if the JVM cannot be started or is already running.
@@ -222,7 +220,8 @@ def startJVM(
     if _JVM_started:
         raise OSError('JVM cannot be restarted')
 
-    has_classloader = _hasSystemClassLoader(jvmargs)
+    # Convert to list
+    jvmargs = list(jvmargs)
 
     # JVM path
     if jvmargs:
@@ -254,49 +253,55 @@ def startJVM(
         # Not specified at all, use the default classpath.
         classpath = _classpath.getClassPath()
 
-    # Modulepath handling
-    old_modulepath = _getOption(jvmargs, "--module-path", _classpath._SEP)
-    if old_modulepath:
-        # Old style, specified in the arguments
-        if modulepath is not None:
-            # Cannot apply both styles, conflict
-            raise TypeError('modulepath specified twice')
-        modulepath = old_modulepath
-    if modulepath is not None:
-        mp = _classpath._SEP.join(_handleClassPath(modulepath))
-        extra_jvm_args += ['--module-path=%s'%mp ]
+# Code for 1.6 release when we add module support
+#    # Modulepath handling
+#    old_modulepath = _getOption(jvmargs, "--module-path", _classpath._SEP)
+#    if old_modulepath:
+#        # Old style, specified in the arguments
+#        if modulepath is not None:
+#            # Cannot apply both styles, conflict
+#            raise TypeError('modulepath specified twice')
+#        modulepath = old_modulepath
+#    if modulepath is not None:
+#        mp = _classpath._SEP.join(_expandClassPath(modulepath))
+#        extra_jvm_args += ['--module-path=%s'%mp ]
 
     # Get the support library
-    supportLib = os.path.join(os.path.dirname(os.path.dirname(__file__)), "org.jpype.jar")
-    if not os.path.exists(supportLib):
-        raise RuntimeError("Unable to find org.jpype.jar support library at " + supportLib)
+    support_lib = os.path.join(os.path.dirname(os.path.dirname(__file__)), "org.jpype.jar")
+    if not os.path.exists(support_lib):
+        raise RuntimeError("Unable to find org.jpype.jar support library at " + support_lib)
 
-    late_load = not has_classloader
-    if classpath:
-        cp = _classpath._SEP.join(_handleClassPath(classpath))
-        if cp.isascii():
-            # no problems
-            extra_jvm_args += ['-Djava.class.path=%s'%cp ]
-            late_load = False
-        elif has_classloader:
-            # https://bugs.openjdk.org/browse/JDK-8079633?jql=text%20~%20%22ParseUtil%22
-            raise ValueError("system classloader cannot be specified with non ascii characters in the classpath")
-        elif supportLib.isascii():
-            # ok, setup the jpype system classloader and add to the path after startup
-            # this guarentees all classes have the same permissions as they did in the past
-            extra_jvm_args += [
-                '-Djava.system.class.loader=org.jpype.classloader.JpypeSystemClassLoader',
-                '-Djava.class.path=%s'%supportLib
-            ]
-        else:
-            # We are screwed no matter what we try or do.
-            # Unfortunately the jdk maintainers don't seem to care either.
-            # This bug is almost 10 years old and spans 16 jdk versions and counting.
-            # https://bugs.openjdk.org/browse/JDK-8079633?jql=text%20~%20%22ParseUtil%22
-            raise ValueError("jpype jar must be ascii to add to the system class path")
+    system_class_loader = _getOption(jvmargs, "-Djava.system.class.loader", keep=True)
+    late_load = not system_class_loader
 
+    java_class_path = _expandClassPath(classpath)
+    java_class_path.append(support_lib)
+    java_class_path = _classpath._SEP.join(java_class_path)
 
-    extra_jvm_args += ['-javaagent:' + supportLib]
+    # Make sure our module is always on the classpath
+    if java_class_path.isascii():
+        # no problems
+        extra_jvm_args += ['-Djava.class.path=%s'%java_class_path ]
+        late_load = False
+    elif system_class_loader:
+        # https://bugs.openjdk.org/browse/JDK-8079633?jql=text%20~%20%22ParseUtil%22
+        raise ValueError("system classloader cannot be specified with non ascii characters in the classpath")
+    elif support_lib.isascii():
+        # ok, setup the jpype system classloader and add to the path after startup
+        # this guarentees all classes have the same permissions as they did in the past
+        extra_jvm_args += [
+            '-Djava.system.class.loader=org.jpype.classloader.JpypeSystemClassLoader',
+            '-Djava.class.path=%s'%support_lib
+        ]
+    else:
+        # We are screwed no matter what we try or do.
+        # Unfortunately the jdk maintainers don't seem to care either.
+        # This bug is almost 10 years old and spans 16 jdk versions and counting.
+        # https://bugs.openjdk.org/browse/JDK-8079633?jql=text%20~%20%22ParseUtil%22
+        raise ValueError("jpype jar must be ascii to add to the system class path")
+
+    if agent:
+        extra_jvm_args += ['-javaagent:' + support_lib]
 
     try:
         import locale
@@ -339,7 +344,7 @@ def startJVM(
     if late_load and classpath:
         # now we can add to the system classpath
         cl = _jpype.JClass("java.lang.ClassLoader").getSystemClassLoader()
-        for cp in _handleClassPath(classpath):
+        for cp in _expandClassPath(classpath):
             cl.addPath(_jpype._java_lang_String(cp))
 
 
--- a/jpype/_core.py
+++ b/jpype/_core.py
@@ -109,13 +109,15 @@ def isJVMStarted():
     return _jpype.isStarted()
 
 
-def _getOldClassPath(args) -> list[str]:
-    for i in args:
-        if not isinstance(i, str):
+def _getOption(args, var, sep) -> list[str]:
+    """ Get an option and remove it from the current jvm arguments list """
+    for i,v in enumerate(args):
+        if not isinstance(v, str):
             continue
-        _, _, classpath = i.partition('-Djava.class.path=')
-        if classpath:
-            return classpath.split(_classpath._SEP)
+        _, _, value = v.partition('%s='%s)
+        if value:
+            del args[i]
+            return value.split(sep)
     return []
 
 
@@ -160,10 +162,6 @@ def _handleClassPath(
     return out
 
 
-def _removeClassPath(args) -> tuple[str]:
-    return tuple(arg for arg in args if not str(arg).startswith("-Djava.class.path"))
-
-
 _JVM_started = False
 
 
@@ -175,6 +173,7 @@ def startJVM(
     *jvmargs: str,
     jvmpath: typing.Optional[_PathOrStr] = None,
     classpath: typing.Union[typing.Sequence[_PathOrStr], _PathOrStr, None] = None,
+    modulepath: typing.Union[typing.Sequence[_PathOrStr], _PathOrStr, None] = None,
     ignoreUnrecognized: bool = False,
     convertStrings: bool = False,
     interrupt: bool = not interactive(),
@@ -197,6 +196,7 @@ def startJVM(
       classpath (str, PathLike, [str, PathLike]): Set the classpath for the JVM.
         This will override any classpath supplied in the arguments
         list. A value of None will give no classpath to JVM.
+      modulepath (str, PathList, [str, PathList]): Set the module path for JVM.
       ignoreUnrecognized (bool): Option to ignore
         invalid JVM arguments. Default is False.
       convertStrings (bool): Option to force Java strings to
@@ -224,7 +224,6 @@ def startJVM(
 
     has_classloader = _hasSystemClassLoader(jvmargs)
 
-
     # JVM path
     if jvmargs:
         # jvm is the first argument the first argument is a path or None
@@ -240,8 +239,11 @@ def startJVM(
         # Allow the path to be a PathLike.
         jvmpath = os.fspath(jvmpath)
 
+    # Handle strings and list of strings.
+    extra_jvm_args: list[str] = []
+
     # Classpath handling
-    old_classpath = _getOldClassPath(jvmargs)
+    old_classpath = _getOption(jvmargs, "-Djava.class.path", _classpath._SEP)
     if old_classpath:
         # Old style, specified in the arguments
         if classpath is not None:
@@ -252,9 +254,19 @@ def startJVM(
         # Not specified at all, use the default classpath.
         classpath = _classpath.getClassPath()
 
-    # Handle strings and list of strings.
-    extra_jvm_args: list[str] = []
+    # Modulepath handling
+    old_modulepath = _getOption(jvmargs, "--module-path", _classpath._SEP)
+    if old_modulepath:
+        # Old style, specified in the arguments
+        if modulepath is not None:
+            # Cannot apply both styles, conflict
+            raise TypeError('modulepath specified twice')
+        modulepath = old_modulepath
+    if modulepath is not None:
+        mp = _classpath._SEP.join(_handleClassPath(modulepath))
+        extra_jvm_args += ['--module-path=%s'%mp ]
 
+    # Get the support library
     supportLib = os.path.join(os.path.dirname(os.path.dirname(__file__)), "org.jpype.jar")
     if not os.path.exists(supportLib):
         raise RuntimeError("Unable to find org.jpype.jar support library at " + supportLib)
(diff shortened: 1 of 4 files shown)


Thrameos:
@tachyonicClock I am taking another crack at the non-ascii paths.   To deal with them I am setting up a temporary directory and just copying the required jar file in place.    I will then try to bootstrap the rest.  

Thus far everything but services are working.

tachyonicClock:
I believe using a temporary directory is likely the most effective workaround.

**Linking vs. Copying:** If possible, we should try to hard link the libraries rather than copying them. This will require a fallback for people linking between disk drives, since you cannot hard-link across storage devices. Since you can symbolic link directories across drives we could try that but it might get messy. Symbolic linking or hard-linking will add close to zero startup time.

**Uniform Behavior**: Branching on `java_class_path.isascii()` creates two different behaviors. It would reduce the possibility of bugs to simply perform the workaround for all paths, regardless of whether they need it. This way, we can catch bugs in the workaround.

**Custom Loaders:** For people who need custom loaders or things on the class path at startup, we could expose a more flexible startJVM function:
  ```
  startJVM(..., classpath: List[Path], system_classpath: List[Path])
  ```
  * `classpath`: The post-startup Unicode-compatible approach to adding files to the classpath.
  * `system_classpath`: The approach that adds files to the JVM classpath options (`-Djava.class.path`). This should perform the temporary directory hack for all system class paths. This way, people can have custom class loaders. The support library can be added to this list.
  * A downside is that it would break backwards compatibility.

astrelsky:
The previous fix that included the agent was mis-behaving because dlls from a different jdk vendor were being loaded with the specified jvm.dll. Why this issue never appeared in the past, I'm not so sure.

Thrameos:
@tachyonicClock   I agree that uniform behavior would be best.  I would make everything go through our system class loader if I through that we could get away with it.   However, there is a warning that is making me nervous...

OpenJDK 64-Bit Server VM warning: Archived non-system classes are disabled because the java.system.class.loader property is specified (value = "org.jpype.classloader.DynamicClassLoader"). To use archived non-system classes, this property must not be set

I have no idea what a Archived non-system class would be or what they are used for. 

https://github.com/NationalSecurityAgency/ghidra/issues/2400

As to linking that is really only a unix thing.   For now I will copy.  I may be able to unlink it after starting but there are risks with that.  More experiments needed.

I copied my whole jpype install under a ./unicode_à😎  directory and trying to get it all working there.   Still a bunch of edge cases.  But if we can get all the tests running there than we should be set.

Thrameos:
Still running into edge cases here.   

Down to two issues.
1) A directory of classes is being dealt with differently than a set of jars.   This is breaking some of our test cases.  I am not sure how the system class loader was dealing with these before.
2) The SQL tests are broken.  Unfortunately, this may mean that our service test was inadequate as JDB are using the service loaded.

native/java/org/jpype/classloader/DynamicClassLoader.java
```diff
@@ -25,16 +27,69 @@
...
+        }
+      }
+      off = next + 1;
+    } while (next != -1);
+
+    System.out.println("jpype.class.path " + cp);
+    System.clearProperty("jpype.class.path");
+    System.setProperty("java.class.path", cp);
+    return path.toArray(new URL[0]);
+  }
```
Thrameos, line 80: @tachyonicClock This section allows for us to load the non-ascii path just like a normal one.  Thus removing the need for late load logic.

Thrameos (commented):
Still a work in progress but getting closer.

Thrameos:
I understand the two issues.   The key problem is that callerSensitive picks up the class loader for the class which was loaded not the one called from Python.  So as org.jpype is loaded from the system classloader it gets the broken ApplicationClassLoader rather than our system specified class loader.  

As usually there are clear gaps in the Java api which are driving this.  Only in the case that ApplicationClassLoader was use can this work perfectly.  

I can make dbapi2 work if I force the caller sensitive logic off, but I am not sure if that works in all cases.

Thrameos:
So close but still no cigar on this one.   Each fix has its own downsides and each is thwarted by something deep in the Java API.  I am guessing next attach will be to make a separate org.jpype.cl jar file with just the dynamic class loader.   The issue all stem from our caller sensitive grabbing the broken application loader.  Thus I need to fool it into thinking we are calling something loaded from the DynamicClassLoader level.  Other than splitting the jars not much chance making such flip.

astrelsky:
> I believe using a temporary directory is likely the most effective workaround.
> 
> **Linking vs. Copying:** If possible, we should try to hard link the libraries rather than copying them. This will require a fallback for people linking between disk drives, since you cannot hard-link across storage devices. Since you can symbolic link directories across drives we could try that but it might get messy. Symbolic linking or hard-linking will add close to zero startup time.
> 
> **Uniform Behavior**: Branching on `java_class_path.isascii()` creates two different behaviors. It would reduce the possibility of bugs to simply perform the workaround for all paths, regardless of whether they need it. This way, we can catch bugs in the workaround.
> 
> **Custom Loaders:** For people who need custom loaders or things on the class path at startup, we could expose a more flexible startJVM function:
> 
> ```
> startJVM(..., classpath: List[Path], system_classpath: List[Path])
> ```
> 
> * `classpath`: The post-startup Unicode-compatible approach to adding files to the classpath.
> * `system_classpath`: The approach that adds files to the JVM classpath options (`-Djava.class.path`). This should perform the temporary directory hack for all system class paths. This way, people can have custom class loaders. The support library can be added to this list.
> * A downside is that it would break backwards compatibility.

Unfortunately, using a temporary directory will cause it to always fail for anyone using windows with non ascii characters in their username.

Thrameos:
We can try clearing TEMPDIR, TEMP, and TMP, or just find the system temp directory rather than the user one on windows.

tachyonicClock:
According to https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir the temp directory python uses on windows are `C:\TEMP, C:\TMP, \TEMP, and \TMP`.

Thrameos:
I read that differently.

> Python searches a standard list of directories to find one which the calling user can create files in. The list is:
> 1. The directory named by the TMPDIR environment variable.
> 2. The directory named by the TEMP environment variable.
> 3. The directory named by the TMP environment variable.
> 4. A platform-specific location:
>    - On Windows, the directories C:\TEMP, C:\TMP, \TEMP, and \TMP, in that order.
>    - On all other platforms, the directories /tmp, /var/tmp, and /usr/tmp, in that order.
> 5. As a last resort, the current working directory.

Does that not mean that on windows systems it will choose "C:\Users\José\AppData\Local\Temp"  before choosing "C:\TEMP"?  That was it did on my system test.   

Looking at the Python code it seems like it always chooses a bad directory on windows systems.
```
def _candidate_tempdir_list():
    """Generate a list of candidate temporary directories which
    _get_default_tempdir will try."""

    dirlist = []

    # First, try the environment.
    for envname in 'TMPDIR', 'TEMP', 'TMP':
        dirname = _os.getenv(envname)
        if dirname: dirlist.append(dirname)

    # Failing that, try OS-specific locations.
    if _os.name == 'nt':
        dirlist.extend([ _os.path.expanduser(r'~\AppData\Local\Temp'),
                         _os.path.expandvars(r'%SYSTEMROOT%\Temp'),
                         r'c:\temp', r'c:\tmp', r'\temp', r'\tmp' ])
    else:
        dirlist.extend([ '/tmp', '/var/tmp', '/usr/tmp' ])

    # As a last resort, the current directory.
    try:
        dirlist.append(_os.getcwd())
    except (AttributeError, OSError):
        dirlist.append(_os.curdir)

    return dirlist
```

astrelsky:
> I read that differently.
> 
> > Python searches a standard list of directories to find one which the calling user can create files in. The list is:
> > 
> > 1. The directory named by the TMPDIR environment variable.
> > 2. The directory named by the TEMP environment variable.
> > 3. The directory named by the TMP environment variable.
> > 4. A platform-specific location:
> >    
> >    * On Windows, the directories C:\TEMP, C:\TMP, \TEMP, and \TMP, in that order.
> >    * On all other platforms, the directories /tmp, /var/tmp, and /usr/tmp, in that order.
> > 5. As a last resort, the current working directory.
> 
> Does that not mean that on windows systems it will choose "C:\Users\José\AppData\Local\Temp" before choosing "C:\TEMP"? That was it did on my system test.
> 
> Looking at the Python code it seems like it always chooses a bad directory on windows systems.
> 
> ```
> def _candidate_tempdir_list():
>     """Generate a list of candidate temporary directories which
>     _get_default_tempdir will try."""
> 
>     dirlist = []
> 
>     # First, try the environment.
>     for envname in 'TMPDIR', 'TEMP', 'TMP':
>         dirname = _os.getenv(envname)
>         if dirname: dirlist.append(dirname)
> 
>     # Failing that, try OS-specific locations.
>     if _os.name == 'nt':
>         dirlist.extend([ _os.path.expanduser(r'~\AppData\Local\Temp'),
>                          _os.path.expandvars(r'%SYSTEMROOT%\Temp'),
>                          r'c:\temp', r'c:\tmp', r'\temp', r'\tmp' ])
>     else:
>         dirlist.extend([ '/tmp', '/var/tmp', '/usr/tmp' ])
> 
>     # As a last resort, the current directory.
>     try:
>         dirlist.append(_os.getcwd())
>     except (AttributeError, OSError):
>         dirlist.append(_os.curdir)
> 
>     return dirlist
> ```

That does appear to be the case.


Will you put this down and come back to it tomorrow or Monday. There's no reason you need to do this today.

tachyonicClock:
Oh, I see I misinterpreted it. Good to know

Thrameos:
Alright I managed to make everything work with a non-ascii path.

Highlights of this PR:
1) Completely rework the DynamicClassLoader to upgrade it to be suitable for the System class loader.
2) Hijacked the multi-release jar system to include a second class file which is late loaded.  This late loaded file is guaranteed to be from our classloader and thus fixes all issues with permissions that we were getting form caller sensitivity classes.   The root of these problems was that Java has a set of hard coded instances such as jdbc connection in which they use the callers classloader to check against loaded service.  If the service was from any other classloader which was not the immediate parent then it would fail.  I thus needed to contrive a method for which our jar would be loaded by the application class loader, but the Reflector0 class would be loaded from our internal jar.  
3) With this manipulation, there is not NO differences between classes loaded at class load time or after the class load.   Our caller sensitive launch will deal with them as if they were identical.   Other callers in Java class land unfortunately will not, so our hack only fixes python calling.
4) In the case jpype itself installed under a non-ascii path, we make a copy in a temporary directory.  This ensures that we should never fail assuming the os allows us create the file. 
5) I removed the agent logic from the startup.  I can remove the entries from the manifest and jar if we aren't planning to use this again.  (Though it may be useful for tracking bugs)

Unfortunately this PR was rather messy.  It took me like 8 different tries to find the magical recipe the simultaneous satisfied all of the requirements for with and without an ascii path.   I really doubted that it was possible at all.  Outward interface should be the same, but I ended up with a large amount of gutting of the internals.

Thrameos:
Hmm.  Still issues on windows.   The problem appears to be an encoding problem.   

Here is the chars I get on windows:
0074 0065 0073 0074 002f 006a 0061 0072 002f 0075 006e 0069 0063 006f 0064 0065 005f 00c3 00a0 00f0 0178 02dc 017d 002f 0073 0065 0072 0076 0069 0063 0065 002e 006a 0061 0072

And here is chars I get on linux:
0074 0065 0073 0074 002f 006a 0061 0072 002f 0075 006e 0069 0063 006f 0064 0065 005f 00e0 d83d de0e 002f 0073 0065 0072 0076 0069 0063 0065 002e 006a 0061 0072

These appear when Java is pulling the values off the System variables.  

Both appear to be properly utf8 encoded when they leave C land for Java.  

Java doesn't have one bug with handling non-ascii paths, but TWO.   Wow, bonus time.

The logical solution would be to URL encode the string in Python land then decode it in Java to avoid the encoding bugs.   The means most likely we are going to have the same problem with modules and any variable that goes through CreateJVM.   Those in the international community how do you get by with this type of treatment from Java?

Thrameos:
Love when they have an error for ages but they won't fix it because it would break compatibility with being broken.

https://bugs.openjdk.org/browse/JDK-8195129

Thrameos:
I don't believe it is possible to fix the issue of the dll located on a non-ascii path on windows.   The problem stems from this code in Java.

```
// Loads .dll/.so and
// in case of error it checks if .dll/.so was built for the
// same architecture as Hotspot is running on
void * os::dll_load(const char *name, char *ebuf, int ebuflen) {
  log_info(os)("attempting shared library load of %s", name);

  void * result = LoadLibrary(name);  <== 
  if (result != NULL) {
    Events::log(NULL, "Loaded shared library %s", name);
    // Recalculate pdb search path if a DLL was loaded successfully.
...```

They are calling the ascii only version of LoadLibrary.  Apparently the correct solution is to convert to wide characters first than call LoadLibraryW.     Unfortunately, this is above my level to deal with.

The only work around I see:
1) Figure out given the current code page the appropriate set of bytes that will make LoadLibrary work.
2) Instead of doing load library, manually install each of our entry points using JNI_RegisterNatives.

The second is possible, but would require a special script just to maintain it.   The first I don't know how to start with because I am not sure how windows code pages even appear in Java.

**Comment by Thrameos** [comment]
@tachyonicClock I found the solution to ascii characters on the path, though it is horrible.   Apparently when you convert the path name to Short it will give you a path which works regardless.  I hacked that into the process.

The last part is intractable.   Whenever you run JPype on a Windows machine with a non-ascii path for the install directory, it is going to need to make a temporary copy of the jar.   BUT it is also going to hold that file handle open forever with no way of deleting it from within the process that I have found.   I put in a good 4 hours trying different work arounds.   I have no clue why Java keeps the file open on Windows and not on Linux.

We are technically working in that all of the functionality is present.  We need help from OpenJDK to implement better support for this problem.  I have done all that is feasible.   Though perhaps someone else as a creative idea.

**Comment by Thrameos** [comment]
The issue isn't the ones that we load.  The issue is that on the system class loader itself.  It is loaded by init level 3 and we will need to delete it at the end.  Because it is the second item loaded in the system (before we have that ability to call a register native or insert class) it occupies a special place.    I will look if perhaps there is a way to trash it on the shutdown sequence, but I suspect that it may be aone and done start given its location.

**Comment by Thrameos** [comment]
@marscher I am hoping to get 1.5.2 out in the next few weeks.   Will you have time to review or should I try to find others to assist?

**Comment by ryanmkurtz** [comment]
> @marscher I am hoping to get 1.5.2 out in the next few weeks. Will you have time to review or should I try to find others to assist?

I was getting ready to upgrade Ghidra to JPype 1.5.1 when the mismatching java versions problem popped up.  We plan on code freezing by the end of the month, so I'm debating if I should hold off for 1.5.2 and keep us at 1.5.0 which means we won't support Python 3.13, or upgrade to 1.5.1 knowing that 1.5.2 might not be out in time.  If i do the latter, i'd like to at least document the known limitations, like the JDK on the PATH needing to match the version that gets passed to JPype.  Maybe this choice will be made moot if 1.5.2 comes out relatively soon.

**Comment by Thrameos** [comment]
Assuming the code is ready I will be releasing this weekend or next.  We don't like leaving broken versions out for long.  

I do think documention of the weird platforms out there may still be necessary.  Given nice, complete,well installed versions are on the test system. And users often have broken, incomplete, mismatched arch versions.  We have no way to know that 1.5.2 won't also be a dud unless I give up on the non-ascii jar path.  Even then once I try to support modules, once again we may produce another dud.   Thus unless you are shipping all the pieces there will always be risk.  That doesn't mean we won't do our best to support the majority of installations, but we are only human.

A number of things are broken in openJDK and have been for 10 years or more.  Much of that came about because things were broken in windows and the Java developers threw up their hands.   Thus we are fighting on a very shaky foundation.

**Comment by ryanmkurtz** [comment]
Sounds good.  We also see every possible combination of JDK's and platforms, so Ghidra will likely be a good real-world testing ground.

**Comment by Thrameos** [comment]
@ryanmkurtz Do you have an opinion on dropping 1.8?   We have been holding out for a long time, but much of what I want to do going forward is going to be module based.  Thus I have a need to drop it.

**Comment by ryanmkurtz** [comment]
We have been switching to the latest LTS release about 6-12 months after it's released, so we are currently on JDK 21.  Dropping 1.8 def. won't affect us our our users.

**Comment by Thrameos** [comment]
@marscher Any chance you can start the reviews this week?

**Comment by Thrameos** [comment]
@astrelsky would you mind reviewing this code instead?  I can't leave a defective version out there and have to start the ball on a replacement.  But if I don't cut a new release soon I may end up in another busy period at work with little time to address issues.

**Comment by astrelsky** [comment]
> @astrelsky would you mind reviewing this code instead?  I can't leave a defective version out there and have to start the ball on a replacement.  But if I don't cut a new release soon I may end up in another busy period at work with little time to address issues.  

I can. I probably won't get to it until tomorrow night.

File: native/python/pyjp_module.cpp [hunk 2]
```diff
@@ -298,6 +314,18 @@ static PyObject* PyJPModule_shutdown(PyObject* obj, PyObject* pyargs, PyObject*
 		return nullptr;
 
 	JPContext_global->shutdownJVM(destroyJVM, freeJVM);
+
+#ifdef WIN32
+	// Thus far this doesn't work on WINDOWS.  The issue is a bug in the JVM
+	// is holding the file open and there is no apparent method to close it
+	// so that this can succeed
+	if (jarTmpPath != "")
+		remove(jarTmpPath.c_str());
```
astrelsky, line 323: You need to use [URLClassLoader.close](https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/net/URLClassLoader.html#close()) on the `SystemClassLoader`. Once closed it will be unable to load anymore classes which have not yet been loaded. Since we are shutting down here, it will be fine, you just need to call close before `shutdownJVM` (obviously). This won't have any effect on standard Java classes since they are in the builtin and platform class loaders.

setupext/build_ext.py
```diff
@@ -311,10 +311,16 @@ def build_java_ext(self, ext):
...
                                (javac, classpath, build_dir, target_version, target_version))
             cmd1.extend(ext.sources)
+            self.announce("  %s" % " ".join(cmd1), level=distutils.log.INFO)
+            subprocess.check_call(cmd1)
 
-            os.makedirs("build/classes", exist_ok=True)
+            cmd1 = shlex.split('%s -cp "%s" -d "%s" -g:none -source %s -target %s -encoding UTF-8' %
+                               (javac, build_dir, os.path.join(build_dir,"META-INF","versions","0"), target_version, target_version))
+            cmd1.extend(glob.glob(os.path.join("native","java0","**","*.java"), recursive=True))
+            os.makedirs(os.path.join(build_dir,"META-INF","versions","0"), exist_ok=True)
```
astrelsky, line 320: Inconsistent use of whitespace between function parameters.

Just pretend I'm an annoying bot 😛.

jpype/_core.py
```diff
@@ -252,41 +278,63 @@ def startJVM(
...
-    late_load = not has_classloader
-    if classpath:
-        cp = _classpath._SEP.join(_handleClassPath(classpath))
-        if cp.isascii():
-            # no problems
-            extra_jvm_args += ['-Djava.class.path=%s'%cp ]
-            jvmargs = _removeClassPath(jvmargs)
-            late_load = False
-        elif has_classloader:
+# Code for 1.6 release when we add module support
```
astrelsky, line 281: This is getting very complicated. I contemplated it a few times with the previous changes but it may be a good idea to split it up more when adding the module path handling later.

astrelsky (approved):
I'll see if I can break it tomorrow afternoon, as I'm burnt out for the night, but this looks good to me.

astrelsky:
I broke it where I did not expect to
image

astrelsky:
Ok good news I can't break it and #1242 is not necessary.

Thrameos:
Okay I will work on making changes today and see if we can get a release candidate ready.  I may need to ask others to test.  Our testing matrix really can't cover all the deployments our there.  Especially those partial jvms and international installs.

setupext/build_ext.py
Thrameos: My autopep8 hook appears to have broken.   It is supposed to reformat on each commit.

native/common/jp_context.cpp
```diff
@@ -159,6 +171,35 @@ void JPContext::attachJVM(JNIEnv* env)
...
+	HMODULE hm = NULL;
+	if (GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | 
+		GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+		(LPCSTR) &getShared, &hm) != 0 &&
+		GetModuleFileName(hm, path, sizeof(path)) != 0)
+	{
+		// This is needed when there is no-ascii characters in path
+		char shortPathBuffer[MAX_PATH];
+		GetShortPathName(path, shortPathBuffer, MAX_PATH);
+		return shortPathBuffer;
```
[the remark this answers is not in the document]
Thrameos: The issue here was that GetShortPathName ate the non-ascii characters which allows java to load the library properly.   This was the only pattern that I found to work consistently.   UNICODE doesn't help the problem as the issue is that Java doesn't accept non-ascii anywhere in the path.   It is better to let it mangle to "a~1/b~2/org.jpype.lib" which somehow allows it to function properly.

native/python/pyjp_module.cpp
Thrameos: I called "close" on the JPypeClassLoader which is the system classloader.   Unfortunately, that didn't change anything.

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\nelson85\\AppData\\Local\\Temp\\tmpfrr4g750'
```

The problem is at the ApplicationClassLoader which is not a URL classloader.

```
    private static class AppClassLoader extends BuiltinClassLoader {
        static {
            if (!ClassLoader.registerAsParallelCapable())
                throw new InternalError();
        }

        AppClassLoader(BuiltinClassLoader parent, URLClassPath ucp) {
            super("app", parent, ucp);
        }

        @Override
        protected Class<?> loadClass(String cn, boolean resolve)
            throws ClassNotFoundException
        {
            // for compatibility reasons, say where restricted package list has
            // been updated to list API packages in the unnamed module.
            @SuppressWarnings("removal")
            SecurityManager sm = System.getSecurityManager();
            if (sm != null) {
                int i = cn.lastIndexOf('.');
                if (i != -1) {
                    sm.checkPackageAccess(cn.substring(0, i));
                }
            }

            return super.loadClass(cn, resolve);
        }

        @Override
        protected PermissionCollection getPermissions(CodeSource cs) {
            PermissionCollection perms = super.getPermissions(cs);
            perms.add(new RuntimePermission("exitVM"));
            return perms;
        }

        /**
         * Called by the VM to support dynamic additions to the class path
         *
         * @see java.lang.instrument.Instrumentation#appendToSystemClassLoaderSearch
         */
        void appendToClassPathForInstrumentation(String path) {
            appendClassPath(path);
        }

        /**
         * Called by the VM to support define package for AppCDS
         */
        protected Package defineOrCheckPackage(String pn, Manifest man, URL url) {
            return super.defineOrCheckPackage(pn, man, url);
        }

        /**
         * Called by the VM, during -Xshare:dump
         */
        private void resetArchivedStates() {
            setClassPath(null);
        }
    }
```

I can't find a close method in any of the source code or documentation.   I would hope there would be a way to close or at least remove the lock.

jpype/_core.py
Thrameos: Do you see a way to clear this up or can we wait for the module path patch first?

native/common/jp_context.cpp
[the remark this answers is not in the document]
astrelsky: Ok. GetShortPathName is a macro though. If UNICODE is defined it will expand to GetShortPathNameW otherwise GetShortPathNameA.

I understand why this is being done. My point is that it can fail. As long as it fails in a way that we'll know what happened when the user reports a problem, then it's fine.

jpype/_core.py
astrelsky: I would just wait for the module patch.

native/python/pyjp_module.cpp
astrelsky: Ah for some reason I was thinking that this was for our system ClassLoader. Nevermind.

native/common/jp_context.cpp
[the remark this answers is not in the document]
Thrameos: All good.  I let it fall through to the old behavior.

PR #1249, opened 2024-12-02
Module

Thrameos:
This is a placeholder for module support.  This is built off the noagent PR,  so it will need to be rebased once that is included.  

Key points:
- Add a modulepath argument to startJVM to support modules.
- Add a modules argument to allow the user to add modules 
- Adds a module-info to org.jpype
- Starts the JVM with org.jpype on the module path and adds org.jpype to the list of initial modules.
- After this change the class loader and context are under org.jpype rather than unnamed module.
- The reflector used to launch from Python is currently under unnamed module.  I am not sure yet if this causes problems.

Needs:
- This needs to be tested extensively under non-ascii paths.  We are going to hit all the same problems that we did with classpaths with the broken application loader and broken System.load command.
- Need to test with  javaFX to make sure we now satisfy the requirements for modules.  
- If the reflector MUST be in a named module we will have to rework the late loader.   (required for JDBC and forName bugs)
- There are questions as to how to support other module arguments "opens", "patch" etc though I believe these are not as high usage as the"adds".
- Currently modules is connecting to "add-modules"  which may cause problems with documentation.   Though 

Review will defer until after the 1.5.1 mess is dealt with.


Thrameos:
Closing this out for now. This was a placeholder exploring module-path support (`--module-path`/`--add-modules`, moving `org.jpype` off the unnamed module), and the same ground is now covered - more completely - by the module-path migration on the `reverse` branch (#1274), which additionally addresses named-module native access (see #1310/#1468 for the interim classpath-based workaround that's shipping in the meantime).

Also, on a purely mechanical level, this branch predates the setup.py -> cmake/scikit-build-core build migration, so it now conflicts with the build system itself (`setupext/build_ext.py` deleted upstream, `module-info.java`'s location changed), not just with feature code - rebasing it standalone isn't worthwhile.

Reopen or extract pieces from this if `reverse` stalls and a smaller module-path-only slice is wanted independently.

PR #1259, opened 2025-01-30
Add signal handling for SIGTERM in addition to SIGINT

### LINKED ISSUE #1258: Handle SIGTERM in a similar manner as SIGINT (^C) when starting jpype with interrupt=False
## Proposal

We are currently using jpype to integrate some java libraries with our project. However, it is vital for us to regain control in python on SIGTERM as well as on SIGINT.

SIGINT handling is currently implemented when the appropriate flag is set (interrupt=False), however this does not work for SIGTERM, and for compatibility reasons, we would prefer both signals to be handed over.

This would improve the usage of jpype in long running, non-interactive processes which need proper shutdown logic, especially if it is hard to use the shutdown hooks for this logic, in our case because it contains a webserver, which requires to be shut down from the main thread.

## Changes necessary

We needed to move fast with our usage of jpype, so we forked jpype already and applied a patch which leads to SIGTERM being handled the same as SIGINT: https://github.com/jpype-project/jpype/compare/master...localstack:jpype:master#diff-49bead28f65cc4854943ff252ab201f7b866795548315fd0c21cfb24d6c672f0

I would be happy to file a PR, should this change be accepted.

dfangl:
## Motivation

This PR fixes #1258 by adding a SIGTERM handler in addition to a SIGINT handler to pass the control back to the python process, should the signal be cought.

## Changes
* Add parameter of the signal value to the `interruptPy` method, to be able to pass the right signal to python
* Install SIGTERM handler

## Discussion

We might want to have the sigterm handling behind a separate flag, as being discussed in #1258. Happy to do this, should it be preferred.

### DIFF
--- a/native/common/jp_context.cpp
+++ b/native/common/jp_context.cpp
@@ -492,7 +492,11 @@ extern "C" JNIEXPORT void JNICALL Java_org_jpype_JPypeSignal_interruptPy
 (JNIEnv *env, jclass cls, jint signal)
 {
 	interruptState = 1;
+#if PY_MINOR_VERSION<10
+	PyErr_SetInterrupt();
+#else
 	PyErr_SetInterruptEx((int) signal);
+#endif
 }
 
 extern "C" JNIEXPORT void JNICALL Java_org_jpype_JPypeSignal_acknowledgePy
--- a/native/common/jp_context.cpp
+++ b/native/common/jp_context.cpp
@@ -489,10 +489,10 @@ extern "C" JNIEXPORT void JNICALL Java_org_jpype_JPypeContext_onShutdown
 
 static int interruptState = 0;
 extern "C" JNIEXPORT void JNICALL Java_org_jpype_JPypeSignal_interruptPy
-(JNIEnv *env, jclass cls)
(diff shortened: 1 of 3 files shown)


Thrameos:
Looking at the PR I don't think a different flag is needed.  But will review further later.

Thrameos:
Looks like there is still a windows issue.   Once that is complete I can review and merge.

dfangl:
I skipped the tests on windows for now. SIGTERM is not a thing on Windows, and it seems like the SIGINT handling does not work either, the process just seems to crash.

"Signal" handling on Windows does not make a lot of sense anyway, but it might make sense to fix the CTRL-C behavior in windows at some point (assuming it actually does not work as the tests suggest, and not just the test setup is wrong).

I don't have a Windows setup right now, but should skipping not be acceptable, I will set one up.

dfangl:
@Thrameos Did you get a chance to look at it, or is there anything I can help?

Thrameos (approved):
IGTM

Thrameos:
Nope.  Just been in work overload mode.   Will include in next release.

PR #1264, opened 2025-03-20
Make JArray register as a sequence and List bug fixes

### LINKED ISSUE #1252: Java arrays are Iterable, but aren't Sequence
I have this test which does some simple checks on Java's arrays and arraylists & Python's arrays and tuples:

```py
from array import array
from typing import Iterable, Sequence

import jpype # from 'jpype1' package

# Function to check if an object is a Java array
def is_java_array(obj):
    return obj.getClass().isArray() if isinstance(obj, jpype.JObject) else False

# Start the JVM
jpype.startJVM()

# Import the Array classes
Array = jpype.JClass('java.lang.reflect.Array')
Arrays = jpype.JClass('java.util.Arrays')

# Create a Python tuple for testing
python_tuple = 10, 20, 30, 40, 50

# Create a Java ArrayList
java_list = Arrays.asList(python_tuple)

# Create a Java array
java_array = Array.newInstance(jpype.JInt, len(java_list))
for i in range(len(java_array)): java_array[i] = java_list[i].intValue()

# Create a Python array for testing
python_array = array('i', java_array)

# Test the function
print(java_array, len(java_array), is_java_array(java_array), type(java_array)) # True
print(java_list, len(java_list), is_java_array(java_list), type(java_list)) # False
print(python_tuple, len(python_tuple), is_java_array(python_tuple), type(python_tuple)) # False
print(python_array, len(python_array), is_java_array(python_array), type(python_array)) # False

print("\nJava's array is an Iterable:", isinstance(java_array, Iterable)) # True
print("Java's array is a Sequence:", isinstance(java_array, Sequence)) # False

print("\nJava's ArrayList is an Iterable:", isinstance(java_list, Iterable)) # True
print("Java's ArrayList is a Sequence:", isinstance(java_list, Sequence), '\n') # False

for int in java_array: print(int)

# Shutdown the JVM (optional)
jpype.shutdownJVM()
```
While an ArrayList is both an Iterable and a Sequence, a Java's array is just an Iterable, but not a Sequence!

Given a Java's array is also subscriptable and we can use Python's `[]` operator w/ positive integer indices, I believe a Java array should also be considered a Sequence in Python, just like an ArrayList.

Thrameos:
Fixes #1252

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -5,6 +5,11 @@ This changelog *only* contains changes from the *first* pypi release (0.5.4.3) o
 
 Latest Changes:
 - **1.6.0.dev0 - 2025-01-20**
+  
+  - JArray is now registered as a Sequence.
+
+  - Fixed bugs with java.util.List concat and repeat methods.
+
 - **1.5.2 - 2025-01-20**
 
   - Roll back agent change due to misbehaving JVM installs.
--- a/jpype/_jarray.py
+++ b/jpype/_jarray.py
@@ -17,6 +17,7 @@
 # *****************************************************************************
 import _jpype
 from . import _jcustomizer
+from collections.abc import Sequence
 
 
 __all__ = ['JArray']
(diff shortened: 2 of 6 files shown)


marscher (approved):
LGTM

Issue #1265, opened 2025-03-25
Potential bug in JVMFinder


Cansisti:
Hello,

I've found a potential bug in `JVMFinder.find_libjvm` when building this package.
It's using `os.walk` there to find `libjvm`, but most recent `openjdk` builds include two `libjvm`, one under `lib/server/libjvm.so` and another one `lib/lib/server/libjvm.so`.
`os.walk` returns paths in arbitrary order, meaning sometimes `find_libjvm` can choose the `lib/lib` one, which results in a change to `java.home` property, making it unable to find `java.security` later (only at `conf/...` in later versions).

Thrameos:
Care to submit a PR?

Cansisti:
Not sure what would be the correct way to fix this. Right now I'm using kind of brute force method to pick the one I want. Maybe adding `/lib/lib/server` and `/bin/bin/server` to `non_supported_jvm`? I'm not much of a java developer, so I can't say what the consequences would be.

Thrameos:
I believe this is more of a Python patch rather than a Java one.   I would guess we should just collect the list and then use some type of ordering function to make sure we always get the right one.

Ie.   lib/server/libjvm.so  vs lib/lib/server/libjvm.so  is 3 elements in path vs 4 elements in path.   Sort by number of elements and take first would always give us the one closest to the JAVA_HOME root.

Just patch the finder code in the jpype python module directry with the 3 lines of python required and verify that works should be all it takes.    I am not sure why they would have ever shipped it with lib/lib and bin/bin.   Seems like a very strange install.

--Karl

PR #1266, opened 2025-03-27
Ensure jvm closest to JAVA_HOME is always selected

Cansisti:
This fixes https://github.com/jpype-project/jpype/issues/1265


marscher:
Thanks for the PR. I changed the test which didn't make sense.

Issue #827, opened 2020-08-02
Test harness is badly named


Thrameos:
The jpype test harness is named "jpype" which means the main module on the Java harness are in conflict.  This one has been around for a very long time, but it will be more of a problem when I start allowing Python module code to be located in Java modules.  I recommend making the transition now by renaming the whole test package to "org.jpype" or possibly "org.jpype.test".

@marscher I know that you would prefer not renaming files to maintain the history, so I won't start this work until I get approval for the rename.

marscher:
please go ahead and rename it, if you think it clears things out.

PR #1268, opened 2025-04-05
Testing updates

marscher:
deals with #827


Thrameos:
I see you are working on one of those maintenance tasks.  Which release are you targeting this for?   We currently have three largish changes in the pipeline.   

- extend Java objects from Python
- A Java to Python bridge
- Modules

Not sure how long it will take to integrate it all but that does mean fairly substantial changes in the repo.  You may want to not hold a maintenance PR too long in queue or it risks conflicts with the others.

Additionally, in order to test the reverse bridge I will need to push it into a branch on the main project as the release pipeline can only be triggered there.  I will remove the branch once it is complete.  The merge will happen as a regular PR.

marscher:
We could also post pone this one in favor of the important changes. I can tidy up things later and do not want to interfere with the integration of the the main features.

PR #1274, opened 2025-04-10
Java to Python bridge

Thrameos:
This is a placeholder for now.

Current checklist:

- [X] Infrastructure
    - [X] Bridge starts from Java 
    - [X] Testbench framework from Java
    - [X] Testbench framework from Python
- [X] Core technology
    - [X] Module and Jigsaw work (in core upgrades)
    - [X] Speed limitations for Proxies (in core upgrades)
    - [X] Replaced exception handling paths for 3.12
    - [X] Automatic dispatch to Python types on return trip
    - [X] Tables/Maps to support 
    - [X] probe/pyobject prototypes
    - [X] Python conversion via class hints
    - [X] Split interfaces/pybase/python C++ from JPClass to optimize conversion paths
    - [X] Added bypass for default on known methods
- [x] Overall presentation in Java
   - [X] Basic types
   - [X] Interpreter
   - [X] Contexts
   - [X] Eval/Exec
   - [x] Callables
   - [x] Kwargs based dispatch 
- [ ] User extensions
   - [X] Service Model
   - [x] Working prototype
   - [ ] Extraction script
 - [ ] Finalization
   - [x] Cleanup
   - [ ] Packaging
   - [ ] Distribution
   - [x] Documentation
   - [ ] Beta test and user input
   - [X] Silly AI art
   
  API considerations:
  - [ ] To support network and subinstances I will need to review all core functions for ownership.   This means that in interfaces constructors will need to be removed.  Static methods in org.jpype should use instance instead to facilitate ownership policies.  Also the global context will need to be using the module one.  Likely should deal with this before proceeding to API finalization.   Estimated time 2 wks.


Thrameos:
With regards to distribution, It requires the Jpype distribution and C++ code to function so it would be bundled with JPype.  Though I expect that if one were to operate purely as a reverse bridge it would like want be formulated as a maven build in which it builds a portion of the support jar, then requires jpype installed in Python for the rest.  Unfortunately, I don't play much in that field so I would be happy if some had more insight.  Under the original plan from years ago I was going to build two jars "org.jpype" for the forward bridge, and "org.epypj" for the reverse bridge.   Both would be built from Python, but the reverse would have a separate packager so that one could install from maven.  For testing I am building the support pieces into our current jar, but that is not the final home.   


As for function as map, I would not want it to give map function freely.   If you want to get the dictionary like behavior then you would ask for the member to produce a wrapper for it.   That is the reason for all those additional interfaces like PyTuple, PyList, PyDict.   Those are wrappers what add to the behavior that is available.  If I were to add dictionary behaviors to the base Object it would conflict with those specializations.  I want all the Python standard collections to map the the same standard collections in Java.  

As for wrapping attributes we would have two choices:
1) An attributions() property that gives a Map like view with Java centered notations (add, remove, size, entrySet, keys, values)
2) Expose the same set of members that Python uses (getattr, hasattr, delattr, setattr)
3) Both??

I did the second as it was quick and dirty for this demonstration, but logically by my philosophy I prefer language bindings to best present in the language being used (except for naming conventions as that will break documentation lookup).  


As to what this PR is....   Currently it is a demonstration with the goal of finding enough interest to make it part of our project.   I did a very intricate design last time geared towards speed, but just didn't find sufficient users to justify the hours required to complete the test bench.   This is a lightweight design could be grown into something larger without the burden of requiring restructuring of the jpype core (which doomed the last effort).

Thanks for the feedback.  As this is an experiment in progress user feedback is required to come up with a design that fits the needs.   


Important design decisions thus far.   Previously, I was returning Java objects because one can't tell if a return object is going to be Python or Java though Java was a rare case.  This forced a huge amount of casting on the user when I was working on testing.     This time I return a PyObjectJava which is a wrapper to access the Java Object.     This is fairly critical as collections need to be of uniform type.   I believe usability favors this direction...

```
   PyObject obj;

   obj.getAttr("field").asFloat()  <== if everything is expected to be PyObject on returns then we don't cast here allowing chaining.

   for (Map.Entry<PyObject,PyObject> entry : obj.attribututes().entrySet())
   {
      // if the entry contains Java you will need to check and call get().
     if (!(entry.getValue() instanceof PyObjectJava))
        continue;  // skipping to the Java attributes

     MyClass c = (MyClass) entry.getValue().get();  // casting was going to be needed anyway so no real loss here
  }
```

astrelsky:
My question on the bundling and distribution is more towards just building a project that would use it. It would be weird to configure a Java project to look for a jar inside a python installation or virtual environment. You wouldn't bundle it with the project since jpype would load it.

Thrameos:
@astrelsky I checked in an example of how it would be packaged.   There would be a separate jar containing all the code required for the binding with the code required to span a Python interpreter scope.   Unfortunately it does not yet work because I haven't installed the magic compiler switched required to get Java to properly load the library meaning it fails on busted symbols.   (I remember it took me a while to find the compiler flags, but it should still be in epypj)

org.jpype.bridge.Bridge does most of the work.   If figures out where Python libraries are located and tests jpype to make sure it can be loaded.  It caches the values to ensure that the probe is only needed once.   Afterward it would directly start by lunching libpython.   There is one support hook required but that can be located in the main JPype dll so no need for a second native library.

One odd bug is a was not able to get it running from within netbeans.   It appears that there is an env variable that must be set for Python to locate packages from site-library.  I don't recall that issue the last time I worked on it.

Thrameos:
@astrelsky I successfully loaded the natives on the linux build.  It also tested fine on the windows build.   Thus all that is missing from it starting a Python session is the entry point for launching an interpreter.

Does this build system make sense to you?

When using the Java->Python bridge:
- There would be a maven project that distributes portion that Java must compile against available in maven-central.
- The compliable portion would mostly be a bunch of interfaces without any substance.
- There would be a single entry point class which finds the Python install locates the real jar file and launches the entry point from the native portion which was shipped in the JPype Python package.
- If JPype/Python is not found it would error out.
(If there were a bundled version it would ship Python and set the properties to go to the bundled version.)

When using the Python->Java bridge:
- JPype ships with both jars in the Python install directory.
- Python launches the org.jpype jar and entry points.
- During startup we load the reverse potion of the bridge from org.jpype.bridge so that the same behaviors used with reverse bridge are available in the forward bridge (if some function takes PyObject in Java we will convert to it).

You can follow the sequence in the Bridge class.

astrelsky:
> @astrelsky I successfully loaded the natives on the linux build. It also tested fine on the windows build. Thus all that is missing from it starting a Python session is the entry point for launching an interpreter.
> 
> Does this build system make sense to you?
> 
> When using the Java->Python bridge:
> 
> * There would be a maven project that distributes portion that Java must compile against available in maven-central.
> * The compliable portion would mostly be a bunch of interfaces without any substance.
> * There would be a single entry point class which finds the Python install locates the real jar file and launches the entry point from the native portion which was shipped in the JPype Python package.
> * If JPype/Python is not found it would error out.
>   (If there were a bundled version it would ship Python and set the properties to go to the bundled version.)
> 
> When using the Python->Java bridge:
> 
> * JPype ships with both jars in the Python install directory.
> * Python launches the org.jpype jar and entry points.
> * During startup we load the reverse potion of the bridge from org.jpype.bridge so that the same behaviors used with reverse bridge are available in the forward bridge (if some function takes PyObject in Java we will convert to it).
> 
> You can follow the sequence in the Bridge class.

Yes this makes sense to me. Thank you.

Thrameos:
@astrelsky  I put another session in.   Does this look like a workable direction for the API?

I broke the Python object into two sets: concrete and protocols.

The basic object is just a fan out to protocols.  I would try to make it castable with the Java cast operations, but there is no way to know if I have the right binding up front and without ASM support I can't freely mixin interfaces.  Thus the user would need to request the protocol that best suits their needs from the object.

We then have a bunch of predefined concrete types which will get both object and protocols as well as any of the special behaviors that define the type.

Last the protocols define behaviors and where possible map back to Java concepts such that Python objects can be handled by typical Java Collections type classes.

As always this is a monster amount of typing and testing to make it work.

astrelsky:
> @astrelsky I put another session in. Does this look like a workable direction for the API?
> 
> I broke the Python object into two sets: concrete and protocols.
> 
> The basic object is just a fan out to protocols. I would try to make it castable with the Java cast operations, but there is no way to know if I have the right binding up front and without ASM support I can't freely mixin interfaces. Thus the user would need to request the protocol that best suits their needs from the object.
> 
> We then have a bunch of predefined concrete types which will get both object and protocols as well as any of the special behaviors that define the type.
> 
> Last the protocols define behaviors and where possible map back to Java concepts such that Python objects can be handled by typical Java Collections type classes.
> 
> As always this is a monster amount of typing and testing to make it work.

I'll look through it over the week as I was already burnt out by the time I saw this today.

As a disclaimer, I hate just about every API until I actually use it.

I see the potential for it's use in Ghidra, which is where I would intend to use it as well, so maybe @ryanmkurtz could offer an opinion as well.


As for the ASM support, it is used in the extension classes anyways so maybe that will free that up? I'll open that pull request as a draft at for now just to get my ass moving on it again. It just needs to pull in commits, deal with conflicts, reformatted and then sit through review.

Thrameos:
I won't expect a lot of love when it comes for attempts to wrap an untyped language in a typed one.   I have torn up the classes three times already and I can't say it is satisfying in any form.  There is always a leakage between the arguments that can be taken, and what Python can wrap properly.  It is also exposing some very strange edge cases in JPype that I will be fighting to fix.  (Did you know that if a Java interface looked like and iterable the JProxy would assume the class methods were a list and tried to press it through a tuple?)  

I will be trying to avoid asm this time for a few reasons.  The main use I had for it was to create mixins so I could take a much of interfaces and mask them together so that the Java isinstance would actually be able to navigate them via casting.  That meant probing a class, finding all its dunder functions and then making a wrapper with the right exposure.  That was cool in that it looked neat, but it has some serious down sides.

1) We will never be able to use it on Android.
2) Objects often had conflicting interfaces which meant it was hard to get all the corner cases without some interfaces being badly compromised.

This time I am just going to accept that there may be many wrappers for the same Python object with different interfaces exposed for each.  Though until I program with it for a while I am not sure how comfortable it will be.   

As for automatically wrapping modules, if we aren't doing it on the fly we don't really need ASM.   We could just generate code and compile it.  So without the mixin there we could avoid it entirely.

Items that remain:
1) Exception handling.
2) Finish the collections behaviors.
3) Testing.

Thrameos:
Very ugly segfault was fixed.   I needed a handful of upgrades to the proxy system and it triggered a bunch of very nasty referencing errors.  I was forced to unroll the whole native changes and apply them one by one to locate the problem.  My secondary issue is that my build system on Windows produces corrupt dlls which means that everything must happen on WSL which is a much slower path.  dlls built from azure work, just my local visual studio install.

Once I get the native part working under windows (may have to trigger a release to get a working binary) I can start asking netbeans to sub some test cases.   That will be the first execise of paths.

astrelsky:
I just thought of this while looking through the code. I'm not sure if you already are or not but you might want to filter out access to the package from Python to prevent anything unexpected if you think that could cause a problem.

native/common/jp_bridge.cpp
```diff
@@ -0,0 +1,210 @@
...
+ * Signature: (Ljava/lang/String;)J
+ */
+JNIEXPORT jlong JNICALL Java_org_jpype_bridge_Native_getSymbol
+(JNIEnv *env, jobject, jstring str)
+{
+	try
+	{
+		jboolean copy;
+		const char* name = env->GetStringUTFChars(str, &copy);
+		for (std::list<void*>::iterator iter = libraries.begin();
```
astrelsky, line 149: I can't tell if this is leftovers that still needs to be refactored but I couldn't find the Java native method declaration.

Also, the access to the `libraries` list is not thread safe. I know msvc does thread safe static globals by default (which is really stupid to create a mutex for every global by default and silently add lock accesses) but I don't know what gcc or clang does. I would like to think they are sane and you would encounter a problem here.

Sorry for going on a minor tangent, I can't help myself sometimes.

Thrameos:
That was code left over from the epypj ap which require name lookup for the CFFI address lookup.   I copied it over just to make sure that the libraries are loaded properly as it provides a method to search for library resources.  The good news is the only code in that file which will be in the finished product is the “start” method which searches for the Python modules “jpype” and “_jpype” and then calls the initialize resources.  That bypasses the entire startJVM path.

I agree it would be a problem (it was marked to refactor to something safe where I copied it from) but as it is now dead code you can safely ignore it.

Thrameos:
@astrelsky It seems like there are some wrinkles with using maven to build the support Jar.   I can't get the order of the main jar to function properly.  The only way that it will work is if the org.jpype components are already in memory prior to start of the Python.  If not then org.jpype doesn't get the permissions it needs to operate.  

I don't think this is a show stopper.  I often run with a org.jpype that did not come installed with Python during testing.  I have even had it named differently than expected.  What it does mean is that rather than using maven I will likely need to build the package the usual way under Python and then use Apache Ivy to publish a bare artifact to maven central rather than an actual maven pom.   Yes that means there will be two copies of org.jpype on the users system.  One under the Python site and the other under the Java resources.   So long as one of them gets loaded on the class path it won't matter which copy we run from.

It does mean I will need to add some version guards to the process as one could imagine cases in which there was version sheer between the two copies.

(Oh well so much for maven.... back to using ant)

Thrameos:
@marscher Sorry about this turning into another monster PR.  I was hoping to just put all of the changes into one directory and leave the rest untouched, but it didn't work out that way.   The only way to get it functional is to place all of the contents in one tree and build a unified jar from it.   That requires moving every file in the native/java into a native/jpype_module/src/main/java.   Most of it is just simple motion, but even with git move it still makes for a huge amount of files changed.

I could split out the C++ changes into a separate PR if that helps.   The C++ changes were to fix three bugs I found in the type system and make it so that one can create JProxy with both inst and dict set which is useful for mapping a new interface over an existing object.

Thrameos:
@marscher would it make sense to try to split some of the PR into smaller more reviewable foundational chucks.   The extensions has a bunch of typo and spacing fixes, a conversion to C++17, and then the actual enhancement.    This one has a maven conversion, a module conversion, two bug fixes, a proxy enhancement, and then the actual bridge code.  I could try to break these up into a series of smaller PRs that you could review separately.  Do you think it is worth the effort to split them?

Thrameos:
@astrelsky wrote

> I just thought of this while looking through the code. I'm not sure if you already are or not but you might want to filter out access to the package from Python to prevent anything unexpected if you think that could cause a problem.

Just to be more specific.  The wrapped items are proxies and their appearance in Python depends on the value of the convert flag.   If they are marked as convert, they have no presence in Python land except if one were to explicitly cast to them.  Then the argument is stored as a field in Java and returned it automatically unwraps back to the Python object upon being returned.   (yes the flag is badly named as it should have been unwrap and it was undocumented such that ChatGPT make random guesses at what it does and even stated that it does the opposite of returning the Java portion rather than Python one)

Conversions take place at two locations.   When returning the argument from Java as part of a field or method request.  Or when used as the self argument for a proxy.   This is the location we have bugs in the self argument gives the inst portion of the proxy regardless of the convert flag (because it was added as a hidden self from a bound method), but the JProxy when passed as the second or third argument.   A Proxy derivative always gives itself.   A dict formed proxy won't give itself at all for methods and yields the JProxy object (not the Java portion either).   The difficulty is those classes have no presence or function without being cast specifically a type requested.

This causes errors if you implement a proxy and pass it through to another proxy that is expecting an interface.  Rather than getting the Java interface which one can act on instead you have a proxy item with no methods at all.   When dealing with a mixed item like a list that leads to very confusing behavior where you must cast each item off the list to the type you want to use.   

The behaviors are manageable but not very useful.  We should have had a more straight forward rule. 

- Objects returned form Java shall come back in the form they were taken.  Python to Python, Java to Java.
- The first argument for the proxy (self) must be the type that was known to Python.  
- All other arguments must be in their Java form for the interface that declared it.  

This would make for consistent behaviors rather than randomly returning different types depending on the convert flag.

astrelsky:
> @marscher would it make sense to try to split some of the PR into smaller more reviewable foundational chucks. The extensions has a bunch of typo and spacing fixes, a conversion to C++17, and then the actual enhancement. This one has a maven conversion, a module conversion, two bug fixes, a proxy enhancement, and then the actual bridge code. I could try to break these up into a series of smaller PRs that you could review separately. Do you think it is worth the effort to split them?

Sorry about the reformatting. I started to get annoyed and said to hell with it and ran the formatter I had setup. As with everything else in software development, my thought at the time was to keep going and it can be fixed later. (later is almost always never)

Thrameos:
Don’t worry about the formatting.  I was using netbeans to maintain all the spacing for a few years, but they dropped support during the transition period to Apache.  The revised beautifier still does a poor job when last I checked.   That meant that I have been doing everything with vim and haven’t had any automated beautifier.    Thus we have a large number of white space errors that have accumulated.

Thrameos:
Still working on getting a clean run.   I have been forced to use a very slow method of building binaries so I can advance one or two bugs a night.

Thrameos:
I now have basic functionality work.   I need to get interactive mode working so that I can do more extensive testing.

Missing items:
1) Testbench
2) Exception handling

I may take a second shot as the Proxy code to see if I can get a better implementation, but it should be ready for some eyes to look it over.

Thrameos:
@marscher @astrelsky This PR will be rebased onto the master once the foundational changes are accepted.  I will continue to work through the exceptions/test code as a draft for now.

Thrameos:
Having gotten most of the basic framework running I will try to plan for speedups and integration of exception handling.  I still need reviewers so that the API meets user needs and doesn't end up being needing major revision after release.

Thrameos:
I began work on the exception integration.   As always it is complicated because both exception systems have their quirks.  

It seems like I will need two additional foundation improvements before the exceptions are possible.
- [ ] Remove the concept of a local java context (was a misfeature before I realized that Java dropped support feature in 1.4)
- [ ] Add the probe capabilities to allow for easy wrapping of Python elements.

astrelsky:
> I began work on the exception integration.   As always it is complicated because both exception systems have their quirks.  
> 
> It seems like I will need two additional foundation improvements before the exceptions are possible.
> - [x] Remove the concept of a local java context (was a misfeature before I realized that Java dropped support feature in 1.4)
> - [ ] Add the probe capabilities to allow for easy wrapping of Python elements.
> 
> 

Haven't been ignoring this. I'll try and come up with a few examples to try out the API. I'll start with something simple like using the python json package from java and then move to other things using io and stuff.

I had a thought this morning. What about something for handling the python context manager protocol so it can be used as a try-with-resource statement. Since there isn't an equivalent to the `__enter__` it might need to be some sort of wrapper class. Would be ideal to remove the checked exception from the java close method though (I'm of the opinion that checked exceptions were a mistake. Even the Java runtime doesn't care, only the compiler does).

Thrameos:
Excellent.   I have been working on the try with resources by making certain wrappers can convert to AutoClosable.  That allows them to be registered as operating in a try with resource branch.   Once I have the probe API fully functional you should see that appear.

Here is the specific details.  Any object with ``__enter__`` is a PyResource.   You call the utility class to emit the enter.

```
   try (PyResource resource = with(ctxt("open('file.bin', 'wb')") )  
   {
     PyObject file = resource.as();
     ....
   }
```

Calling ``with`` on anything that can be used as a resource will create a wrapper called a resource.  Using ``as`` on the resource gets the value of the resource which is a file.  Still a work in progress, and yes the names need a lot of work, but that is what I hope it to look similar to.

The probe API should make it so that Java casting and instanceof will work.

Thrameos:
My current struggle is properly deciding whether a type is better exposed as a map or as a sequence on the Java side.   In Py 3.10 they added some bit flags to sort it out, but I want to target back to 3.8.    Also there are many types like numpy.array which just don't report to the abc types at all.  It will be hard to get accuracy without a detailed probe of usage or looking at the lint types.

astrelsky:
> My current struggle is properly deciding whether a type is better exposed as a map or as a sequence on the Java side.   In Py 3.10 they added some bit flags to sort it out, but I want to target back to 3.8.    Also there are many types like numpy.array which just don't report to the abc types at all.  It will be hard to get accuracy without a detailed probe of usage or looking at the lint types.

In what scenario for deciding if a type is better exposed as a map or a sequence? What would the sequence be? If there is any ambiguity then the user could have to decide which.

Thrameos:
As for how slicing will work....  here is an example.

```
public class Test {
  void call(PyIndex ... a)
  {}
  
  void main() {
    call(PyInt.of(1),PySlice.of(1,5));
  }
  
  interface PyIndex
  {}
  
  interface PySlice extends PyIndex  {
    static PySlice of(int i, int j) {
      return null;
    }
  }
  
  interface PyInt extends PyIndex {
    static PyInt of(int i) {
      return null;
    }
  }
}
```

Certain classes will inherit for PyIndex.   Those class automatically pack on the Java side into something requiring ``PyIndex...`` this gets them to the other side where they can be converted into a tuple and the request to ``__getitem__`` can be issued.    It is uglier than just placing an integer into the call list, but it will be usable.   Java doesn't give me a lot of meta language tools.

Hopefully that make sense.

Thrameos:
For convenience we could make some static methods to shorten this to ``call($i(1), $s(1,5))`` or ``call(i_(1), s_(1,5))`` 

Do you have a preference?

astrelsky:
> For convenience we could make some static methods to shorten this to ``call($i(1), $s(1,5))`` or ``call(i_(1), s_(1,5))`` 
> 
> Do you have a preference?

I'll avoid making any opinions on this until I've sat down and worked with it. At a glance and without context, they're both gross and I hate them :joy:

Thrameos:
As per your suggestion, if I don't get any indication I will make it a Map+  which will have overloaded get functions.  Those that are definitively Sequence such as tuple and list will get the correct wrapper.

Thrameos:
Seems like shortcut methods are not your thing.   Not to worry as nothing will force you to use them.

The indices interface is for the ``get`` method for map and list.   Nothing will force the user to use the shortcuts.   It is simply that Java doesn't have a good way to say stuff like    ``a[1,3:]`` so I have make it into something unpleasant like ``a.get(PyTuple.of(PyInt.of(1),PySlice.of(3, null)))``.   I can make it look gross but more compact as ``a.get($int(1), $slice(2,null))``

astrelsky:
For a slice what about just a two and three parameter `get` method?

Thrameos:
From the Python slot prospective there are only two slots
 
 - sq_item, sq_ass_item  correponding to ``__getitem__(int)`` and ``__setitem__(int, obj)``
 - mp_subscript, mp_ass_subscrip corresponding to ``__getitem__(obj)`` and ``__setitem__(obj, obj)``

In the language itself there are two extensions
- auto slice notation which converts ``a[1:5]`` to ``a[slice(1,5)]``
- auto tuple notation which converts ``a[1,2,3]`` to ``a[(1,2,3,)]``
I am not sure how many programmers actually know this detail of the language.  Thus some may think we don't support it.  


Thus the minimum functional interface is simply
- ``obj.get(int)`` and ``obj.set(int, obj)``
- ``obj.get(PyObject)`` and ``obj.set(PyObject, Object)``


If I am adding a convenience methods I can add

``obj.get(PyIndex...)`` which supports a type safe way for the numpy notation ``a[1,3:6,2,4]`` by knowing that when multiple indexing those with the Python ``__index__`` protocol are generally supported for sequence indexing with the addition of slice and ellipses.   This would help with the most common way to use indices.  There is a minor danger of conflict between ``get(PyObject)`` and ``get(PyIndex...)``
 
``indices(PyIndex...)`` which makes a type safe way of ``a[1,3:6]=1`` with ``a.set(indices($int(1), slice(3,6)), 1)``  

I think it would be bad form to try to add
 - ``a.get(1, 5)`` == ``a[1:5]``   how do you get to ``a[:5]`` and ``a[3:]`` notation?  Or was the user intending ``a[1,5]``?
 - ``a.get(1, 5, 2)`` == ``a[1:5:2]`` how do you get to ``a[::2] or a[3::2] or a[5::-1]`` Or was the user intending ``a[1,5,2] or a[1:5,2] or a[1,5:2]``

Those are all available if the user adds slice 
``a.get(slice(1,5)) = a[1:5]``
``a.get(slice(null, 5)) = a[:5]``
``a.get(slice(null,null,-1)) = a[::-1]``

If someone needs to get at something outside of the type restriction they would need to form the object themselves
``a[5,(4,6,9)]``  would be ``a.get(tuple($int(5), tuple($int(4), $int(6), $int(9))))``

The issue being no matter what solution we use it will end up ugly in some way.  Java just doesn't support the types of notation used in Python.

Thrameos:
Side note… None and null are the same in JPype so you can just use null in all the examples unless one absolutely needs to specify “None”.    “None” will appear in the static builtins along with “True”,”False”, “ellipse” and “stop” as they are all singletons.  They are a bit hard to get bootstapped as normal final statics, because they must appear before the interpreter is started.   I will likely just make the PyObjectDeferred in the declaration and have the start up routine fill in their values on start.  The downside of this is I will need an auto unboxing hack to make them appear correctly when passed to Python land.


While I understand the suggestion…. It is still ambiguous as to whether or not the user wants auto slice or auto tuple.

Java A.get(1,5)   == Python A[1:5] or A[1,5] ?

As someone who mainly does numpy with minimum of two dimensions I would find it very confusing to have auto slice rather than auto tuple when used for indexing.   After all : and , are very different symbols.   Swapping the meaning of , to be slicing would surely confuse people.

Is Java A.get(slice(1,5)) really that bad looking?


Note that slice operates identically in Python.
>>> A = [1,2,3,4,5]
>>> A[slice(1,None)]
(2, 3, 4, 5)
>>> A[slice(1,3)]
(2, 3)
>>> A[slice(None,3)]
(1, 2, 3)

If I were to have to have 2 and 3 argument version of get, it would have to be A.getSlice(1,5)  but if that was the case what is being saved?   Also note that we would be opening up a lot of ambiguity if we make “get(Object, Object, Object)” to support “A[None:None:None]”   so I would end up with something like “get(Integer, Integer, Integer)” meaning lots of unnecessary boxing.  Thus we have 5 methods to just add a slight bit of support for 1 dimensional slicing.   Doesn’t sound like the best tradeoff.

The simplest is just give the user only get(int) and get(PyObject) and make them be explicit.
Python A[1:5]     Java A.get(slice(1,5))
Python A[1,5]     Java A.get(tuple(1,5))
Python A[1:4]=range(2,5)     Java A.set(slice(1,4), range(2,5))

This does leave to problem that if you have to mix primitives and PyObjects that is not supported in Java.   Hence someone would need to create PyObject versions of each primitive, hence the need for $int and $float.

I know I am being very numpy centric in my thinking, but that is my primary use case.


I think I will need to table this until I get the probe and the exceptions API running.   Once I have those working it will be much easier to discuss code.

astrelsky:
That's a fair point I didn't know using a comma would create a tuple so I didn't consider the conflict.

Thrameos:
Started work on unittesting and documentation.   That is a very large amount of typing.   

I am still trying to decide if the protocols rightly belong in the python.lang package or if they should be in their own python.protocol package.   I moved PyBuiltIn to the python.lang as it looked better there.

Still up for work.   
- [ ] Finish the probe mixin interface.
- [ ] Plan how services will work so that we can implement extended wrappers for modules.
- [ ] Exceptions
- [ ] Resync the Python model with the updated Java code base.
- [ ] Get a module built on the jpype.release so that I can test everything.

Thrameos:
@astrelsky Looks like it won't be until next weekend that I finish with the probe interface which creates mixins.  But I did get a lot of the documentation and a portion of the test bench so that I can see what the code may actually look like.   

Most of the Python language features are in "PyBuiltIn" which is intended to use the the statement "import static python.lang.PyBuiltIn.*".   I have tried to keep it to Python names with moderate type safety.   The rest is just wrappers to make it easier to interact with Python objects though those generally have Java naming convensions.  I dropped the "as*" interfaces in favor of actual casting because the number of protocols had grown unwieldy.  That means PyObject is very minimal without same cast applied to it.  

The item that bothers me the most is what should the method name to retrieve the Python static type element from each class.  I had a conflict with getType() in base PyObject, though I have dropped the method and can rename them now.  Having two class systems will be a challenge.

The intent is that is we do have an actual Python module converted to Java we would start by scanning the type info and autogenerating a Java stub package which then registers itself as a service.  THat would add the new types to the autoconversion facility.  It will still be clucky to as most interfaces just give you a PyObject and then the user needs to know what to cast it to.  

I haven't looked into jep to see how much we would have to do to be on parity in terms of capabilities.  I assume that I have overshot in that I am implementing significant details of the Python language.

astrelsky:
I could have sworn I said that I was going to try and put an example together using json this weekend so I can try it out and provide feedback, but I can't find where I said this.

Anyway I'm aware that it isn't currently in a functional state. Let me know when it is and the commit hash and I'll set aside some time to put something together and provide feedback. I'll need the commit hash because I anticipate you'll keep working and break things, this way I don't need to go searching to find a working state.

Thrameos:
Still working on the probe api.  While I am making progress it takes a while to work through all the details required to pull this off.  I implemeted the backend to match the current interfaces, but still need to add C++ hooks to make tgr type conversion magic all work.

astrelsky:
> I could have sworn I said that I was going to try and put an example together using json this weekend so I can try it out and provide feedback, but I can't find where I said this.
> 
> Anyway I'm aware that it isn't currently in a functional state. Let me know when it is and the commit hash and I'll set aside some time to put something together and provide feedback. I'll need the commit hash because I anticipate you'll keep working and break things, this way I don't need to go searching to find a working state.

Just wanted to clarify that I'm in no rush. I actually didn't want to yesterday but decided to try and take a look anyway because I thought I said I would.

Thrameos:
Still making progress.  Python now launches successfully from Java and I was able to pass the majority of Python types into Java now that they have a named body.   But still miles of work before it is fully tested.  Everything has 20 edge cases.

Thrameos:
Finally got through the maze that was broken APIs and broken Python installs.

``PYTHONPATH="../.." java -cp target/test-classes/ --module-path=target/jpype-1.7.2.dev0.jar -Dpython.executable=python3.12 -Dpython.config.home=/usr --add-modules=org.jpype runner/HelloWorldMain``

```
package runner;

import java.nio.file.Path;
import java.nio.file.Paths;
import org.jpype.bridge.Context;
import org.jpype.bridge.Interpreter;
import python.lang.PyObject;

public class HelloWorldMain
{

  {
    try
    {
      Interpreter.getInstance().start(new String[0]);
      Context context = new Context();
      context.exec("msg = 'Hello World from Python'");
      PyObject result = context.eval("msg");
      System.out.println("Python returned: " + result);
    } catch (Throwable ex)
    {
      ex.printStackTrace();
      System.exit(1);
    }
  }
}

```

**Python returned: Hello World from Python**

Thrameos:
@marscher @ryanmkurtz  I am at the point were I actually have to consider deployment.  Currently I have a maven build that would distibute the Java module but then it still needs its other half which it unfortunately tied to python versions.  

I set up a probe and cache method which looks to find python and if it found runs python checking to see if jpype is installed and it's module location and version.  If everything matches it adds it to the cache so the next time it is launched it just skips straight to launch.  Then I exposed every possible python init setting.  

I don't have a good concept of how this is supposed to be deployed as don't think we want to be in the business of distributing python and I am not sure how maven handles a dependency like here is a jar which will work if you have something else installed.  I may as the ai to recommend a plan.

It was 4 hours of torture because my particular version of python had a bad copy of its library path coded in so it attempts to probe to get `encodings` when run as a library failed spectacularly.  Eventually fnding the magic flag `python.config.home=/usr` I may have to transfer that over from the probe as well but the entire python embedded launch process seems like an issue generator as they redid it for 3.11 and it has many rough edges that people like blender have to struggle with.    The number of things that can be wrong with a person's python install is enormous.

After packaging, I have test bench and then user defined extensions to demonstrate.   Exception routing still needs plumbing work.  Then I would need to get rough edges like starting jpype from the command line.  I can see the finish line.

Thrameos:
AI rendered version of what I am about to attempt.

image

marscher:
> I am at the point were I actually have to consider deployment. Currently I have a maven build that would distibute the Java module but then it still needs its other half which it unfortunately tied to python versions.

Can you please explain what is the relation of the Java support library with the Python extension? Currently the extension just requires an unconditional version of the support lib, right?
Is it now going going to be a bijective relation?

Thrameos:
The current system is we are started from Python and the support library is manually started from the JVM.  So the org.jpype is never distributed outside the python package system.     It just hangs out in the site-packages pretending to be a Python support data file.

Under new model where we start Python from Java the situation is different.  The support library is just a regular jar module you can load from maven.  When it starts it needs to find python.exe.  It probes python to find out all the venv settings and if set to install (typically as part of maven install procedure), it marries itself to the python install.    That allows you to then run from Java launching CPython as a script server.  

Both copies of the support library are identical.  They just exist on two different distribution systems.  PyPi and conda for Python starts, and maven for Java starts.  Thus dualistic not bijective.  Two different ways of starting the bridge (Java side or Python side) at least as far as distribution goes.

When it comes to runtime yes there is a huge change.  The key piece is JPype support library now offers a wrapper called `python.lang.PyObject` with the auto convert set on.  Any Python object is passed as a return or an argument to that particular type immediately creates a proxy and has presence on the Java side.  While we were strictly Java appears in Python before, we now have a two way street allowing a Java coder to interact.   If you want to run PyTorch or NumPy from Java that is entirely possible.  You don't even have to start from the Java side to use it.  Any java code that compiles against org.jpype can use Python freely.  Meaning someone can download the maven package, make a jar that depends on org.jpype and then launch from Python and use that code to mix and match freely.

Thrameos:
Miles of clean up remain.  But I got through almost the entire testbench when executed from Python.  That is excellent progress.   I will try to clean up the dead code and get the Java side of the house working in full in my next weeks session.

Thrameos:
@marscher Still plowing onward on this one.  Not sure how we are going to review a 58k line /500+ file PR, but I will work it to completion.   This will then get rebased on all of the outstand PRs and git pushed to branch ``alpha`` as one squash merge so that anyone interested in helping testing can see it.

Thrameos:
Superseded by #1475, which reconciles this branch with array-transfer-phase3 (#1470) onto `embedding`. Closing in favor of that PR.

PR #1277, opened 2025-04-18
Fixes and enhancements to JProxy

Thrameos:
This is part of the foundation for the Java to Python bridge.

This PR fixes bugs related to creation of proxies that inherited from Map in Java. It also allows for proxies to have both an inst and dict argument.   In the rare case that a JConversion is specified between to two Java classes that is also now handled properly

There is no changes outwardly in this PR other than the removal of the restriction.  In the back end we added an additional argument to the JProxy so that both types of argument are supported.

### DIFF
--- a/native/common/include/jp_proxy.h
+++ b/native/common/include/jp_proxy.h
@@ -39,10 +39,14 @@ public:
 		return m_Context;
 	}
 
-	virtual JPPyObject getCallable(const string& cname) = 0;
+	virtual JPPyObject getCallable(const string& cname, bool& addSelf) = 0;
 	static void releaseProxyPython(void* host);
 
-protected:
+	PyJPProxy* getInstance()
+	{
+		return m_Instance;
+	}
+
 	JPContext*    m_Context;
 	PyJPProxy*    m_Instance;
 	JPObjectRef   m_Proxy;
@@ -55,7 +59,7 @@ class JPProxyDirect : public JPProxy
 public:
 	JPProxyDirect(JPContext* context, PyJPProxy* inst, JPClassList& intf);
 	~JPProxyDirect() override;
-	JPPyObject getCallable(const string& cname) override;
+	JPPyObject getCallable(const string& cname, bool& addSelf) override;
 } ;
 
 class JPProxyIndirect : public JPProxy
@@ -63,7 +67,7 @@ class JPProxyIndirect : public JPProxy
 public:
 	JPProxyIndirect(JPContext* context, PyJPProxy* inst, JPClassList& intf);
 	~JPProxyIndirect() override;
-	JPPyObject getCallable(const string& cname) override;
+	JPPyObject getCallable(const string& cname, bool& addSelf) override;
 } ;
 
 class JPProxyFunctional : public JPProxy
@@ -71,7 +75,7 @@ class JPProxyFunctional : public JPProxy
 public:
 	JPProxyFunctional(JPContext* context, PyJPProxy* inst, JPClassList& intf);
 	~JPProxyFunctional() override;
-	JPPyObject getCallable(const string& cname) override;
+	JPPyObject getCallable(const string& cname, bool& addSelf) override;
 private:
 	JPFunctional *m_Functional;
 } ;
--- a/native/common/jp_classhints.cpp
+++ b/native/common/jp_classhints.cpp
@@ -85,7 +85,10 @@ JPMethodMatch::JPMethodMatch(JPJavaFrame &frame, JPPyObjectVector& args, bool ca
 }
 
 JPConversion::~JPConversion() = default;
-JPClassHints::JPClassHints() = default;
+JPClassHints::JPClassHints()
+{
+	convertJava = false;
+}
 
 JPClassHints::~JPClassHints()
 {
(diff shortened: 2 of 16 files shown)


native/common/include/jp_proxy.h
```diff
@@ -39,10 +39,14 @@ class JPProxy
 		return m_Context;
 	}
 
-	virtual JPPyObject getCallable(const string& cname) = 0;
+	virtual JPPyObject getCallable(const string& cname, bool& addSelf) = 0;
```
marscher, line 42: passing bool by reference seems uncommon, but works anyway. Feel free to change it or not.

native/common/jp_classhints.cpp
```diff
@@ -324,16 +329,6 @@ class JPHintsConversion : public JPConversion
 	JPMatch::Type matches(JPClass *cls, JPMatch &match) override
 	{
 		auto *pyhints = (PyJPClassHints*) cls->getHints();
-		// GCOVR_EXCL_START
```
marscher, line 327: the evil of code coverage exclusion hid the fixed bug?

native/common/jp_classhints.cpp
```diff
@@ -85,7 +85,10 @@ JPMethodMatch::JPMethodMatch(JPJavaFrame &frame, JPPyObjectVector& args, bool ca
 }
 
 JPConversion::~JPConversion() = default;
-JPClassHints::JPClassHints() = default;
+JPClassHints::JPClassHints()
+{
+	convertJava = false;
```
marscher, line 90: please keep the "m_" prefix for member variable code style. And maybe use initializer lists.

native/common/jp_classhints.cpp
```diff
@@ -642,6 +635,15 @@ class JPConversionObject : public JPConversion
 		JP_TRACE("assignable", assignable, oc->getCanonicalName(), cls->getCanonicalName());
 		match.type = (assignable ? JPMatch::_derived : JPMatch::_none);
 
+		// User has request a Java class to class conversion.  We must pass through check it.
+		if (!assignable)
```
marscher, line 639: would it be hard to create a Python regression test for this one?

native/common/jp_proxy.cpp
```diff
@@ -93,7 +101,7 @@ extern "C" JNIEXPORT jobject JNICALL Java_org_jpype_proxy_JPypeProxy_hostInvoke(
 
 			// convert the arguments into a python list
 			JP_TRACE("Convert arguments");
-			JPPyObject pyargs = getArgs(context, parameterTypePtrs, args);
+			JPPyObject pyargs = getArgs(context, parameterTypePtrs, args, proxy->m_Instance->m_Target, addSelf);
```
marscher, line 104: Is m_Instance always valid?

marscher: It is being initialized in the ctor, but never checked. This could cause a crash.

test/jpypetest/test_fault.py
```diff
@@ -342,13 +342,13 @@
...
-            _jpype._JProxy(None, [])
-        with self.assertRaises(TypeError):
-            _jpype._JProxy(None, [type])
-        _jpype.fault("JPProxy::JPProxy")
+#        with self.assertRaises(TypeError):
+#            _jpype._JProxy(None, None)
+#        with self.assertRaises(TypeError):
+#            _jpype._JProxy(None, [])
+#        with self.assertRaises(TypeError):
+#            _jpype._JProxy(None, [type])
```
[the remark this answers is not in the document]
marscher: Are postponing the tests, or do these cases simply do not make sense any more? I'd prefer to have some test cases as mentioned earlier.

marscher (approved):

marscher:
Thanks for fixing the bugs. How did you discover the missing increment of the Python ref count?

native/common/jp_classhints.cpp
Thrameos: The code that depends on this change does not appear in this branch so it appears untested.   I can try to replicate a similar situation artificially.  Though we don't currently have a way to delete a conversion once added, so I will need to find something that wasn't harmful.

native/common/jp_classhints.cpp
Thrameos: This code was moved to a new common location where it was unconditionally exercised.  I don't think there was a bug here other than hints only getting created occasionally and this code path or returning `_none` thus being the weird edge case that happen once in the whole execution.

native/common/jp_proxy.cpp
Thrameos: If m_Instance was null we would have big problems because that was part of the object that invoked it.  m_Target also can't be null or we would have already crashed Python on the clear method.

Locations it is called from  pyjp_proxy.cpp
```
     self->m_Proxy = new JPProxyDirect(context, self, interfaces);
```
jp_functional.cpp
```
        self->m_Proxy = new JPProxyFunctional(context, self, cl);
```

I could add a check, but I would never be able to exercise such a check and we would have crashed by construction given the value passing in is the reference to the Python object which is constructing it.

Thrameos:
I don't see a missing ref count in the PR.  I did change some of the reference counting code to make it more readable when I added the extra argument.   It was dec referencing the match type properly before, but a change in code paths made the code much more dangerous which tripped some segfaults during my development.   

I have addressed all the current review comments (except the fault check that I don't think I can test.)

PR #1281, opened 2025-04-19
Fixes float16 subnormal numbers

Thrameos:
Found the issue that was causing tests to occasionally fail.  The subnormal number conversion had a bug causing subnormal numbers to get extra bits set.


Thrameos:
Skipping the review on this one as it was a single function math error.

PR #1284, opened 2025-04-22
Context

Thrameos:
This PR is built on the Proxy upgrade PR.   It is to be reviewed after the Proxy is incorporated.

This PR corrects the issues with JPContext.  

JPContext was added in order to potentially support reloading the JVM.  Unfortunately, Java 1.4 removed the hooks necessary for that task but left the API.   Thus while we could support it if there was a conforming JVM, no such beast exists.   This left our code with lots of unnecessary handling what essentially was a singleton.  Because it appeared in many forms in the code often with different rules this disguised global structure is based often unnecessarily and is checked randomly for consistency.

This PR converts usage to two forms.   

- Those in which there is JavaFrame already open and we can just access the context in through JPJavaFrame::getContext().
- Those in which the Python code can execute without the JVM running or where we need to condition on the current state of teh JVM in which JPContext_global is best.

Unfortunately, there was a very large amount of clean up required, so this is a large PR.

This also resynced many of the long ago auto generated files which have been hand edited and thus drifted apart.  While they are still too different to be autogenerated, they are now closer than before.

### DIFF
--- a/native/common/include/jp_array.h
+++ b/native/common/include/jp_array.h
@@ -28,7 +28,6 @@ public:
 	~JPArrayView();
 	void reference();
 	bool unreference();
-	JPContext *getContext() const;
 public:
 	JPArray *m_Array;
 	void *m_Memory{};
--- a/native/common/include/jp_booleantype.h
+++ b/native/common/include/jp_booleantype.h
@@ -36,15 +36,11 @@ public:
 		return v.z;
 	}
 
-	JPClass* getBoxedClass(JPContext *context) const override
-	{
-		return context->_java_lang_Boolean;
-	}
-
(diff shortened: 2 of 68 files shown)


marscher:
Thanks for tidying up this mess. Should I take a look now or after you've merged #1277?

Thrameos:
As for the mess cleanup, it was one of those "I just can't take this"  moments where I was trying to push around some code and this managing kruft was taking more time than what I was trying to change.  I could keep pushing it around or burn a day to remove it entirely.   In the long run it will save time.

marscher:
I like this attitude! Thanks again! LGTM.

marscher (approved):

PR #1289, opened 2025-04-30
Userguide update

Thrameos:
Our userguide though remarkably complete for a project this size was getting out of date and did not cover some relevant topic areas.   With some assistance from an AI I went through the document.  Took some effort to ensure it didn't keep propagating myths like users should always call shutdownJVM and convertStrings is a very important option to set (please use the default.)  But otherwise I think the document revision came out well.   I was able to feed it some of the undocumented sections of our code and with a bit of editing I believe them to be better covered.

I have run this through the html generation step and encountered no errors, but that does not necessarily mean success on read-the-docs.  I have unpinned all versions to let it move to more modern tools, but we likely will want to repin them to working versions once this passes the tests.

### DIFF
--- a/doc/userguide.rst
+++ b/doc/userguide.rst
@@ -1256,9 +1256,9 @@ quality can improve to exact or derived depending on the cast type.
 Not every conversion is possible between Java types.  Types that cannot be
 converted are considerer to be conversion type "none".
 
+Details on how method overloads are resolved are given in `Method Resolution`_.
 Details on the standard conversions provided by JPype are given in the section
 `Type Matching`_.
-
 
 .. _cast:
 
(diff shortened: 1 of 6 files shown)


doc/userguide.rst
```diff
@@ -5,15 +5,195 @@ JPype User Guide
...
+
+Step 1: Start the JVM
+~~~~~~~~~~~~~~~~~~~~~
+JPype requires the JVM to be started before interacting with Java. Use the
+`jpype.startJVM()` function to start the JVM::
+
+    import jpype
+
+    # Start the JVM
+    jpype.startJVM(classpath=[])
```
marscher, line 111: If we show the user the classpath argument, it should at least be explained or referred to a more advanced use case.

doc/userguide.rst
```diff
@@ -5,15 +5,195 @@ JPype User Guide
...
+Step 3: Use Java Objects
+~~~~~~~~~~~~~~~~~~~~~~~~
+Create and manipulate Java objects just like Python objects::
+
+    java_string = String("Hello from Java!")
+    print(java_string.toUpperCase())  # Output: HELLO FROM JAVA!
+
+.. _introduction_step_4_shut_down_the_jvm:
+
+Step 4: Shut Down the JVM
```
marscher, line 133: Not to be nitpicky, but the capitalization is inconsistent, isn't it?

marscher:
Thanks for updating the guide! Just a quick question about the label hack. Could it be, that this was mandatory, because we've unpinned the sphinx dependencies? What was the behavior like without the add_labels script?
Anyways it seems to be super useful.

marscher (approved):

Thrameos:
The add labels script was my attempt to maintain labelling while the AI was working on the text.   It frequently lost labels, placed duplicate named sections, or malformed the header levels.  

For context, I am only given a very primitive version of ChatGPT 4o which is allowed to answer 30 questions and has a small context window.   I am also allowed only one attachment.   I attached the userguide and then tried to create a boiler plate first query that defined everything "header levels are blah blah, referencing uses this format, sections are organized this way, do not add speculative information, the definition of speculative is ..."   Even then it would frequently lose its mind before I could get to the 30 questions.   I have come to refer to it as Mr. Meeseeks.   

Because of those restraints I found it necessary to automatically maintain the labels in a stricter fashion than I would normally.   I was hoping that I wouldn't be forced to rename the duplicated labels, but alas I couldn't find a way to make the auto toc use the specified labels.   Although it didn't fix my current problem, I felt the script was still very useful for maintenance so I left it.

doc/userguide.rst
```diff
@@ -148,14 +158,14 @@ Save the following code in a file named `hello_jpype.py`:
     import jpype.imports
 
     # Start the JVM
-    jpype.startJVM(classpath=[])
+    jpype.startJVM(classpath=["../jar/*;../classes;com.amce-1.0.jar"])
```
marscher, line 161: nice example, thank you!

doc/userguide.rst
```diff
@@ -148,14 +158,14 @@ Save the following code in a file named `hello_jpype.py`:
...
 
     # Import Java classes
     from java.lang import String
 
     # Use the Java String class
     java_string = String("Hello from Java!")
-    print(java_string.toUpperCase())  # Output: HELLO FROM JAVA!
+    print(java_string.toUpperCase())  # Output: Hello from Java!
```
marscher, line 168: the output was correct, wasn't it?

marscher:
We could also write the docs im Markdown, which is more favored among  a lot of people. It can be done easily via [MyST](https://github.com/executablebooks/MyST-Parser). The existing docs can be converted using pandoc. 
But this is low-prio actually, isn't it?

doc/userguide.rst
```diff
@@ -158,7 +158,7 @@ Save the following code in a file named `hello_jpype.py`:
     import jpype.imports
 
     # Start the JVM
-    jpype.startJVM(classpath=["../jar/*;../classes;com.amce-1.0.jar"])
+    jpype.startJVM(classpath=["../jar/*","../classes","com.amce-1.0.jar"])
```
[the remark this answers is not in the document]
Thrameos: The reason we went to a list is because the separator differs from system to system.  Windows and cygwin are ";".  Linux and unix ":".   Thus users should always use the list formulation if they want to write portable code.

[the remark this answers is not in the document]
marscher: Right, I just noticed that I did not spot the list formulation because of the missing space after the coma. My bad. Thanks for the update.

Issue #1291, opened 2025-05-01
Support for readonly objects in jpype.nio.convertToDirectBuffer()


tnaroska:
`jpype.nio.convertToDirectBuffer()` allows wrapping byte-like python objects in a java direct buffer.
As of jpype 1.5.2, the call only allow wrapping writable objects (i.e. `bytearray`).

The ask is to generalize this and allow wrapping readonly `bytes` and `memoryview` as well. To keep the readonly contract, the ByteBuffer returned by  `convertToDirectBuffer()`  shall be a readonly instance. (see [ByteBuffer.asReadOnlyBuffer()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/ByteBuffer.html#asReadOnlyBuffer()))

With that support in place, this code snippet is expected to succeed:

```
import jpype
from jpype.nio import convertToDirectBuffer

# Launch the JVM
jpype.startJVM()


# wrap an RW object in direct buffer
rwbytes = bytearray(b"0123456789")
rwbuffer = convertToDirectBuffer(rwbytes)
assert rwbuffer.isReadOnly() is False


# following code currently raises ValueError but should return a readOnly direct buffer instead

# wrap an RO memoryview in direct buffer
roview = memoryview(rwbuffer).toreadonly()
robuffer = convertToDirectBuffer(roview)
assert robuffer.isReadOnly() is True

# wrap an RO object in direct buffer
robytes = b"0123456789"
robuffer = convertToDirectBuffer(robytes)
assert robuffer.isReadOnly() is True
```

Thrameos:
I will look into it.

tnaroska:
@Thrameos thanks!  I do have a draft implementation here https://github.com/jpype-project/jpype/pull/1293/files
It seems to work, but needs more vetting, updated unit-tests, documentation, ...

PR #1293, opened 2025-05-02
Support for readonly objects in jpype.nio.convertToDirectBuffer

tnaroska:
Fixes https://github.com/jpype-project/jpype/issues/1291

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -21,6 +21,9 @@ Latest Changes:
 
   - Fixed a bug in which JPype did not respect a JConversion between two Java classes.
 
+  - Enhancement in convertToDirectBuffer to support wrapping bytes and readonly
+    memoryviews as readonly java ByteBuffers.
+
 - **1.5.2 - 2025-01-20**
 
   - Roll back agent change due to misbehaving JVM installs.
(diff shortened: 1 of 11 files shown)


Thrameos:
It is unclear to me why we would need two copies of the convert to direct buffer.  It is a private function so you can just add an argument for "is read only"

tnaroska:
Hi @Thrameos, I did the requested change to restore the single `convertToDirectByteBuffer` function. Adapted and added unit-tests and ensured code-coverage passed.
Let me know if anything else is needed.

Thrameos:
I will give it a review.  Did you add a notice in the changelog and look to see if any changes would be required in thr userguide?

tnaroska:
Thanks for merging. Is there any timeline on when to expect a jpype release?

Thrameos:
I am currently trying to get a few major items which are on the order of 2-3 months.    

1) Cross Language Garbage Collection - Automatically detects cyclic references between the two languages and cleans them up.
2) Java to Python bridge - Allows Python to be treated as a Java native.
3) Extension support- Allows Python to create on the fly Java extension classes.
4) Module support - Adds module options to the startJVM.

The top three are big ticket items which will take a lot of review and discussion.     Though I could do an intermediate release of 1.6  if there is community interest.

tnaroska:
Understood. Those sound like bigger features indeed... In this case I'd definitely be interested in an intermediate release if possible.

Thrameos:
Alright I will try to get a intermediate release targeted for two weeks from now.

Issue #1301, opened 2025-05-15
Schedule release for 1.6


Thrameos:
@marscher i have a large number of items on the burner, but I also have requests for a maintenance release.  I will be starting the release scripts for 1.6 next Wednesday and pushing all outstanding tasks to 1.7.  If there is anything that seems like it should be included (i don't see anything in that category) then we should merge it.

marscher:
I agree with your plan. Please go ahead.

Thrameos:
Still running a bit behind.  Heavy work weeks.

Thrameos:
- [ ] Start from the release branch
      ``git checkout release``
- [ ] Make a new branch for the release cycle
      ``git checkout -b releases/{version}``
- [ ] Check the .azure scripts to see if they are up to date.
      Look on https://devguide.python.org/versions/ to see what versions can be dropped
      Check Python versions for Windows
      Check Python versions for OSX
      Check the manylinux image for Linux
- [ ] Merge the current master with the release
      ``git pull origin master``
- [ ] Start a release
      ``bumpversion release``
- [ ] Edit doc/CHANGELOG.rst
- [ ] Send the release to be evaluated
      ``git push``
- [ ] Verify CI on azure
- [ ] Manually trigger a ``jpype.release`` on azure
      If successful, download the artifacts for publication.
- [ ] Advance the release pointer
      ``git checkout release``
      ``git merge releases/<version>``
- [ ] Publish the release
  - Add draft release on github
  - Attach the artifacts to the release.
- [ ] Start master on a new cycle
  - Use a PR to pull release back to master
  - ``git checkout master``
  - ``git checkout -b cycle``
  - ``git merge release``
  - ``bumpversion patch``
  - Use PR to insert the cycle in master

Thrameos:
FIXME add directions for checking manylinux to release instructions
FIXME add a requirements file for release in .azure to make it easier to conduct a release.
FIXME move the doc/CHANGELOG checks earlier in process.
FIXME pointers to azure locations would help

petrushy:
Many thanks for this release! 

Just a heads-up as I don't see it on the list - pypi package needs an update as well

pelson:
@marscher & @Thrameos: Given the azure pipeline does the wheel building, is there any reason not to setup trusted publisher releases to PyPI, so that it ends up there automatically on git tag?

petrushy:
[...]

PR #1304, opened 2025-06-01
Cycle

Thrameos:
Release for 1.6.0 and start of the next cycle.

### DIFF
--- a/.azure/build.yml
+++ b/.azure/build.yml
@@ -79,19 +79,19 @@ jobs:
       #  jdk.version: '22'
       # Windows
       windows_py39_jdk11:
-        imageName: "windows-2019"
+        imageName: "windows-2022"
         python.version: '3.9'
         jdk.version: '11'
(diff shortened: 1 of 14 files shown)


Thrameos:
Fixes #1301

Thrameos:
Well that took a while.   I needed to some major surgery on the release scripts as we were way over time on azure cloud due to more Python versions manylinux release.  I am not sure that is going on with the 13t.  I suspect is it the variant without a global lock. They also put beta releases which is great to get some test time, but not something I want to wait around for because I don't plan to publish for the future.  I was forced to split build into smaller pieces.  I also added documentation and stubs for future builds.   

Despite my best efforts releases still eat nearly a full day of developer time.  I add more automation, but the build times keep growing and I have to cycle many times to get them correct.

marscher (approved):
sorry, that this review took so long. Thanks a lot for fixing and modernizing the build pipeline.

Issue #1307, opened 2025-07-07
JPype "unable to convert" 2d boolean NumPy array as of NumPy 2.3


john-hen:
Not sure if it's a problem with JPype or NumPy (could be the latter), but as of NumPy 2.3, the following script
```python
import jpype
from jpype import JArray, JBoolean
from numpy import array

jpype.startJVM()
bool_2d = JArray(JBoolean, dims=2)(array([[True, False], [False, True]]))
```

fails with
```
Traceback (most recent call last):
  File "convert_bool2d.py", line 6, in <module>
    bool_2d = JArray(JBoolean, dims=2)(array([[True, False], [False, True]]))
TypeError: Unable to convert
```

It does not fail with NumPy 2.2.6, but does with NumPy 2.3.0 and 2.3.1. (Regardless of the JPype version, 1.5.2 or 1.6.0.)

It still works for 1d boolean arrays (`dims=1`), and for `JInt` and `JDouble` arrays of any dimension (that I've tested).

The work-around is to call `.tolist()` on the boolean NumPy array object, then it works again. My test suite caught this, and yes, I could just use the `.tolist()` call going forward. Just wanted to bring it to your attention, maybe there's a reason for this. If it *is* a bug in NumPy, then I wouldn't know how to describe it.

Tested on Windows with Java 17 (via `winget install Azul.ZuluFX.17.JDK`).

Thrameos:
It is suddenly failed then most likely it is change in the buffer protocol implementation.   While it may be clear JPype actually has no integration with numpy whatsoever (we used to).  Instead we just request a buffer of a specific type and see if the type responds.   We made the switch circa 0.7.   There was a change in 1.6 to allow readonly buffers to be requested (or was it reverse I forget).    But either way we don’t even know what numpy is and depend entirely on the  buffer conversions.   Of course it was supposed to just report fail so we can try other methods.  If they fail on that the process would stop.

What this means is that you can use memoryview or other Python native method to trigger this error by requesting the buffer yourself.  (assuming that error came from numpy and not us).

So first confirm which party produced the error.  Then use memory view to trigger it in the new (and hopefully not the old).  Then you will have something you can meaningfully report.

It could be that we requested “byte” conversion and they now consider “bool” != “byte”  but that is only speculation.

--Karl

marscher:
This is the tracing output using Numpy 2.3.1

```
< PyJPModule_newArrayType
> PyJPArray_new
  > PyJPValue_alloc
[...]

PR #1320, opened 2025-09-30
[JPConversionAsBoolean] fixed numpy 2.3 breaking change of np.bool_  not being a Python bool anymore.

marscher:
Historically, NumPy had its own boolean scalar type, numpy.bool_, which is a subclass of numpy.generic but not of Python’s built-in bool. For a long time it was also an instance of PyBool_Type — so PyBool_Check(obj) returned true for both True/False and for numpy.bool_ scalars.

Fixes https://github.com/jpype-project/jpype/issues/1307

### DIFF
--- a/native/python/pyjp_module.cpp
+++ b/native/python/pyjp_module.cpp
@@ -24,6 +24,29 @@
 #include <Windows.h>
 #endif
 
+namespace {
+int init_numpy_bool_type()
+{
+	JP_TRACE("init_numpy_bool_type()");
+	PyObject *numpy = PyImport_ImportModule("numpy");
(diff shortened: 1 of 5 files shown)


native/python/pyjp_module.cpp
```diff
@@ -24,6 +24,30 @@
...
+		return -1;
+	}
+
+	/* store as PyTypeObject* for fast checks */
+	_NPBool_Type = (PyTypeObject *)t;
+	return 0;
+}
+
+}
+
```
Thrameos: Looks good that will make it a soft requirement.

PR #1322, opened 2025-10-08
Migrate scikit build

marscher:


marscher:
@Thrameos Nowadays it seems idiomatic to bundle all files belonging to a Python package within the package. Currently we have the extension and the jar lying around in site-packages along with the Python package. Should I tailor them to be included in the package, or are there strong objections against doing so?

PR #1330, opened 2025-11-19
Correct typo in _jcustomizer.py docstring

ryanmkurtz:
Fix typo in docstring regarding user specified conversions.


marscher:
thanks!

PR #1333, opened 2025-12-04
Fix initialization of mutex for cross module calls

### LINKED ISSUE #1316: import jpype fails with jpype1.6.0
When I import jpype from django console, it fails with error message "process finished with exit code -1073741819 (0xc0000005)". It is probably segmentation fault in windows.

I install jpype1.6.0 with pip, My OS is windows 10, jdk version is openjdk 11.0.2

marscher:
fixes #1316


Thrameos (approved):
LGTM

PR #1336, opened 2026-01-12
Fix PySlice_CheckFull on 32-bit architectures

### LINKED ISSUE #1029: Test failure on i586
Test failure on Python 3.9 that was encountered a while ago on openSUSE, only occurs on i586

```
 =================================== FAILURES ===================================
__________________________ JClassTestCase.testAsArray __________________________

self = <test_jclass.JClassTestCase testMethod=testAsArray>

    def testAsArray(self):
        jo = JClass('java.lang.Object')
>       self.assertEqual(jo[:], JArray(jo))
E       TypeError: Bad array specification on slice
```

cjwatson:
When a stepping-forward slice has no upper bound, `PySlice_Unpack` sets `stop` to `PY_SSIZE_T_MAX`, not `-1`.  They happen to be the same on some architectures, but that isn't portable.

Fixes: #1029

### DIFF
--- a/native/python/pyjp_class.cpp
+++ b/native/python/pyjp_class.cpp
@@ -726,11 +726,7 @@ static bool PySlice_CheckFull(PyObject *item)
 		return false;
 	Py_ssize_t start, stop, step;
 	int rc = PySlice_Unpack(item, &start, &stop, &step);
-#if defined(ANDROID)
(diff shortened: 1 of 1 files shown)


Thrameos (commented):
LGTM

PR #1348, opened 2026-03-29
Minimal version check

Thrameos:
Simplified version check.


Thrameos:
@marscher  Sorry for my long absence.  My eyesight is back up to working levels and I have projects require use of JPype again so my stopwork is finally cleared.  

 I would like to get the much delayed maintenance release out.    I looked over the other Java 9 release code but it looks like it is taking a very long path for what should be a simple test using the JNI version call.

PR #1350, opened 2026-03-31
Deadlock fix

Thrameos:
Fixes #1349


dev.mk
```diff
@@ -0,0 +1,31 @@
+# Variables
```
marscher: neat Makefile, I guess it is required for quicker build cycles. Do you mind moving it into the project dir, as it is not mandatory to use it for the sake of just installing/compiling the project?

marscher (approved):

PR #1352, opened 2026-04-05
Releases/1.7.0

Thrameos:
Review changes for release.


CMakeLists.txt
```diff
@@ -4,11 +4,13 @@ project(jpype-native)
 option(ENABLE_TRACING "Enable tracing in the extension" OFF)
 option(ENABLE_COVERAGE "Enable compilation for coverage measurement" OFF)
 option(ENABLE_BUILD_JAR "Enable building of the Java JAR" ON)
+option(BUILD_TEST_HARNESS "Build the Java test harness" ON)
 # Optionally set Python GIL disabled flag
 # set(PY_GIL_DISABLED ON)
 set(CMAKE_CXX_STANDARD 14)
 
-find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
+find_package(Python3 COMPONENTS Interpreter Development.Module REQUIRED)
```
marscher: is Dev.module more specific than just dev?

marscher (approved):

CMakeLists.txt
Thrameos: Yeah the problem was Development required the full .so while Dev module only required the portion available on the manylinux.  

The Python config says it was installed under /opt/_internal but that was missing, so CMake was just refusing.  Eventually gemini told me to give up and modify the CMake, but no being a CMake user I was just taking blind shots.

.azure/scripts/build-wheels.sh
```diff
@@ -1,43 +1,80 @@
...
+    auditwheel repair --plat "$PLAT" "$whl" -w /io/wheelhouse/
 done
 echo "=============="
 
-# Install packages and test
+# Simple verification
 for PYBIN in "${pys[@]}"; do
-    echo "Test install $PYBIN $package_name"
-    "${PYBIN}/python" -m pip install $package_name --no-index -f /io/wheelhouse
-    # Manylinux does not have a JVM so there is no way to test the wheel in the docker
```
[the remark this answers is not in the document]
marscher: The arm64 builds are taking up a lot of wall time. The other platforms are quite fast and running the tests there would not hit any boundary I guess.

Splitting up packages sounds like extra maintenance effort and I'm not sure what the benefits would be exactly.

.azure/scripts/build-wheels.sh
```diff
@@ -1,43 +1,80 @@
...
 
 for pybin in /opt/python/cp*/bin; do
-    # Exclude 3.6, 3.7, 3.8 and any alpha/beta/rc release
-    if [[ "$dir" =~ ^cp3(6|7|8|9|14)-cp3(6|7|8|9|14)t?$ ]]; then
+    # Get the directory name (e.g., cp310-cp310)
+    dir=$(basename $(dirname "$pybin"))
+
+    # EXCLUSION LOGIC:
+    # Drop EOL versions (3.6, 3.7, 3.8, 3.9)
+    # Note: 3.9 is EOL as of late 2025, but keep it if your users require it.
```
pelson, line 17: We support Python 3.8 & Python 3.9... https://github.com/jpype-project/jpype/blob/eeefa8dae9aecac6a1cd86e70a2e13cee72a86c6/pyproject.toml#L16

We should synchronise these two values. I think we should either drop Python 3.8 & 3.9, or provide wheels for them.

The problem with the current state is that `pip` sees that JPype supports Python 3.9, but since there is no wheel it builds from source. If instead we declared that Python 3.9 wasn't supported, then `pip` would fall back to the last version of JPype that did support it.

jpype/_core.py
```diff
@@ -49,18 +49,31 @@
...
+                current = getattr(_jedi_access, 'ALLOWED_GETITEM_TYPES', ())
+                setattr(_jedi_access, 'ALLOWED_GETITEM_TYPES', current + (_jpype._JMethod, _jpype._JField))
+            except (ModuleNotFoundError, AttributeError):
+                # If Jedi moves things again, we just move on.
+                pass
+    except (ModuleNotFoundError, ImportError):
+        # No Jedi installed, or something went wrong during early lookup
+        pass
+
+if 'jedi' in sys.modules:
```
pelson, line 75: This is different to what was done before. Does it actually work? Is it guaranteed that you need jedi in the sys modules to want to enable jpype completion? (not a user of Jedi, but flagging that this is a change in behaviour)

.azure/scripts/build-wheels.sh
ryanmkurtz: FWIW, a lot of PyGhidra users are still on Python 3.9.

marscher: Did your Python users already suffer from 1.7.0 being reported as Python 3.9 compatible?

ryanmkurtz: Because of #1316 we have Ghidra's version fixed to 1.5.2, so it hasn't affected anyone yet.

jpype/_core.py
Thrameos: Jedi was supposed to be optional.  So long as it was encountered first we would load support for it.      Does that not work?

.azure/scripts/build-wheels.sh
Thrameos: I can put it back in the build for 1.7.1 if needed.   This is one of those things where we updated the build system, but not the release procedures.    Do we only release for version that have not hit EOL?   or should be trail be a version?

jpype/_core.py
pelson: Then the behaviour should be what is implemented :+1: 
Previously we were importing Jedi unconditionally - just flagging it as a behaviour change. Personally, I'm not a user of Jedi.

.azure/scripts/build-wheels.sh
pelson: If easy to do, then let's do it. Internally, I have published a 3.8/3.9 manylinux wheel so that my users aren't affected, but if you release 1.7.1 without the compatible wheel and without #1359 then it would cause a bit more pain (with an easy fix on my side, to be fair).

jpype/_core.py
Thrameos: Sounds good.  I don't actually use it, nor even have it installed.  And because the way it was configured it broke on us entirely because jedi has not stable API for reporting user extension that I am aware of , it was causing problems again.   I reactivated the tests and decided the old behavior of using it so long as it was installed was likely a bug. 

I guess I should have labeled it "jedi was supposed to be optional and now is"

.azure/scripts/build-wheels.sh
Thrameos: SO 1.7.0 is broken because that segfault issue, so I was going to make a 1.7.1 to address all the issues.  I can release it with 3.8 and 3.9 in the build system so that we have a full "this release is the last good one for 3.8, 3.9.

Does that work?

marscher: @pelson How did you create this 3.8 wheel? I guess you did not use pip, because this complains about 3.8 not being supported any more by a lot of pkgs being dragged in as dependencies.
```
2026-04-21T13:17:04.8670961Z Failed to build jpype1
2026-04-21T13:17:04.9020267Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/e0/f0/8a2806114cd36e282823fd4d8e88e3b94dc943c2569c350d0c826a49db38/pip-25.1-py3-none-any.whl (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9022449Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/79/67/c06f625e2968c417052b3a4a0eef40656d5d4d44033e57b40ec474af1d28/pip-25.1.tar.gz (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9025816Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/29/a2/d40fb2460e883eca5199c62cfc2463fd261f760556ae6290f88488c362c0/pip-25.1.1-py3-none-any.whl (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9029858Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/59/de/241caa0ca606f2ec5fe0c1f4261b0465df78d786a38da693864a116c37f4/pip-25.1.1.tar.gz (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9033890Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/b7/3f/945ef7ab14dc4f9d7f40288d2df998d1837ee0888ec3659c813487572faa/pip-25.2-py3-none-any.whl (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9039049Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/20/16/650289cd3f43d5a2fadfd98c68bd1e1e7f2550a1a5326768cddfbcedb2c5/pip-25.2.tar.gz (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9041857Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/44/3c/d717024885424591d5376220b5e836c2d5293ce2011523c9de23ff7bf068/pip-25.3-py3-none-any.whl (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9045753Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/fe/6e/74a3f0179a4a73a53d66ce57fdb4de0080a8baa1de0063de206d6167acc2/pip-25.3.tar.gz (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9049632Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/69/00/5ac7aa77688ec4d34148b423d34dc0c9bc4febe0d872a9a1ad9860b2f6f1/pip-26.0-py3-none-any.whl (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9053187Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/44/c2/65686a7783a7c27a329706207147e82f23c41221ee9ae33128fc331670a0/pip-26.0.tar.gz (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9057125Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/de/f0/c81e05b613866b76d2d1066490adf1a3dbc4ee9d9c839961c3fc8a6997af/pip-26.0.1-py3-none-any.whl (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9060758Z   Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/48/83/0d7d4e9efe3344b8e2fe25d93be44f64b65364d3c8d7bc6dc90198d5422e/pip-26.0.1.tar.gz (from https://pypi.org/simple/pip/) (requires-python:>=3.9)
2026-04-21T13:17:04.9157089Z ERROR: Failed to build installable wheels for some pyproject.toml based projects (jpype1)
2026-04-21T13:17:04.9837502Z Traceback (most recent call last):
``

Comment (marscher) [reply .azure/scripts/build-wheels.sh:17]: Ah was looking in the wrong place and  Karl already fixed it. I've re-added Python 3.8 and 3.9 to the CI. It should be shipped for the next release.

Comment (pelson) [reply .azure/scripts/build-wheels.sh:17]: Unfortunately 1.7.1 didn't get manylinux wheels for 3.8/3.9 either.

I don't see the relevant release entry in https://github.com/jpype-project/jpype/pull/1361/changes for manylinux.

Comment (pelson) [reply .azure/scripts/build-wheels.sh:17]: FWIW, I built one with the following (AI generated) script:

```
#!/bin/bash
# Build a manylinux wheel for a specific Python tag.
# Usage: ./build_wheel.sh cp39-cp39
set -e -x

PYTHON_TAG=${1:?Usage: $0 <python-tag>  e.g. cp39-cp39}
PLAT=${PLAT:-manylinux_2_28_x86_64}
IMAGE="quay.io/pypa/manylinux_2_28_x86_64"
PYBIN="/opt/python/${PYTHON_TAG}/bin"

SDIST_URL="https://files.pythonhosted.org/packages/48/a2/5d27e81d24eef64668bf702bfe0e091cc48388b4666f36e025243eb9d827/jpype1-1.7.1.tar.gz"

mkdir -p wheelhouse
# wheelhouse/repaired is created inside the container (which runs as root).

# Build a cached local image with the JDK and ant pre-installed (done once).
CACHED_IMAGE="jpype-build-env:${PYTHON_TAG}-v2"
if ! docker image inspect "${CACHED_IMAGE}" &>/dev/null; then
    TMPDF=$(mktemp)
    cat > "${TMPDF}" <<EOF
FROM ${IMAGE}
RUN dnf install -y java-11-openjdk-devel ant && dnf clean all
RUN java11=\$(find /usr/lib/jvm -maxdepth 1 -type d -name 'java-11-openjdk-*x86_64' | sort | tail -1) && \\
    alternatives --set java    "\$java11/bin/java"  && \\
    (alternatives --set javac  "\$java11/bin/javac" 2>/dev/null || true)
ENV JAVA_HOME=/usr/lib/jvm/java-11
EOF
    docker build -t "${CACHED_IMAGE}" -f "${TMPDF}" .
    rm -f "${TMPDF}"
fi

# Build the wheel and repair it inside the cached container.
docker run --rm \
    -v "$(pwd):/io" \
    -e PLAT="${PLAT}" \
    "${CACHED_IMAGE}" \
    bash -esx -c "
        PYBIN='${PYBIN}'
        mkdir -p /io/wheelhouse/raw /io/wheelhouse/repaired
        export JAVA_HOME=\$(readlink -f /usr/lib/jvm/java-11)
        export PATH="\$JAVA_HOME/bin:\$PATH"
        export JAVACMD="\$JAVA_HOME/bin/java"
        \${PYBIN}/pip install --upgrade pip
        \${PYBIN}/pip wheel '${SDIST_URL}' -w /io/wheelhouse/raw/ -v
        for whl in /io/wheelhouse/raw/jpype1-*.whl; do
            auditwheel repair --plat ${PLAT} \"\${whl}\" -w /io/wheelhouse/repaired/
        done
    "

echo "Repaired wheels in wheelhouse/repaired/:"
ls wheelhouse/repaired/
```

marscher: right, the manylinux image we're using dropped support for these Python versions and would be needed to manually added to the container.

Please add an issue for this and stop replying to this one.

PR #1354, opened 2026-04-07
Support for hotpatch of static fields

### LINKED ISSUE #1328: Accessing Java variables with illegal Python characters
Hopefully this is a question with an answer instead of an issue.

If I am trying to access the Java variable `MyClass.foo_$bar` from JPype, how to I do that?  Including the `$` results in `SyntaxError: invalid syntax`.

Thrameos:
This is a specific edge case.  It is an edge case where you want to be able to get to an aliased copy of a member.   To make it nicer deal with the descriptors I have added a "." leader which prevent descriptors from being dereference. 

Fixes #1328


Thrameos:
I will need to refresh this one.  Unfortunately it exposed a bug that it is impossible to access members that start with underscore from java because we unconditionally gave that space to Python.  I also found it very difficult to perform renaming tasks.  This PR will fix both those issues.

PR #1356, opened 2026-04-12
[cmake] fixing jni fallback + jar cp on windows

marscher:
During the build on conda-forge (https://github.com/conda-forge/jpype1-feedstock/pull/60) I encountered some issues, which are fixed by a patch in the feedstock. Now I want to integrate them back here.

1. If the JNI header cannot be found (due to some exotic paths inside the conda-forge build environment), we use the provided fallback header in this repo. This is exactly the behavior we used to have prior the transition from setuptools.
2. On Windows utilizing MSBuild there was a race condition in the jar copy section.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,26 +12,46 @@ set(CMAKE_CXX_STANDARD 14)
 find_package(Python3 COMPONENTS Interpreter Development.Module REQUIRED)
 find_package(Python3 COMPONENTS Development.Embed)
 find_package(Java REQUIRED COMPONENTS Development)
-# todo: add fallback jni.h in case it aint found (no jdk required, or not found).
-#set(JAVA_INCLUDE_PATH "${CMAKE_SOURCE_DIR}/native/jni_include" ${JAVA_INCLUDE_PATH})
-find_package(JNI REQUIRED)
+find_package(JNI)
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -12,26 +12,46 @@ set(CMAKE_CXX_STANDARD 14)
...
+    message(STATUS "System JNI not found. Falling back to bundled jni.h.")
+
+    # 2. Populate the standard variable that FindJNI usually sets
+    set(JNI_INCLUDE_DIRS "${CMAKE_SOURCE_DIR}/native/jni_include")
+
+    # 3. Create a mock JNI::JNI target
+    # This ensures that if your CMakeLists relies on modern target-based
+    # linking (e.g., target_link_libraries(my_ext PRIVATE JNI::JNI)),
+    # it won't break.
+    if(NOT TARGET JNI::JNI)
```
marscher: actually we do not link against JNI, right? So this can be removed, can it not?

Thrameos:
Thanks for the hard work on this one.  I saw reruns on the conda system running all weekend.  Hopefully not too bad.

marscher:
Thanks! The Windows problem was a tough one. The msbuild isn't covered (yet) by our pipeline. I intend to add the conda-recipe here to catch such things earlier.

PR #1359, opened 2026-04-17
[pyproject.toml] updated python requirements.

marscher:
also updated author, maintainer fields.

@Thrameos would you mind adding a public email address for the Python Package Index?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ maintainers = [
 description = "A Python to Java bridge"
 readme = "README.rst"
 requires-python = ">=3.10"
-license = {text = "Apache-2.0"}
+license = "Apache-2.0"
 classifiers = [
     'Programming Language :: Python :: 3',
     'Programming Language :: Python :: 3.10',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ maintainers = [
 ]
 description = "A Python to Java bridge"
 readme = "README.rst"
-requires-python = ">=3.10"
+requires-python = ">=3.8"
 license = "Apache-2.0"
 classifiers = [
     'Programming Language :: Python :: 3',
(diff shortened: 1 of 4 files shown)


Thrameos:
Sure will do.

marscher:
Thanks for adding your mail address.

In some comments of yours @Thrameos I've noticed that we would be able to still support python 3.8. I wonder if that's really the case. We've dropped it from ci for a reason other than these version are EOL. At least we should re-add them to the test matrix, if we want to release it as supported, right?

pyproject.toml
```diff
@@ -7,22 +7,22 @@ name = "jpype1"
...
 maintainers = [
-    {name = "Luis Nell", email = "cooperate@originell.org"},
+    {name = "Martin K. Scherer", email = "martin.k.scherer@gmail.com"},
 ]
 description = "A Python to Java bridge"
 readme = "README.rst"
-requires-python = ">=3.8"
-license = "Apache-2.0"
+requires-python = ">=3.10"
+license = {text = "Apache-2.0"}
```
marscher: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

The table spec got deprecated.

marscher:
The tests still pass on Py38, I want to include a CI variant using conda-build, so we can easily test these old versions without the install hassle on Azure.

pyproject.toml
Thrameos: Got it.  Revert the line.  The main plan here with 3.8 is to make sure we have a clean build then update the requires to deal with EOL.   Technically we could support them until they are broken, but like our Java 1.8 eventually it will start causing friction (as well as needing to upload insane number of files)

marscher:
conda-build with conda-forge is not an option to test Python 3.8. I think we would need to use the manylinux Docker images with the old versions.

PR #1360, opened 2026-04-18
Fix numpy null

### LINKED ISSUE #1358: SIGSEGV in JPBooleanType::findJavaConversion when numpy is not installed (regression in 1.7.0)
## Summary

JPype 1.7.0 crashes with a SIGSEGV when calling a Java method that has both `boolean` and `String` overloads, passing a Python `str` argument. This worked correctly in JPype 1.6.0.

## Environment

- JPype: 1.7.0
- Python: 3.12
- JDK: OpenJDK 21.0.9
- OS: Ubuntu 25.04 (WSL2)
- numpy: **not installed**

## Reproduction

```python
import jpype
import jpype.imports

jpype.startJVM(classpath=["your.jar"])

# org.openqa.selenium.MutableCapabilities.setCapability has these overloads:
#   setCapability(String, boolean)
#   setCapability(String, String)
#   setCapability(String, Platform)
#   setCapability(String, Object)

from org.openqa.selenium.remote import DesiredCapabilities
caps = DesiredCapabilities()
caps.setCapability("appium:newCommandTimeout", "3600")  # <-- SIGSEGV here
```

## Crash output

```
#  SIGSEGV (0xb) at pc=0x...
# Problematic frame:
# C  [python3+0x16042ef]
```

Stack slots in the JVM crash log:
```
stack at sp + 1: _ZN11JPJavaFrameD1Ev    (_jpype.so)
stack at sp + 7: _ZN13JPBooleanType18findJavaConversionER7JPMatch    (_jpype.so)
si_addr: 0x0000000000000008
```

## Root Cause

PR #1320 added numpy 2.3 compatibility by checking `_NPBool_Type` in `JPConversionAsBoolean::matches()` inside `native/common/jp_booleantype.cpp`:

**Before (1.6.0):**
```cpp
JPMatch::Type matches(JPClass *cls, JPMatch &match) override
{
    if (!PyBool_Check(match.object))
        return match.type = JPMatch::_none;   // str exits here safely
    match.conversion = this;
    return match.type = JPMatch::_exact;
}
```

**After (1.7.0, introduced by PR #1320):**
```cpp
JPMatch::Type matches(JPClass *cls, JPMatch &match) override
{
    PyObject* obj = match.object;
    if (PyBool_Check(obj) || PyObject_IsInstance(obj, (PyObject*)_NPBool_Type))
    //                                                   ^^^^^^^^^^^^^^^^^^^
    //                                                   nullptr when numpy is not installed!
    {
        match.conversion = this;
        return match.type = JPMatch::_exact;
    }
    return match.type = JPMatch::_none;
}
```

When numpy is not installed, `init_numpy_bool_type()` returns -1 and `_NPBool_Type` remains `nullptr`. When overload resolution encounters a method with a `boolean` overload and a Python `str` is passed:

1. `PyBool_Check(str_obj)` → `false` (no short-circuit)
2. `PyObject_IsInstance(str_obj, (PyObject*)nullptr)` is called
3. CPython reads `nullptr->ob_type` → dereferences address `0x0 + 8 = 0x8`
4. **SIGSEGV at `si_addr: 0x0000000000000008`** — matches the crash log exactly

## Suggested Fix

Add a null guard in `native/common/jp_booleantype.cpp`:

```cpp
if (PyBool_Check(obj) ||
    (_NPBool_Type != nullptr && PyObject_IsInstance(obj, (PyObject*)_NPBool_Type)))
```

## Workaround

Pin JPype to 1.6.0:

```
JPype1==1.6.0
```

Thrameos:
The numpy bool fix #1320 introduced a regression where systems without NumPy would SIGSEGV. This occurred because PyObject_IsInstance was called with a nullptr for the type, as the CPython API lacks internal null guards.

Key Changes:

- Safety & Speed: Replaced the PyObject_IsInstance calls in native/common/jp_booleantype.cpp with our internal PyJP_IsInstanceSingle. This provides a fast-path check and inherent safety against nullptr handles.   It also ensures that if for some reason someone inherited from numpy.bool it would still work.

- Refined Logic: Updated the conversion script portion to respect derived types. Replaced janky strcmp logic with proper type-hierarchy awareness.  This also allows for derived types to work successfully.  This fixes the rare case that numpy is renamed.

- CI Reinforcement: Added a linux_py312_jdk17_no_numpy job to the Azure Pipeline matrix. This ensures that JPype’s "vanilla" Python support remains stable and is never again regressed by optional dependencies.

Fixes #1358

### DIFF
--- a/.azure/build.yml
+++ b/.azure/build.yml
@@ -80,6 +80,11 @@ jobs:
         imageName: "ubuntu-latest"
         python.version: '3.12'
         jdk.version: '17'
+      linux_py312_jdk17_no_numpy:
+        imageName: "ubuntu-latest"
+        python.version: '3.12'
+        jdk.version: '17'
+        skip_numpy: 'true'
       linux_py313_jdk17:
         imageName: "ubuntu-latest"
         python.version: "3.13"
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -38,6 +38,7 @@ steps:
       pip install -r test-requirements.txt
     fi
   displayName: 'Install test requirements'
+  script: bash
 
 - script: |
     ant -f test
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -29,7 +29,7 @@ steps:
 #    python -c "import jpype"
 #  displayName: 'Build/install/check module (editable)'
 
-- script: |
+- bash: |
     # Create a temporary requirements file without numpy if skip_numpy is true
     if [ "$(skip_numpy)" == "true" ]; then
       grep -v "numpy" test-requirements.txt > test-requirements-no-numpy.txt
@@ -38,7 +38,6 @@ steps:
       pip install -r test-requirements.txt
     fi
   displayName: 'Install test requirements'
-  script: bash
 
 - script: |
     ant -f test
--- a/.azure/scripts/test.yml
+++ b/.azure/scripts/test.yml
@@ -30,7 +30,13 @@ steps:
 #  displayName: 'Build/install/check module (editable)'
 
 - script: |
-    pip install -r test-requirements.txt
+    # Create a temporary requirements file without numpy if skip_numpy is true
+    if [ "$(skip_numpy)" == "true" ]; then
+      grep -v "numpy" test-requirements.txt > test-requirements-no-numpy.txt
+      pip install -r test-requirements-no-numpy.txt
+    else
+      pip install -r test-requirements.txt
+    fi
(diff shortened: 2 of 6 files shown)


Thrameos:
@marscher I think we will have to make a 1.7.1 release to deal with the segfault issue.   There apparently is also packaging issues.  I think we should do a patch release this weekend and then perhaps dump the broken version?

marscher:
I got sloppy when trying to address the numpy issue. Sorry for that. I agree we should release shortly after merging this and reconsidering the python3.8/9 handling.

Thanks for this proper fix!

PR #1364, opened 2026-05-07
Bump version: 1.7.1.dev0 → 1.7.2.dev0

marscher:


marscher:
@Thrameos  I've observed these crashes on Windows in CI for quite some time for no obvious reason. Apparently we do not have a debug enabled build in CI, which we should. So unluckily we lack information about the crash. 

```
jpypetest\test_jclass.py::JClassTestCase::testStaticMethod PASSED        [ 42%]
jpypetest\test_jclass.py::JClassTestCase::testUnsatisfied PASSED         [ 42%]
jpypetest\test_jdouble.py::JDoubleTestCase::testArrayBadDims PASSED      [ 42%]
##[error]Cmd.exe exited with code '-1073741819'.
```

When I randomize the test execution order with pytest-randomly, I can reproduce crashes. Need to investigate further.

marscher:
This fails randomly and even provokes the crash later on:

pip install pytest-randomly

then execute a couple of times
pytest test/jpypetest/test_overloads.py


FAILED test/jpypetest/test_overloads.py::OverloadTestCase::testFunctionalInterfacesWithDefaults - AssertionError: TypeError not raised by testFunctionalInterfaces

Thrameos:
Okay I will see if I can get the failure to appear.     The only weakness that we have is with that memory model after the 3.11 changes but I don't see why it would be os specific.   Should we cut a debug version in the release version?   I am not quite sure how that would work though as I don't know how one would switch the backend that was loaded.

marscher:
see also #1365

Thrameos:
@marscher I think this should get merged.  I have fixed #1365 so no need to hold off.

PR #1366, opened 2026-05-09
Method dispatch correction

### LINKED ISSUE #1365: test_overloads:testFunctionalInterfacesWithDefaults randomly fails
Steps to reproduce:

0. Used Java 26 (also fails with OpenJDK-11)
1. Install pytest-randomly
2. Run in a loop pytest test/jpypetest/test_overloads.py

After some runs, the test testFunctionalInterfacesWithDefaults fails on the first assert.

``` python
    def testFunctionalInterfacesWithDefaults(self):
        def my_fun(x, y=None):
            return 'my_fun'
    
        def my_fun_vargs(x, *vargs):
            return 'my_fun_vargs'
    
        def my_fun_kwargs(x, **kwargs):
            return 'my_fun_kwargs'
    
        def my_fun_kw(*, keyword_arg=None):
            return 'my_fun_kw'
    
        class my_fun_class:
            def __call__(self):
                return 'my_fun_class'
    
        test2 = self.__jp.Test2()
>       self.assertRaisesRegex(
            TypeError, 'Ambiguous overloads found', test2.testFunctionalInterfaces, my_fun)
E       AssertionError: TypeError not raised by testFunctionalInterfaces

test/jpypetest/test_overloads.py:262: AssertionError
```

Thrameos:
This deal with a hashing issue on methods involving lambdas and functors.  The hashcode for two different argument lists were coming up the same.  It also deals with a multithreads cache in which if another thread somehow changes the cache in flight it would result in a potential bad usage.

Fixes #1365

Note that the bumpversion appears to be on 1.7.1 dev rather than 1.7.2dev so I think the last stage of the bumprev has not been completely applied, but I have just left that for now.


Thrameos:
@marscher I am not sure that either of these explain the issue on windows segfault.   The order bug was just a hashcode being incomplete so that if it saw an argument number that matched the overload with the ``def foo(x,y=1):`` is would allow it to go through.   I tested and both paths were completely safe no matter which path was taken.   It randomly chose whether it dispatched to one argument, two arguments, or ambiguous overload.   So it was a red herring as the action was simply to prime the overload resolution causing to to dispatch to 1 or 2.  

I could resolve it one of two ways.  I can update the hash with additional information OR I can check to see if the parameter list included a callable and then let it fall through standard resolution as that avoided a second search through the argument list.   There is the off chance that one can still exercise the hash instability via passing a Java method directly to the method overload.  But I am not really sure how to exercise that as it requires tricking the resolver to consider one particular overload into the slot first then coming back with a different function.   If we can exercise it the fix is the same, if method type add the unique instance into the hash rather than the type.

The thread instability was possible to exercise but only if we were running GIL free as we should have been safely protected by the global lock.  So unless 3.12 on windows is allowing us to reach that point without a GIL guard it could not result in an error.  Unfortunately, that means we may still lack a reproducer of that rare race bug.   Not that your efforts are misplaced as this was certainly a bug that was going to bite us in a GIL free world, but it means that we have exactly $\pi$ bugs remaining to solve.

marscher:
Thanks! Your pi bug amount joke made my day :+1: 

I'd go for the bulletproof way in this case. There should be no path left which can be "abused"  to trigger a wrong resolution. Even if this means we have to run an extra iteration. I guess the usage patterns of passing Java callables are rare anyways.

marscher:
We do not know if this was really the cause of the Windows crash in CI. I do not believe that the code path is GIL free in Python 3.12. 
I've turned on "ReleaseWithDebInfo" as default build type for now and hope that if the crash occurs again, we receive some kind of insight.

Thrameos:
The AI suggested the issue might stem from a padding problem within one of our structs. MSVC is apparently much less stable when handling misalignment. However, the AI couldn’t provide specific suggestions on how to exercise or fix the bug, so I still believe we need a reproducer first. It’s a classic Heisenbug.

marscher:
Heisenbugs are the worst :skull:
I also intend to add a crash-handler which should produce stacks for all three components (Py, JVM, Extension), so we get some meaningful output in the future.

marscher (approved):

PR #1370, opened 2026-05-19
split matrix for OSX to contain both amd64 and arm64 wheels

marscher:


marscher:
release pipeline updates needs to be pushed to this repo, instead of a fork in order to test them.

PR #1372, opened 2026-05-20
Add arm64 Windows to _jvmfinder.

tdunkleArm:
On an arm64 Windows laptop, I'm able to build a native wheel for jpype without any code changes. However, the built wheel can only start the JVM if explicitly passed the arm64 `jvm.dll` path. This change solves that, so that the JVM can start on an arm64 Windows machine without the DLL path being passed explicitly. I've also added a unit test that confirms that `_checkJVMArch()` passes in the case of 0xAA64 (43620).

This change would hopefully enable native support for the arm64 Windows ecosystem for jpype, and CI/CD should be manageable with minimal overhead via the GitHub-hosted `windows-11-arm` [runners](https://docs.github.com/en/actions/reference/runners/github-hosted-runners).

### DIFF
--- a/jpype/_jvmfinder.py
+++ b/jpype/_jvmfinder.py
@@ -317,6 +317,7 @@ def _checkJVMArch(jvmPath, maxsize=sys.maxsize):
     IMAGE_FILE_MACHINE_I386 = 332
     IMAGE_FILE_MACHINE_IA64 = 512
     IMAGE_FILE_MACHINE_AMD64 = 34404
+    IMAGE_FILE_MACHINE_ARM64 = 43620
 
     is64 = maxsize > 2**32
     with open(jvmPath, "rb") as f:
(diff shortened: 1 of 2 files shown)


Thrameos:
Thanks!  This looks like an excellent contribution.

Thrameos:
@marscher I think you would better assess the runners as you have been working on them.   Is 1.71 pulled back into the main line?

marscher:
Thank you for this useful extension. I think we currently have no proper way of testing this, because our current CI pipeline runs on Azure and they do not offer a free ARM64 (despite OSX) runner for Windows.
I'm at doubt to add (and maintain) a duplicate of the testing pipeline for Github Actions just for this case and gonna trust it because of the implementation test you've thankfully provided.

marscher (approved):

PR #1385, opened 2026-06-29
Ambiguity overload fix

### LINKED ISSUE #1371: Incorrect ambiguity report
In resolving foo(Bar, Bar) vs foo(Bar...)  an ambiguous method overload is being reported.  Apparently, this antipattern is properly handled in Java.   The fix is to check for an ambiguity in org.jpype.typemanager.MethodResolution.java and add the varargs on to the veto list.   Easy fix for anyone who has time.

https://github.com/jsboige/CoursIA/issues/1325

Note the comment did not consider the 5th special case:
```
  // Special handling is needed for varargs as it may chop or expand.
    // we have 4 cases for a varargs methods
    //    foo(Arg0, Arg1...) as
    //       foo(Arg0)
    //       foo(Arg0, Arg1)
    //       foo(Arg0, Arg1[])
    //       foo(Arg0, Arg1, Arg1+)
```

Thrameos:
Fixes #1371 - Fix varargs vs fixed-arity method resolution

Varargs methods were incorrectly claiming specificity over fixed-arity
methods with the same parameter count, causing ambiguity errors or
wrong method selection.

Modified MethodResolution.isMoreSpecificThan() to ensure varargs methods
never outrank fixed-arity methods when both match, consistent with Java's
overload resolution rules (JLS §15.12.2.5).

Test cases added:
- testMinimalVsVarArgs: foo(String) beats foo(String, String...)
- testExpandedVsVarArgs: foo(S,S,S) beats foo(String, String...)  
- testFixedVsVarArgs: foo(S,S) beats foo(String, String...)


test/jpypetest/test_overloads.py
```diff
@@ -330,3 +330,133 @@ def testDerivedSub(self):
...
+        """Test that more specific varargs wins over less specific"""
+        test1 = self.__jp.Test1()
+        
+        # With String arguments, String... should be preferred over Object...
+        self.assertEqual('String,String...', test1.testVarArgsVsVarArgs('a', 'b', 'c'))
+        self.assertEqual('String,String...', test1.testVarArgsVsVarArgs('a', 'b'))
+        self.assertEqual('String,String...', test1.testVarArgsVsVarArgs('a'))
+
+    def testDiagnosticVarArgs(self):
+        """Diagnostic test to examine method resolution order"""
```
marscher, line 402: This test is only for human inspection, right?
There is no benefit of executing it every time the test suite runs (during CI). We can add unittest.skip("diagnostic test") as decorator.

Thrameos: done

marscher:
I think we should exclude mypy on the test bench, shouldn't we?

Thrameos:
Yeah, mypy has been a nuisance.  I suppose the best solution if we are going to apply it is make it part of the dev.mk system so that I can see it locally.  The problem with it is I get something done and then go to push and end up getting stuck doing cleanup activities.  I will put in a local formatter soon, so that burden will go down but these remote checkers end up causing cycle after cycle trying to clear.

marscher:
maybe we should use precommit for this purpose. It runs the formatters etc upon every commit and enforces all policies. I'm happy to discuss the usefulness of mypy or any other tool. I also noticed that codecov also complains a lot about patch diffs. But it helps me to spot uncovered code paths. It is just to nit-pick.

marscher:
may I ask, if the code has been generated too, or is this really your handwriting? I would have to invest more time to check everything if the code has been generated.

Thrameos:
Yes, definitely.   Once we have gotten the reverse bridge in place I recommend that we apply an agent to once again go through and systematically increase the coverage.  Doing it before a major push is just going to generate conflicts, but it would serve to lock our progress.   Codecov is great, but it does pick up a lot of noise.  You will notice that the “document only” PR have code coverage diffs.  Meaning that even if you have perfectly covered the changes it is often unhappy.

Thrameos:
The last batch of PR were generated automatically as I was trying to clear our issues list using the remains of my monthly allocation.   I did provide heavy prompting to aid it and made effort to locally review, but it made mistakes that I often terminated.   In particular,  it was instructed to make a minimum reproducer, get instructions on how to solve it, then complete the task, and then move it to the test bench in the same style as the original.  I then ran Q&A to figure out if it has considered each edge case.    That generally resulted in a pipe of extra testing some of which was necessary.  

It failed of formating most often, style frequently, poor solutions least frequency. Oten took unnecessary changes and was told to roll back and simplify.

marscher:
have you tried https://github.com/DietrichGebert/ponytail?

I think this approach is fine when fixing no brainer bugs. Method resolution however lies at the heart of JPype and I'm concerned this might change behavior unexpectedly (even though we provide lots of test cases).

Thrameos:
No have not tried ponytail as my employer limits me to their specific API access, so I get the very best in 2 year old tooling.

I can't say that my original was particularly good, though it may be worth more time processing this one.  The question that was being resolved here was the equals case  "foo(String, String...) vs foo(String, String) vs foo(String, String, String)".    The issue is that the code needs to test both orderings to see  A>B, B<A, A=B, B=A to make its decision about whether or not to add a overloads entry in the table.  That means both sides of the test must be resolve so that we can perform the sort and the overload table build.   The only case that needs adjustment is foo(String, String...) < foo(String,String) where previously it came out as equality.

The secondary constraint is the system needs to consider the primitive promotion and overloads.   Thus we need to test Parent, Child, Other.  Also required to handle each of the 8 promotion types that are possible in Java.   

It sounds like the best thing to add confidence is a deeper unittest.   We can have an agent write out N*(8+3)+2* (..., array end) combinations to depth 3 to verify that there are no missed cases.

Thrameos:
@marscher  We have another windows crash in the test matrix, but I have identified the source.  It appears to be a time bomb int the testGUI.  Once it runs the system crashes a fixed time from the execution.   I will place that a separate issue as it has nothing to do with the PR.

marscher (approved):

PR #1386, opened 2026-06-30
Fix for array slice issue

### LINKED ISSUE #845: Java array slice disregarded on copy constructor
```
>>> import jpype
>>> jpype.startJVM()
>>> ja=jpype.JArray(jpype.JDouble)([1,2,3,4,5,6])
>>> print(jpype.JArray(jpype.JDouble)(ja))
[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]
>>> print(jpype.JArray(jpype.JDouble)(ja[0:3]))
[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]
```

It should have only given the first 3 elements on the second query.

Thrameos:
Fix to long standing bug.  #845 

Fixes #845

### DIFF
--- a/native/python/pyjp_array.cpp
+++ b/native/python/pyjp_array.cpp
@@ -73,6 +73,24 @@ static int PyJPArray_init(PyObject *self, PyObject *args, PyObject *kwargs)
 			JP_RAISE(PyExc_TypeError, "Class must be array type");
 		if (arrayClass2 != arrayClass)
 			JP_RAISE(PyExc_TypeError, "Array class mismatch");
+
+		// Check if the input is a PyJPArray and if it's a slice
+		// If so, we need to clone it to get the actual sliced elements
+		if (PyObject_IsInstance(v, (PyObject*) PyJPArray_Type))
+		{
+			JPArray* srcArray = ((PyJPArray*) v)->m_Array;
+			if (srcArray->isSlice())
+			{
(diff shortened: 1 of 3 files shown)


native/python/pyjp_array.cpp
```diff
@@ -73,6 +73,24 @@ static int PyJPArray_init(PyObject *self, PyObject *args, PyObject *kwargs)
 			JP_RAISE(PyExc_TypeError, "Class must be array type");
 		if (arrayClass2 != arrayClass)
 			JP_RAISE(PyExc_TypeError, "Array class mismatch");
+
+		// Check if the input is a PyJPArray and if it's a slice
+		// If so, we need to clone it to get the actual sliced elements
+		if (PyObject_IsInstance(v, (PyObject*) PyJPArray_Type))
+		{
+			JPArray* srcArray = ((PyJPArray*) v)->m_Array;
```
marscher, line 81: is m_Array always a valid pointer?

marscher (approved):

native/python/pyjp_array.cpp
Thrameos: I thought that m_Array being non null was an invariant in the wrapper creation. Do I need to investigate?

marscher: You're right, it is an invariant indeed.

PR #1388, opened 2026-06-30
One line mismatch on calling doc fixed

### LINKED ISSUE #1149: help(Class) doesn't work
When I attempt to call `help(Var)` on a class named Var directly imported from the jar (which is in the same path as its javadoc jar counterpart):

```
File "C:\Users\lucad\AppData\Roaming\Python\Python310\site-packages\jpype\_jclass.py", line 166, in _jclassDoc
    jd = jde.getDocumentation(cls)
TypeError: No matching overloads found for org.jpype.javadoc.JavadocExtractor.getDocumentation(_jpype._JClass), options are:        
        public static org.jpype.javadoc.Javadoc org.jpype.javadoc.JavadocExtractor.getDocumentation(java.lang.Class)
```

Thrameos:
Fixes #1149

This was a long standing bug that was forgotten.

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -15,6 +15,8 @@ Latest Changes:
 
   - Fixed issue with library loading on Chinese systems. #1380
 
+  - Fixed type mismatch in JavadocExtractor causing help() to fail on Java classes. #1149
+
 - **1.7.1 - 2026-05-06**
 
   - Binaries for OSX ARM64.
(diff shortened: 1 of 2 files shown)


marscher:
Thanks for the fix! I thought we had a test for this, haven't we?

Thrameos:
I think we tested python slicing not Java to Java slicing.

marscher:
I guess you wanted to reply to the array PR. I had in mind that we have a java doc extraction test, which should have failed. But probably that's not the case.

marscher (approved):

PR #1390, opened 2026-06-30
Investigated and documented root cause

### LINKED ISSUE #992: No method overload found if the argument type is Enumeration object
I have an "enum" class(**InterfaceRoleType**) in java which implements "Enumerator" class.
There is another class named **Interface** which have method called "setRole(InterfaceRoleType agrObj)" which takes enum object as an argument.
Now, while using the Interface class via "jpype 1.3.0" to set the Role(setRole) i am passing an object of InterfaceRoleType(obviously), but jpype could not able to find the overload method in Interface class(which actually exist) and also jpype is throwing the "TypeError: No matching overload found" and also printing the options as the same method which is required and present (To be more clear below is stacktrace: )
```
 JPMethodDispatch::findOverload: EXCEPTION THROWN 0x854e60 No matching overloads found for com.model.impl.InterfaceImpl.setRole(com.model.InterfaceRoleType), options are:
	public void com.model.impl.InterfaceImpl.setRole(com.model.InterfaceRoleType)
      JPMethodDispatch::findOverload: EXCEPTION FROM:  native/common/jp_methoddispatch.cpp 184
      JPMethodDispatch::findOverload: EXCEPTION FROM:  native/common/jp_methoddispatch.cpp 195
```

Could you please help in pointing how i can resolve this issue, or any newer version in which we dont have the same problem.

Thrameos:
Fixes #992

This is not a JPype bug but rather a misunderstanding of how classes from different class loaders operate.  There is a workaround and it has been documented.

### DIFF
--- a/doc/userguide.rst
+++ b/doc/userguide.rst
@@ -6629,6 +6629,21 @@ does so), and once the JVM has been destroyed it restores the pre-JVM
 handlers — so a faulthandler enabled before ``startJVM`` is functional again
 after shutdown, rather than leaving faults routed into a destroyed JVM.
 
+This safety net covers the specific enable-before/disable-after sequence
+described above, plus any handler replacement still in place at
+``shutdownJVM``.  It does **not** cover every possible case: if code installs
+a new SIGSEGV/SIGBUS/SIGILL/SIGFPE handler *after* ``startJVM`` and that
+handler is still active, unremoved, at the moment a safepoint or JIT
+implicit-fault poll fires (i.e. before the next ``shutdownJVM`` call has a
+chance to detect and restore it), the JVM's fault will be routed into that
+handler instead of HotSpot's.  JPype cannot intercept arbitrary third-party
+calls to ``signal.signal()`` or ``sigaction()`` to prevent this — there is no
+portable hook that fires on every signal-handler change.  **The rule remains:
+do not replace SIGSEGV/SIGBUS/SIGILL/SIGFPE handlers while a JVM is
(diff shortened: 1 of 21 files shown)


doc/userguide.rst
```diff
@@ -2855,6 +2855,59 @@ Common Issues and Troubleshooting
     import _jpype
     _jpype.enableStacktraces(True)
 
+7. **Class Loader Conflicts with Third-Party Frameworks**: In environments
+   with multiple class loaders (e.g., OSGi, Java EE application servers,
+   Eclipse Modeling Framework), you may encounter "No matching overloads found"
+   errors even when types appear to match. This occurs because Java considers
+   classes loaded by different class loaders to be incompletely different types,
```
marscher, line 2865: dont you mean "to be completely different types"?

marscher:
Thanks for this update, this JVM behavior is very intuitive and therefore it is very nice to have it in the documentation.

doc/userguide.rst
Thrameos: That is one heck of a double negative sentence.  Yes "completely" would be better.  

 If should state only classes loaded by the same class loader are equivalent and you can create two copies that lack equality especially when working with 3rd party platform loaders.  

This one was very hard to replicate, but watching the ai struggle made me realize the only way this could happen was if there are two copies with the same name.

Thrameos: I will touch the wording up and see if I can fix a few more assuming I have tokens left before my monthly quota resets.

marscher (approved):

PR #1391, opened 2026-06-30
Documentation update on zerocopy arrays

### LINKED ISSUE #1341: Documentation: best way to access numpy arrays from Java without copying
Sorry if this is a FAQ, but I could not find this in the documentation, hence a request to improve the documentation.

Does `JArray.of(numpy_array)` map the data, or copy? The documentation is not clear, but I assume it copies.

The documentation states:
> NumPy arrays integrate seamlessly with JPype and allow fast, memory-efficient data transfers to Java. For example, a NumPy array can be mapped directly to a Java primitive array, enabling high-speed operations without unnecessary copying.

but there is no code example given that *maps* the data. Maybe this refers to the `JArray.of` function, but from the documentation this seems to return, e.g., `double[][]`; which given Java's memory layout is likely only possible with copying.
(well, if the data is row-major and has the right endianess, you could point java primitive arrays to the backing buffer, but this sounds horribly fragile with GC to not cause invalid frees?)

So assume we have some large data in the Python ecosystem, which often means in an *existing* numpy array.

I'd like to access this data from Java *without copying* (because of memory). It's fine if I need an adapter on the Java side, my code hopefully has the typing for that; even if that means I may need to implement slicing in Java. Say, I am fine with supporting *dense and contiguous* ndarrays only, so this will basically just boil down to offset computations into a direct buffer and potentially endianness conversions, I assume. (I understand that we will not get Java arrays from it, I am talking about buffers.)

How can we do this?

The documentation I found <https://jpype.readthedocs.io/en/latest/userguide.html#buffer-backed-numpy-arrays>
always *creates* a *new* buffer. Is there a way to expose an *existing* numpy buffer to Java without copying?

Can we simply convert the `.data` (and other fields of the ndarray)?
```python
jb = jpype.nio.convertToDirectBuffer(numpyarray.data)
```
– I could not find instructions on how to *zero-copy map* an existing numpy array to Java; but I'd imagine that a lot of users will be looking for this.

Thank you.

Thrameos:
Closes #1341

This was a documentation task only so resolve with a documentation update.

### DIFF
--- a/doc/userguide.rst
+++ b/doc/userguide.rst
@@ -1181,10 +1181,12 @@ JVM effectively.
    To ensure smooth data exchange, consider the following strategies:
 
    1. **Use NumPy Arrays**: NumPy arrays integrate seamlessly with JPype and
-      allow fast, memory-efficient data transfers to Java. For small to medium
-      arrays, use ``JArray.of()`` to convert NumPy arrays to Java arrays.
-      For large arrays, use direct buffers via ``jpype.nio.convertToDirectBuffer()``
-      to enable zero-copy access where both languages operate on the same memory.
+      allow fast, memory-efficient data transfers to Java. Use ``JArray.of()``
+      to convert NumPy arrays to Java arrays for best performance. For cases
+      where you need shared memory (changes in Java visible in Python) or when
+      memory is extremely constrained, use direct buffers via
+      ``jpype.nio.convertToDirectBuffer()`` to enable zero-copy access where
+      both languages operate on the same memory.
 
    2. **Leverage Java Buffers**: Java's `nio` buffers provide a mechanism for
       shared memory between Python and Java. These buffers are particularly
@@ -4219,8 +4221,9 @@ Transferring Arrays to Java
 ---------------------------
 
 NumPy arrays can be transferred to Java using the ``JArray.of()`` function. This
-**creates a copy** of the NumPy array as a Java multidimensional array. For zero-copy
-access to large arrays, see the section on "Zero-Copy Access to NumPy Arrays from Java".
+**creates a copy** of the NumPy array as a Java multidimensional array, which is
+typically the fastest approach. For scenarios requiring shared memory or extreme
+memory constraints, see the section on "Zero-Copy Access to NumPy Arrays from Java".
 
 **Example: Transferring a NumPy Array to Java**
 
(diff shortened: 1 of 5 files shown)


marscher (approved):

doc/userguide.rst
```diff
@@ -4242,6 +4244,357 @@ the structure of a NumPy array to a Java multidimensional array.
...
+
+- ✓ No memory copying (fast for large arrays)
+- ✓ Changes visible in both Java and Python
+- ✓ Memory efficient for large datasets
+- ✗ Requires manual index calculations for multi-dimensional arrays
+- ✗ Must ensure array is contiguous
+- ✗ More complex to use in Java code
+
+Choose based on your performance requirements and array sizes. For most use cases
+under 100MB, the simplicity of ``JArray.of()`` outweighs the copy cost. For larger
```
marscher: have you actually benchmarked this, or is this the assumption of the ai?

Thrameos: Definitely an ais imagination.  Ir ran some tests trying to figure out how the api worked, but not a full benchmark. 

I will see about getting a benchmark.   It is rather complex though as zero copy in jvm terms is relative.  Yes you can directly share a piece of memory, but as buffers cant directly become arrays that means you are only changing who is doing the copy.  

There is a jvm calls like map with critical section or copy to this buffer which we currently don't use.   I did a bunch iof benchmarks with panama on this where I implemented a single push one copy which was the exact sane thing as this buffer style pushing. (Panama segmenrs are directly appalled to buffers)  My conclusion was.... it saved memory but was actually slower for short buffers.  It may flip at some place.   The problem is transfers in java have guard checks and safe points.  The overhead of spinning up is a buffer is high.  Thus they are doing way more work.  The advantage thus wasn't in speed but rather smoothness of the jvm.  Making massive copies in C meant locking a thread from responsiveness.

marscher:
I'm not particularly happy about the benchmark stuff (which is nice per se) landed directly in the project root @Thrameos 

Especially because they are AI generated files. At least hide them somewhere (e.g. /project)

Thrameos:
Ouch!  Those were on my exclude list, they must have gotten dragged in when I pushed the documentation revision.   I guess the approval that was on the PR was stale as I was looking over clean up actions and never even noticed there was more than 1 file attached.  Unfortunately they are several merges back so I will just have to remove them as otherwise it will require rewriting the history.

We are going to need to find some way to contain CMakefile system, it makes my work directories into a massive mess of untracked files especially when I am building with 5 copies of Python, windows and linux in the development root.   I think the one on reverse branch is better at it shifts the noise to "build/".  The problem is the  ``editable.mode = "inplace"`` which just buries my work directory in garbage directories.

PR #1414, opened 2026-07-01
#1413: Handle "+" character in file paths

### LINKED ISSUE #1413: Literal "+" characters in file paths are incorrectly converted to spaces
I discovered this small bug while attempting to enumerate the contents of a `JPackage` via `dir` when that package was provided by a JAR with a `+` in the file name.

```python
import jpype
import jpype.imports

# Say com.foo.bar is provided by a JAR with the name foo-1.2.3-SNAPSHOT+456-g0123456789.jar`
# and com.foo.bar contains a class Baz
import com.foo.bar

print(dir(com.foo.bar))
# Output: []
# Expected: ['Baz']
```

Directly importing the class works just fine (i.e., `from com.foo.bar import Baz`). I believe that this is because this results in the direct instantiation of a `JClass`, rather than a `JPackage`, and `JPackage.__dir__` invokes `JPypePackageManager.collectContents`, where this `toURI` re-encoding workaround is used:

```java
 uri = new URI(
         uri.getScheme(),
         URLDecoder.decode(uri.getSchemeSpecificPart(), StandardCharsets.UTF_8),
         uri.getFragment()
 );
```

The documentation for `URLDecoder.decode` says:

>  The following rules are applied in the conversion:
>    * The alphanumeric characters "a" through "z", "A" through "Z" and "0" through "9" remain the same.
>    * The special characters ".", "-", "*", and "_" remain the same.
>    * **The plus sign "+" is converted into a space character "   " .**
>    * A sequence of the form "%xy" will be treated as representing a byte where xy is the two-digit hexadecimal representation of the 8 bits. Then, all substrings that contain one or more of these byte sequences consecutively will be replaced by the character(s) whose encoding would result in those consecutive bytes. The encoding scheme used to decode these characters may be specified, or if unspecified, the default encoding of the platform will be used. 

Thus, I believe the most targeted fix should just be URL-encoding "+" characters before the decoding workaround. I've linked a PR that does just that.

Thank you for this awesome project! 😄

baxter-huntington:
`URLDecoder.decode` applies the `application/x-www-form-urlencoded` MIME format, which includes converting "+" characters to space characters. If a path contains "+" characters, upon re-encoding, they will be converted to `%20` instead of `%2b`, yielding a different file path than we started with.

We handle this by converting "+" to "%2b" before decoding to ensure that it is preserved across the round trip.

Resolves: https://github.com/jpype-project/jpype/issues/1413

### DIFF
--- a/native/jpype_module/src/main/java/org/jpype/pkg/JPypePackageManager.java
+++ b/native/jpype_module/src/main/java/org/jpype/pkg/JPypePackageManager.java
@@ -469,9 +469,12 @@ public class JPypePackageManager
       // Java 8 bug https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8131067
       // Zip file system provider returns doubly % encoded URIs. We resolve this
       // by re-encoding the URI after decoding it.
+      //
+      // Replace "+" with its URL-encoded representation to avoid conversion to
+      // a space and ensure correct round-trip encoding.
       uri = new URI(
               uri.getScheme(),
-              URLDecoder.decode(uri.getSchemeSpecificPart(), StandardCharsets.UTF_8),
+              URLDecoder.decode(uri.getSchemeSpecificPart().replace("+", "%2b"), StandardCharsets.UTF_8),
               uri.getFragment()
       );
 


Thrameos:
Seems very reasonable, though we will likely need to improve it a bit before inclusion.

1) Is '+' the only problem character or do we need more substitutions?  

AI states this is likely needed.
```
/ Centralized protection for characters that have special meaning in 
// x-www-form-urlencoded but are literal characters in file paths.
private static String prepareForDecoding(String input) {
    return input.replace("+", "%2b")
                .replace("&", "%26")
                .replace("=", "%3D")
                .replace("#", "%23");
}
```

2) We need a doc/CHANGELOG entry so we track the improvement
3) We need something in test/jpypetest that exercises the bug.

I can generate those for you when I have development cycles free.  Or you can add it and we can skip straight to review.

Thanks for the contribution!

baxter-huntington:
@Thrameos I appreciate the quick look and response 😃 I'd be happy to address those items. I should have some time to get to them tomorrow.

marscher:
Thank you for you contribution. It was included in #1449

PR #1427, opened 2026-07-11
Fix issue #598: bytearray ambiguity between byte[] and char[]

### LINKED ISSUE #598: Char[] and byte[] ambiguity between bytes, str, and bytearray
One of the left over issues from Python 2 is the ambiguity between bytes, bytearray, and str when selecting the overload for byte[] and char[].   Simple commands like ``JString(bytearray([1]))`` currently fail with ambiguity between  ``java.lang.String(char[])`` and ``java.lang.String(byte[])``.  bytearray is pretty obviously best matched to ``byte[]`` but we have three implicit conversions defined that match ''PyUnicode'', ''PyBytes'', and ''PySequence''. 

I would recommend modifying the matches to be 
byte[] <=> str (none)    reason: string to math type prohibited)
byte[] <=> bytes, bytearray (exact)  reason: bytes are bytes
byte[] <=> Sequence (implicit)  reason: a list of numbers can be converted
char[] <=> str (exact)
char[] <=> Sequence, bytes, bytearray (implicit) 

Personally I think the last one should be "none" but it would break code.

Thrameos:
Resolved ambiguous overload resolution when passing bytearray to Java constructors/methods that accept both byte[] and char[].

Changes:
- bytearray now matches byte[] with _derived priority (beats char[])
- Added bytearray support to JPConversionByteArray
- Reordered conversions to check byteArray before buffer
- Excluded bytearray from buffer conversion for byte[]/char[]
- Added unit tests for bytearray and bytes conversions
- Updated CHANGELOG.rst

Tests:
- test_jstring.py: testBytearrayConstructor, testBytesConstructor
- test_conversion.py: testBytearrayToByteArray, testBytesToByteArray

All existing tests pass. Backward compatible.

Fixes #598


Thrameos:
@marscher  Claude found an instability in the Reference counting system but is buried deep in the reverse bridge.  I can try to get a backport to see if we can resolve these testing issues.  The problem was that certain traverals were not be launched, but I am not sure how related it is to this crash shown in testing.

Thrameos:
```

test/jpypetest/test_reflect.py::ReflectCase::testAccessPublicField PASSED
test/jpypetest/test_reflect.py::ReflectCase::testAnnotation PASSED
test/jpypetest/test_reflect.py::ReflectCase::testAccessPrivateField PASSED
test/jpypetest/test_jpackage.py::JPackageTestCase::testDir WARNING: JNI local refs: 41, exceeds capacity: 40
WARNING: JNI local refs: 74, exceeds capacity: 73
WARNING: JNI local refs: 107, exceeds capacity: 106
PASSED
test/jpypetest/test_jpackage.py::JPackageTestCase::testCreate PASSED
test/jpypetest/test_jpackage.py::JPackageTestCase::testFail3 PASSED

```

Okay so that is one of the frame safety and a loop that is not cleaning up its local references.  I can investigate that.

marscher:
> Okay so that is one of the frame safety and a loop that is not cleaning up its local references. I can investigate that.

Is this really a problem? Does it leak or is it just not optimal behavior? If it leaks we should fix it soon, if not we should just put on the low-prio list.

Thrameos:
It is suboptimal as with a large enough package you could smash the local registry in JNI.  I fixed it as it was just a missing DeleteLocal in the loop.   I think it was in the consolidated patch.

marscher (approved):

PR #1435, opened 2026-07-11
Add jdk.zipfs module dependency for jlink/jdeps support

### LINKED ISSUE #908: jdeps does not detect jdk.zipfs dependency
One of the tools required for jlink to build a limited jre distribution is jdeps.  Jdeps scans a jar file for module dependencies.  Unfortunately it can't detect any dependency which is invoked using reflection.  That included filesystem dependencies.  This means that trying to get the dependency list of org.jpype will give an incomplete dependency list and the resulting jre will not work.   

I have searched the documentation and tried adding a module-info.class and a MANIFEST entry.  Unfortunately there is no clear way to expose a hidden dependency in a jar file.  Nor can I find a method of exposing the requirement without manually importing a private internal class.   There has to be a way to handle this properly, but so far it eludes me.  Until then jdk.zipfs will have to be added manually.

Thrameos:
Fixes issue #908 where jdeps could not detect the jdk.zipfs dependency because it was loaded via reflection through FileSystemProvider.

JPypePackageManager uses FileSystemProvider.installedProviders() to access JAR filesystems, which loads jdk.zipfs via ServiceLoader. Since this is done reflectively, jdeps cannot detect it from bytecode analysis alone.

Adding 'requires jdk.zipfs;' to module-info.java makes the dependency explicit, allowing jlink to properly include it when building custom JRE distributions.

Fixes: #908

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -19,6 +19,8 @@ Latest Changes:
 
   - Fixed JArray constructor ignoring slice bounds when creating from sliced array. #845
 
+  - Added jdk.zipfs module dependency to module-info for proper jlink/jdeps detection. #908
+
 
 - **1.7.1 - 2026-05-06**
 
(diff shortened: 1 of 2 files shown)


Thrameos:
This was resolved with a module change.  At the time it came up we were not yet a module.

marscher (approved):

PR #1436, opened 2026-07-11
Fix crash when isinstance() is called with JException before JVM starts

### LINKED ISSUE #1329: Unhandled exception on isinstance(..., JException)
Thanks again for the project!

I have some code to wrap exceptions and send them to parent (python) process. For jpype, I had to add this because `JException` cannot be serialized:
```python
def wrap_into_traceback(e: BaseException, trim: int = 0) -> traceback.TracebackException:
    ...
    if isinstance(e, JException):
        e = RuntimeError(f"{e.message()}\nTraceback:\n{e.stacktrace()}")
```
This works most of the time except raising from `startJVM(jvmpath="does-not-exist")`. Trying to reproduce this in a minimal code, I see this error as `OSError: [Errno 0] JVM DLL not found` but in the production example it ~~segfaults~~ raises unhandled `JPypeException` on isinstance (fortunately, can see the traceback). Not sure how to debug this.
```python
terminate called after throwing an instance of 'JPypeException'
Fatal Python error: Aborted
...
Current thread 0x00007f42b216a740 (most recent call first):
File "/home/pulkin/.../something.py", line 87 in wrap_into_traceback
```

Thrameos:
Fixes issue #1329 where calling isinstance(obj, JException) would crash with an unhandled C++ exception when the JVM had not been started or had been shut down.

The problem occurred in PyJPClass_instancecheck() which created a JPJavaFrame without first checking if the JVM was running. This caused an unhandled JPypeException to be thrown from C++ code when users tried to wrap exceptions from failed startJVM() calls.

The fix adds a JVM running check before creating the frame, following the same pattern already used in PyJPClass_subclasscheck(). When the JVM is not running, the function now falls back to Python-only type checking.

Added comprehensive test coverage for:
- isinstance() before JVM starts
- isinstance() with non-Java objects after JVM starts
- isinstance() with actual Java exceptions

Fixes #1329 


Thrameos:
@marscher  simple change to prevent a raise condition that happened during exception handling.

test/jpypetest/test_jexception_isinstance.py
```diff
@@ -0,0 +1,46 @@
...
+            self.assertFalse(result)
+        
+        # Test with regular Python object
+        obj = "not a java object"
+        result = isinstance(obj, jpype.JException)
+        self.assertFalse(result)
+
+    def testIsInstanceWithJavaException(self):
+        """Test isinstance with actual Java exception"""
+        common.JPypeTestCase.setUp(self)
```
marscher: this one does not work as you expected. The JVM is being fired up in the conftest.py session fixture. The setup thingy is legacy code.
One solution would be to use the subrun method you developed.

marscher: what I'm saying is that the other test assumes the JVM is not yet running, which ain't true.

Thrameos:
Darn.  Looks like I will need to rebase these to get clean differences.  I will try to get to it on Saturday.

marscher (approved):

PR #1438, opened 2026-07-12
Revised memory model for JPype

Thrameos:
This is a WIP prototype that resolves problem of the Python unstable allocator by tricking Python into accepting a Java slot in every class.  It works by splitting the model into abstract classes that are object sized and concrete classes that contain a slot.  The only horrible hacks have  to do with variable sized objects which are fortunately resolvable using a slot trick.

Still a lot of clean up before it can be considered

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -7,6 +7,15 @@ Latest Changes:
 
 - **1.7.2.dev0**
 
+  - Fixed heap corruption when boxing large `JLong`/`JInt`/`JShort`/`JByte`/`JBoolean`
+    values on Python 3.8-3.11, caused by a fixed-offset allocator layout
+    assumption colliding with CPython's own implicit `__dict__` slot for
+    variable-length int subclasses.
+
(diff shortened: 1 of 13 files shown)


Thrameos:
@marscher As I never got feedback from the CPython core team, I decided to just go ahead and execute the plan.  Currently I have a memory corruption in <3.12 which I will need to resolve and deal with that long standing random coverage crash on random test order.  I will then backport all of the memory model fixes that were discovered in the reverse bridge.  That should bring us back to a stable point.   Unfortunately, my weekly tokens are completely dry so nothing more until next weekend.

marscher:
so without tokens no more work is possible? :skull:

Thrameos:
Deep debugging of segfaults is certainly possible but I have burned days of time on it in the past finding the tiny order error.  So I will hold off debugging this one until my home account pool recharges. 

Right now my work has a token crisis.  Better to tell management no tokens=no work until they figure out it is cheaper give tokens than having me spend 8 hours explaining i need tokens.

Thrameos:
superseded by #1445

PR #1440, opened 2026-07-12
Setup javac independence again

### LINKED ISSUE #1418: Source (sdist) builds now require a full JDK + Ant - prebuilt org.jpype.jar was dropped from the 1.7.x sdist
Like in title.

Would it be possible to restore a way to build the package without installing a JDK and Ant - e.g. by shipping the prebuilt org.jpype.jar in the sdist again (as 1.6 did), or by making the Ant/JDK jar build opt-out with the bundled jar as a fallback?

My situation: 
I don't use JPype directly. It comes in transitively as a peer dependency of another dependency in a project I do use, 
so I have no control over it being pulled in. I'm on arm32, where there are no prebuilt wheels, 
so installation always falls back to a source build - and I really don't want to install a Java toolchain on this system just to satisfy an indirect dependency. On 1.6.x this built fine with only a C++ compiler.

marscher:
If we have the support library in the source dir, we just it, instead of rebuilding it. 

There are two main paths here:
1. When built directly via CMake (e.g. via an IDE, shell etc.), the building of the Jar is enabled.
2. When built via pip/scikit-build-core, we disable the automatic building and try to rely on the cache.

fixes #1418

Minor changes:
- Updated CI to use uv pip for better download caching.
- Split CMakeLists.txt to a main file and one for the native native code.
- The wrapped test harness ant build cannot succeed in a sdist installl (no proper destination).


Thrameos:
Perhaps we should shift our test pipeline to be closer to our release pipeline and construct the source package and build from it.   One of the core discrepancies is the main build can read all green and the release system a total mess.  I don't want to run the full release as testing, but we could change it to use the same two phase structure.     I don't see any problems with this PR.


---

Again sorry about the files that leaked into the project root.   I will be more mindful.  But while we are on the topic of structure, I have been trying to route the next version to something a bit cleaner.   Given the scipy with inplace is a horrible mess, I did some restructuring on a local branch to try to make it more contained.

Bottom line, it is going to get even more complex when we are building both the maven and pip system in the same directory structure when the reverse bridge complete.    Here is what I think it should ultimately be laid out as

```
.azure/
.bumpversion.cfg
.git*
.readthedocs.yml
.agent(?) <= likely a hidden agent file that tells it what the structure of the project needs to look like so it stays in lane, not sure what should be named
AUTHORS.rst
CMakeLists.txt  <= Will move to be in src/main/cpp
LICENSE
MANIFEST.in
NOTICE
README.rst
codecov.yml
coverage.sh  <= Likely can be removed or moved to .azure/ or project/
doc/
examples/
jpype/
lgtm.yml
src/  <= Was native but will be src/(main,test)/(java,cpp) etc as per maven 
pom.xml <= the sole java trigger( can run Cmake), but we will have a seperate trigger for running cmake separately during build.
project/ <= everything project development related such as scripts, requirements, benchmarks, beautifier
pyproject.toml
test/ <= we could structure this to be in the maven src/test/py system but I don't think that has much value.
test-requirements.txt  <= not sure we need this here anymore 
```

(ivy, resolve.sh, archive/, ant build system all removed)

The only transient files should be 
```
_jpyne.(version).so/dll  <= This one was needed to install the prehooks for Java launch because Java loadLibrary does not import with the right flags.  (name is negotiable)
_jpype.(version).so/dll  <=  (I got it to restore the file extensions, without which I am constantly fighting which version do I have built)
target/ <= The only transient at root level
issues/  <= issues pulled for for work
pr/ <= active pr to match with issues and identified related tasks
```

project/ also adds beautifiers applies in the pom.xml and triggered on checkin hooks.

The key place we have to watch is plans/ which often get generated.  They can have a lot of value in they show what the what prompting/motivation was give, but they are often obnoxiously verbose AI files so I have been trying excluding them.   But they tend to escape when I have it working and committing as it goes.  I can shove it under project/plans/ or exclude entirely.

Thrameos (commented):
Did you mean to remove MANIFEST.in?

marscher:
I totally agree to the sdist first step in the testing pipeline. Then the only gap between testing and release is the platform dependent wheel building. We used to do it like that in the past.

The discussion about project file layout should be a separate issue/discussion. I think what you've just sketched here makes sense.

Manifest is only used in setuptools projects. The files set building is done in pyproject.toml.

PR #1445, opened 2026-07-16
Rework Java-backed object layout: fixed offsets, fungible boxed values

Thrameos:
Replaces the runtime allocator that re-derived each Java-backed Python object's layout from version-sensitive CPython internals (reading private PyLongObject fields on every single value access) with fixed, type-baked offsets resolved once at type-creation time. The abstract/concrete type split makes this offset a hard invariant rather than something callers branch on or recompute.

Builds on that to eliminate per-instance Java-value storage entirely for the boxed Long/Boolean/Character wrapper family: their JPClass* moves onto the shared metaclass (redundant per instance otherwise), and their jvalue is reconstructed on demand from the instance's own native representation -- boxing via a real JNI call only when a boxed wrapper's value actually crosses back into Java. This also removes a version-gated digit-layout workaround that existed solely to keep a trailing value at a fixed offset inside a variable-length PyLongObject subtype, along with a related heap corruption bug it caused on Python 3.8-3.11 when combined with an implicit __dict__ slot. A per-leaf-class null-boxed singleton represents JObject(None, cls), since a genuinely null Java reference has no native Python representation to reconstruct from.

No user-visible API change. Boxed wrapper instances no longer retain Java-side reference identity across repeated round-trips through Python -- this was never guaranteed for freshly-converted values either, so the change is specifically about previously-received-then-passed-back boxed objects.


Thrameos:
@marscher This is the final version of the remove all memory hacks in JPype memory model.  The initial state is we use an allocator to add memory for our object.  But that model was becoming increasingly hard to maintain as they both added more memory with inline dict, and then locked down all hooks required.  

To resolve this we had to introduce some polymorphism.  Python doesn't allow multiple inheritance once the object basesize is changed.  That means java.lang.Object and all Java interfaces can never uncreate the size.  To handle that we needed to split the tree into abstract (no change free to mixin) and concrete.  We also needed to handle the conflict that without the  allocator operating on the Python long get very challenging for boxed types.

This was the 3rd shot at a Python version free fix.  It adds minor cost as we must remint when handling boxed types, and more routine logic through the paths.   It now fully complies with the Python memory model, meaning we are unlikely to run into issues in Python until they themselves switch to light boxed types (on their roadmap).

marscher (approved):

Issue #1415, opened 2026-07-08
Crash in JPPyErrFrame::normalize()


marscher:
Stacktrace:
``` c++
#5  0x00007f131febd707 in JVM_handle_linux_signal (sig=<optimized out>, info=0x7ffc1fd58930, ucVoid=0x7ffc1fd58800, abort_if_unrecognized=<optimized out>)
    at src/hotspot/os/posix/signals_posix.cpp:652
#6  <signal handler called>
#7  PyType_HasFeature (type=<optimized out>, feature=<optimized out>) at /usr/local/miniforge3/envs/jpype/include/python3.14/object.h:795
#8  JPPyErrFrame::normalize (this=this@entry=0x7ffc1fd594b0) at /home/marscher/sources/jpype/native/python/jp_pythontypes.cpp:489
#9  0x00007f1321496e52 in JPypeException::toPython (this=this@entry=0x5613e5388c50) at /home/marscher/sources/jpype/native/common/jp_exception.cpp:329
#10 0x00007f132147be9b in PyJPModule_rethrow (info=...) at /home/marscher/sources/jpype/native/python/pyjp_module.cpp:871
#11 0x00007f1321477937 in PyJPArray_assignSubscript (self=0x7f129092d090, item=<optimized out>, value=<optimized out>) at /home/marscher/sources/jpype/native/python/pyjp_array.cpp:259
#12 0x00005613d51ddb3b in _PyEval_EvalFrameDefault (tstate=0x5613d55f1280 <_PyRuntime+315680>, frame=0x7f1436fb44a8, throwflag=0)
``` 

I can rarely reproduce it with:

py.test --randomly-seed=595374371 test

marscher:
It seems some Python versions do not check if the argument "type" is a valid pointer, e.g. it directly references type->tp_flags.

marscher:
Full stack trace extracted from a failed ci job:

``` c++
Thread 1 (Thread 0x7fb26deeab80 (LWP 3503)):
#0  __pthread_kill_implementation (no_tid=0, signo=11, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=11, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=11) at ./nptl/pthread_kill.c:89
[...]

PR #1447, opened 2026-07-19
Gc exception race2

Thrameos:
Okay this is the completed work for the exception race condition.  The problem is that any actions between the python exception being thrown and being normalized are dangerous.  If the GIL is ever lost a separate thread can come in an stomp the in flight exception causing corruption of memory.  The GC ping was the culprit, but it was in no way exclusive.  

We fixed this by moving normalization directly into the constructor for throw and then created a soak test to verify that it is fixed under stressful conditions.    I think this is a good candidate for review and then rebasing the existing PR as that will clear up our  PR board as only real errors will appear going forward.  

I will submit a separate PR that does  refactor of this exception code with the goal of making it more rational an direct to minimize this in the future.  The original monoexception path is the root cause of many of these bugs as we are constantly probing to confirm the current path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,5 @@ cmake_install.cmake
 _jpype.so
 org.jpype.jar
 .build_history
+CLAUDE.md
+plan/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,18 +220,7 @@ elseif(APPLE)
             VERBATIM
     )
 elseif(LINUX OR UNIX) # make sure APPLE is handled before this.
-    # Linux and OSX settings.
-#    find_library(LIBSTDCXX_SHARED NAMES stdc++ DOC "LibStdC++ runtime library required." REQUIRED)
-#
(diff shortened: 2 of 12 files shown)


Thrameos:
Drat the coverage one is still failing.  Will investigate.  I know however that this one does close a real race condition bug.

Thrameos:
I have determined the second crash is an independent JVM tear down bug.  I will isolate that and fix it, though it will be built on this PR.

Thrameos:
One crash has been linked to the race condition in the shutdown of jacoco. What appears to be happening is we trigger a shutdown and both JPype and Jacoco attempt to shutdown but Jacoco still has active resources pointing into JPype which then are nuked in a thread race.  Will need to add special logic, but I suspect this is not related to the current crash I am working on.

Thrameos:
Fixes #1415  but required additional PR to deal with 2 other paths related to shutdown and fault handlers.

Thrameos:
Superseded by #1448.

PR #1416, opened 2026-07-08
Enforce dynamic linkage to libstdc++

marscher:
Fixes #1415

Issue #1141, opened 2023-07-07
Segfault in JTerminate


atti92:
We have some rare problem in latest jpype, when the python process is closing.

```
Fatal Python error: Segmentation fault

Thread 0x00007fd21d345740 (most recent call first):
  File "/usr/local/lib/python3.10/site-packages/jpype/_core.py", line 356 in _JTerminate

[...]

Issue #842, opened 2020-08-16
SIGSEGV on termination


ShaheedHaque:
As per #720, to get 1.0.2 working for us, I moved JPype initialisation to be delayed it until actually needed. On process exit, it seems that if the initialisation code is not actually called, the process exists with a SIGSEGV:
```
Fatal Python error: Segmentation fault

Thread 0x00007f84e4194740 (most recent call first):
  File "/usr/local/lib/python3.8/dist-packages/jpype/_core.py", line 321 in _JTerminate
```
And here is line 321:
```
# In order to shutdown cleanly we need the reference queue stopped
# otherwise, we can experience a crash if a Java thread is waiting
# for the GIL.
def _JTerminate():
    try:
        _jpype.shutdown()  <<<<<<<<<<<<< 321
    except RuntimeError:
        pass
```
The *only* reference to jpype that this program could have made is an import as part of its transitive fanout:
```
import jpype
```
and it won't have invoked startJVM().

Thrameos:
Thanks for the report.  I will look into this as soon as my IT folks get my development machine repaired.   What is the value of “isJVMStarted()”?   Does adding an if statement for it (in Python or in pyjp_module) fix the issue?

ShaheedHaque:
First, I have noticed that the SIGSEGV is not easy to reproduce, happening perhaps 1 time in 20.

Second, I've added a print() of “isJVMStarted()”, and not seen the failure after a handful of runs.

Will report back after gathering more data.

ShaheedHaque:
I tried before, but there is a lot of stuff, and when I trimmed too far it stopped failing. Now that we have a slightly different problem, I'll try once again. I'll report back with any results.

mariusvniekerk:
is it possible that it gets added to the atexit handler twice?

mariusvniekerk:
What are the ramifications of doing?

```python
def _JTerminate():
    try:
         if _jpype.isStarted():
              _jpype.shutdown() 
    except RuntimeError:
        pass
```

Thrameos:
It should make sure that Java closes properly when Python exits.   Before we did not guarantee that Java files were properly closed nor that Java threads had terminated.   If you call shutdown manually then isStarted will be false and it will just operate as normal.  If Python exits without closing Java we perform the Java shutdown first.  If there are non-daemon threads then it will wait for them to terminate.

ShaheedHaque:
> What are the ramifications of doing?
> 
> ```python
[...]

PR #1448, opened 2026-07-19
Comprehensive segfault PR: Libstdc++, Normalization, GC Race, and Shutdown signal fix

### LINKED ISSUE #1144: Rare crashes on interpreter exit
Hi,
I am using JPype in a pytest context to execute SQL commands on an oracle database via JayDeBeApi. And I am experiencing segfaults of the python interpreter in less than **1% of the cases** after the testrun finishes. 

I set up a minimal test that just connects to the database and disconnects again with an execution time of ~5 seconds, which allows me to reproduce this bug fairly quickly in a loop. 

Unfortunately no python backtrace is printed and the generated coredump does not seem to yield a valid stacktrace, so maybe the bug is related to memory corruption? Note that I can fix/**work around it by calling jpype.shutdownJVM() manually**, but my understanding was that this should not be necessary, correct?

Is there anything I can do to analyze this issue so this could be fixed properly?

Thrameos:
This is aimed at resolving all the remaining shutdown hazards.  Built on the prior #1416 and #1447.   

Fixes  #1144, #1141, #842

This fix was sampled 500 times under different conditions on linux (all runs were clean).  We suspect a second error on Windows branch as the current signal handler would not explain the faults in #1447 windows run.  (extended testing found one hazard in the addition 500 runs and documented the root cause).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,12 @@ module = [
 ignore_missing_imports = true
 
 [tool.pytest]
-addopts = ["--verbose"]
+# pytest's built-in faulthandler plugin calls faulthandler.enable() before the
+# JVM exists and faulthandler.disable() at exit, which reinstalls the pre-JVM
+# SIGSEGV handler over HotSpot's.  The JVM then dies on the first armed
+# safepoint poll during shutdown (random ~25% crash rate in CI).  Keep the
+# plugin off; conftest.py enables faulthandler itself before startJVM.
+addopts = ["--verbose", "-p", "no:faulthandler"]
 testpaths = ["test", "jpype/_pyinstaller"]
 
 
--- a/test/jpypetest/conftest.py
+++ b/test/jpypetest/conftest.py
@@ -54,13 +54,8 @@ def jvm_session(request):
     logger = logging.getLogger(__name__)
 
     assert not jpype.isJVMStarted()
-    try:
-        import faulthandler
-        faulthandler.enable()
-        faulthandler.disable()
-    except:
-        pass
-
+    import faulthandler
+    faulthandler.enable()
 
     classpath = request.config.getoption("--classpath")
     convertStrings = request.config.getoption("--convertStrings")
@@ -72,7 +67,9 @@ def jvm_session(request):
     jvm_path = jpype.getDefaultJVMPath()
     logger.info("Running testsuite using JVM %s" % jvm_path)
     classpath_arg = "-Djava.class.path=%s"
-    args = ["-ea", "-Xmx256M", "-Xms16M"]
+    args = ["-ea", "-Xmx256M", "-Xms16M",
+            "-Xrs"  # disable signal handler for sigsegv, sigabrt
+            ]
     if checkjni:
         args.append("-Xcheck:jni")
     if classpath:
--- a/test/jpypetest/conftest.py
+++ b/test/jpypetest/conftest.py
@@ -51,6 +51,7 @@ def jvm_session(request):
     import warnings
     import faulthandler
 
+    # This installs signal handlers before the JVM is started.
     faulthandler.enable()
 
 
@@ -70,7 +71,7 @@ def jvm_session(request):
     logger.info("Running testsuite using JVM %s" % jvm_path)
     classpath_arg = "-Djava.class.path=%s"
     args = ["-ea", "-Xmx256M", "-Xms16M",
-            "-Xrs"  # disable signal handler for sigsegv, sigabrt
+            #"-Xrs"  # disable signal handler for sigsegv, sigabrt
             ]
     if checkjni:
         args.append("-Xcheck:jni")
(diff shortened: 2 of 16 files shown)


Thrameos:
During extended hazard testing there was one window in which setting signals before and then executing the shutdown outside of our internal safe code path with restore triggered a hazard.   It was at a rate of 1/500 in which the JIT triggered exactly in the window of the signal restore window.   I documented it as a limitation.  If you don't reset the signal handlers after JVM start than the risk is zero.

Thrameos:
@marscher there is a way to further reduce the hazard by setting it up for a periodic restore for the duration of the JVM operation, but it is for the rare case of a user messing with signal handlers during operation.  I should also note that the crashes we see here are NOT likely in user code.  The only reason we are experiencing them in unittesting is because the pytest is messing with the signal handlers.  That is why I have been running for years and never experienced this hazard with thousands of runs.  It requires a 3rd party library to trigger the signal handler mutation.

I will defer to you on whether a "maintainJVMHandlers" is required.

marscher:
Thank you very much for figuring out the very rare condition of the JVM faulthandler and GC circumstances. I'm also sure that the explicit dynamic libstdc++ linkage will catch some very rare crashes in the future.

I think we should cut another more stable patch release soon and then integrate the new (and exciting features).

test/jpypetest/conftest.py
```diff
@@ -69,7 +69,9 @@ def jvm_session(request):
     jvm_path = jpype.getDefaultJVMPath()
     logger.info("Running testsuite using JVM %s" % jvm_path)
     classpath_arg = "-Djava.class.path=%s"
-    args = ["-ea", "-Xmx256M", "-Xms16M"]
+    args = ["-ea", "-Xmx256M", "-Xms16M",
+            "-Xrs"  # disable signal handler for sigsegv, sigabrt
```
marscher: I think we should not disable JVMs own signal hanlding in testing, right? It will prevent the condition we are trying to test here.

pyproject.toml
```diff
@@ -66,7 +66,12 @@ module = [
 ignore_missing_imports = true
 
 [tool.pytest]
-addopts = ["--verbose"]
+# pytest's built-in faulthandler plugin calls faulthandler.enable() before the
```
marscher, line 69: this aint true. We explicitly called it our selfs in conftest.py, which is the main entry point for a pytest session (fixtures, configuration etc.) - claude should have known.

marscher: faulthandler.disable() is NOT being called at the session end. That's just hallucinated. I do not know how this wrong assumption does affect the solution the thing proposed.

marscher (approved):

Thrameos:
Claude was struggling over the 4 sessions it took to pin down race conditions, along with me explicit telling to to hypothesize different sources.   The source of the problem was real.  Any change to the fault handlers cause JIT to crash at random between the start of JVM and end.  

I did get good mileage out of the last of my Fable tokens as quite frankly I dont think this interaction is published anywhere.  So it constructed a massive test bench  of different possible causes and then statistically generated a reproducer for both gc and fault handler.   I then had to drop down to low effort to transfer the results back into the main code hence some bad comments getting transferred over.  The problem is both me and claude assumed it had something to do with test randomly meaning we generated hypothesis after hypothesis on interactions based on the random order from different trims.  

Regardless, it is still a timing race; we just closed the windows for the race to the minimum.  The good news is if you don't set the fault handlers it shouldn't blow up.  The gc exception interation on the Python execption path and the fault shutdown were some of the hardest bugs I think I have ever dealt with.  It also makes me very skeptical about GIL free Python ever being stable.  The only reason this fix worked is there was a timing where I could gather without releasing the GIL.  Without it any python code triggered by the gc link will cause the exception path to bust. 

There were two alternative solutions one using a java signal linker library (extra dependency) and the other monkey patching python fault handler installer.  But both were problematic as there was a second path in python we could not patch.  Meaning they were half measures. 

It is hard to assign blame.  Java is explicit about setting and even testing them during jvm start up.  And python at least as far as I know blindly sets then without considering proper chaining.  So we have two languages trying to be king of the world.

I am just glad we found any viable solution.  Are you sure through that pytest doesn't mess with the fault handlers?  I know that I had to do special configuration at some point because one version of pytest added a fault handler installer on me.  Through that was so long ago I can't recall the details.

Thrameos:
Regarding the fresh release, yes definitely.  I lined up a set of PR that I think constitutes a release in the milestone tag.   (Sorry for sending you through another pile of backlogged issues with ai pull requests!)   I will be taking my August vacation, so not sure I can be much assistance.    We can cut a fresh release with the current set or wait until the 3rd weekend of August when I can set another token bonfire to close all remaining issues.  

($20 a month only buys me 16 hours of code assistance spread over the reserve bridge, j2ni, and jpype maintenence then divided over 10  5 hour refill windows throughout the week.  Still I think they are losing money as I have burned it to negative tokens every single time.)

We still had one more hazard in Windows to debug.  Fortunately that one is tied to the gui test and only in headless.  I am not sure yet how to replicate it as a headless windows system with claude running is going to hard to source.  You may just want to add a skip test on windows for the problematic test the triggers a crash 10 seconds later.

PR #1449, opened 2026-07-20
Fix 1413 uri encoded plus rebased

### LINKED ISSUE #1413: Literal "+" characters in file paths are incorrectly converted to spaces

Thrameos:
Rebases #1414 by @baxter-huntington onto #1448 (their original commit is preserved as the base of this branch) and completes 
the items requested in review:
 - Added a doc/CHANGELOG.rst entry.
 - Added a regression test (test/jpypetest/test_startup.py::testPlusAndSpecialCharsPath) using a classpath jar in a directory named plus+path&has=special#chars, verifying +, &, =, and # all round-trip correctly. Empirically verified &/=/# were never actually mishandled by URLDecoder (only + was), so no extra escaping beyond the original fix was needed.
 - Fixed a real compile break in the original commit: URLDecoder.decode(String, Charset) requires Java 10+, but this module builds at --release 9, so the original fix never actually compiled under the project's real build. Switched to the Java 9-compatible URLDecoder.decode(String, String) overload.
 
Fixes #1413. Supersedes #1414 (opened from a fork branch we can't push to directly). All credit for the root-cause fix to @baxter-huntington.


project/jars/plus_path/sample_package/src/org/jpype/sample_package/A.java
```diff
@@ -0,0 +1,5 @@
+package org.jpype.sample_package;
```
marscher, line 1: I noticed you're moving towards this "project" structure for the reverse bridge. I think i might be easier (for an agent :P), if we keep the legacy layout (testing jars are still in test/jars and so on), to just transition once to the new layout in one shot.
At least I'm not confused.

It'd be also great not to commit binary files (jars itself) to the repo, although they are tiny (in this case). On some branch I had added the building of the testing jars to Ant, but I forgot which one.

marscher:
@Thrameos in https://github.com/jpype-project/jpype/pull/1268 I reorganized the testing structure a bit. The harness goes into the test subpackage inside `org.jpype`. In addition the building of the testing jars is automatized. At least the last step IMO is better than introducing binary files in the repo. What's your opinion on that?

Thrameos:
I am all for the reorg on testing.  I recommend it go in then we get through the pr back log, then I will rebase the reverse on it entirely to get it fresh.

Once we have it like that we should then take the current structure of the repo and create the reactory mavenized version with jpype/, modules/ target/ in place.  (The test harness will then be part of the maven build)  we run maven once to make the source dist which pulls in all deps.  (No review of reverse yet)

We then bust reverse into two pieces core changes (all c++ support and the changes required to remove all globals making ours subinterpreter safe), then a second feature add version with the whole embedded system (pyrhon.lang, JSR emdedding, linked stdio, spi wrappers)

That would leave you with two change reviews and one new code review.

marscher:
Sounds like a solid plan! Thanks!

Thrameos:
Agreed.  Project is supposed to be for project management not test files.
Not sure why it would place this here.

Once we reorganize it will be a maven like, so src/<main,test>  with
everything built landing in target.  (So we can strip and ship for the
source package)

Do you need me to clean it up next session or can we press on for
1.7.2/1.8.0?    I just finished the speed optimization and was aiming to
get reverse rebates on the revised model this weekend.   I know it has been
a hard press to get through all the issue backlog, but we are getting
close.

marscher:
Thanks. I think pushing forward towards the new release is more important than project organizational issues. We can sort this out later with the new Maven structure as you said.

PR #1450, opened 2026-07-20
Exception refactor

Thrameos:
Body:

## Depends on #1448

This branch is built on top of #1448 (shutdown-signal-handler fix) and
includes its commits. Please review/merge #1448 first — once it lands,
this diff will shrink to just the 7 commits below. Until then, ignore
`64023929`..`61c7d7b5` in the commit list; they belong to that PR.

## What

Replaces `JPypeException` — the single C++ class every exception crossing
the Java/Python/C++ boundary was carried as, tagged with a `JPError` enum
and a `void*`/`int` union payload — with one concrete type per exception
origin:

- `JPBaseError` — shared stack-trace bookkeeping, abstract `toPython()`/`toJava()`
- `JPJavaError` — a live `jthrowable`
- `JPPythonError` — an already-fetched, normalized Python exception
- `JPInternalError` — everything JPype raises directly (`JP_RAISE`, plus
  the startup-only OS-error paths)

Each type only carries (and only knows how t
for its own origin, instead of relying on comments to track which union
member goes with which tag value.

## Why

The tag-plus-union design directly contribut
project spent multiple days chasing (#1415, the GC-reentrancy crash, and
its `PyJP_SetStringWithCause` follow-on): `J
`_python_error`-tagged exception without an explicit type backing the
invariant "this holds an already-fetched Pyt
invariant existed only in a comment. A dedicated `JPPythonError` type
makes it a compile-time-checkable question i

## How it was done

Staged across 6 commits, each independently
test suite + a GC-reentrancy soak test (`project/bugs/gc_exception_race_soak.py`,
900+ iterations per stage) in an isolated ve

1. `10738693` — introduce the new classes, u
   `JPypeException`.
2. `dfd35fac`, `bf1bccd9`, `a07b802c` — migr
   (the `JP_RAISE*` macros and two direct throws) onto the new hierarchy.
3. `a987b918` — retarget the trace/stack-ann
   now-dead `catch (JPypeException&)` sites.
4. `7b126e53` — delete `JPypeException` once

**One real deviation from the original stagi
for review**: migrating the two hottest construction sites
(`JP_RAISE_PYTHON()` and `JPJavaFrame::check
call) ahead of schedule broke real functionality — proxy Java-calls-Python
dispatch and module resource loading both st
"unknown error occurred" instead of the real exception. Caught by the full
suite, not by inspection. Root cause: those
of the ~27 *scattered* direct `catch (JPypeException&)` sites (destructors,
buffer-protocol callbacks, resource cleanup)
two central "funnel" points. Fixed by migrating the affected catch sites
in the same commit (`bf1bccd9`) rather than
intermediate state.

## Test plan

- [x] Full suite green at every commit (1676/172 baseline, unchanged
      throughout; 1679/172 after picking up
      on rebase)
- [x] GC-reentrancy soak test clean at every
      no corruption)
- [x] Built and tested from a fresh clone in
      incrementally
- [x] `grep -rn "JPypeException"` returns no
      comments
- No user-facing behavior change; internal-o

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,5 @@ org.jpype.jar
 .build_history
 CLAUDE.md
 plan/
+project/issues
+project/pr
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -15,6 +15,13 @@ Latest Changes:
     pre-JVM handlers once the JVM is destroyed, making the JVM's lifetime
     signal-handler transparent on POSIX systems.
 
+  - Refactored internal exception handling: replaced the single
+    ``JPypeException`` tag-plus-union class with distinct C++ types per
+    exception origin (``JPJavaError``, ``JPPythonError``,
+    ``JPInternalError``), so each type only carries the payload valid for
+    it instead of relying on convention. No user-facing behavior change;
+    done to prevent future bugs of the class fixed by #1415 below.
+
   - Fixed a rare crash where Python's cyclic garbage collector firing
     while a Python exception was mid-unwind through the reverse-bridge
     C++ layer could corrupt the in-flight exception. #1415
--- a/native/common/include/jp_context.h
+++ b/native/common/include/jp_context.h
@@ -111,6 +111,12 @@ public:
 	friend class JPJavaFrame;
 	friend class JPypeException;
 	friend class JPClass;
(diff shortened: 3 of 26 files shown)


marscher:
That's a nice improvement. Thanks!

marscher (approved):

PR #1456, opened 2026-08-02
Caching

Thrameos:
This is speed improved JPype including caching for conversions to reduce the cost of type hint conversions.  This is built on model3.


marscher:
I'm sorry, what's model3?

Thrameos:
Model3 is #1445, a PR to change the memory model to remove the private allocator.  It splits objects into abstract and concrete privately so all objects have the jvalue in the object.  Jclass moved to pyJClass and the only hard case boxed types are rebuilt rather than reused.     Thus making us mostly immune from upstream changes to the memory model.

It had a secondary side effect of removing the type hack we previously used and thus doubled the number of objects we could allocate per second, so I used it as the basis for speed improvements.

Regarding this PR there is a lot of noisy files in benchmarking where the ai is debating feature differences and architectural costs by comparing to pyjnius, jep and jpy.    I have tried to keep it confined to project/.   I plan to clean up and squash once we close all gaps.  You can see the last one in the other benchmarking with list to array.  

I left this PR as it was about half way separately. the upgrade process so the rest of the speed tweeks can be viewed separately.

Thrameos:
Superseded by #1470, which is a superset of this PR's changes plus the phase-3 bulk array-transfer follow-on work. Closing in favor of that one.

PR #1458, opened 2026-08-02
Jvm destroy segfault

### LINKED ISSUE #1454: Add test and handling for corrupt stack after JVM shutdown
There is a rare case in which a user falils to mark a thread that calls a proxy to a blocking call than then outlives the JVM.  If this call is ever unblocked after the jvm it will result in termination.   

Though technically an error on the users part we can check for JVM state before returning on a proxy.  If the JVM is no longer present we can issue a warning and then terminate the thread or park it.  This will prevent the process from terminating with a corruption.    

Steps 
- [x] create a reproduced
- [x] add a state variable check in proxy
- [ ] test on Linux and windows to decide when it is better to park or terminate thread.
- [x] unit tests and changelog

Thrameos:
This fixes a rare segfault on shutdown where a daemon thread is still blocked during the JVM and then attempts to return to Java.

Fixes #1454

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -7,6 +7,18 @@ Latest Changes:
 
 - **1.7.2.dev0**
 
+  - Fixed a segmentation fault when a Python proxy (e.g. implementing
+    ``Runnable``) was invoked from a *daemon* Java thread and was still
+    blocked inside its Python callback when ``shutdownJVM()`` ran.
+    ``DestroyJavaVM()`` only waits for non-daemon Java threads, so shutdown
+    would complete - freeing every internal class/context resource - while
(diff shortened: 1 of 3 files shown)


Thrameos:
Added a documentation commit on top of the daemon-thread fix: a "Deciding Whether a Thread Should Block Shutdown" section in the userguide's Threading chapter, answering #1169's original question (when to call `detach()`) by spelling out the two options - attach as a user thread and architect an explicit wind-down signal, or attach as daemon and rely on the parking/termination this PR adds instead of the old segfault-on-resume.

Made sense to fold into this PR rather than a separate doc PR since the "safely parked instead of crashing" half of that explanation only exists once this fix lands.

PR #1459, opened 2026-08-03
Cmake build isolation

marscher:
This fixes the circumstance, that cmake generated files landed in the src dir.


Thrameos:
There is a change in reverse branch that deals with make cmake properly put things in the build dir that may help.  Not that it is urgent as everything will go to target once we switch to a maven like structure.

